PR #22, opened 2019-04-26
Solve incorrect "Show on GitHub" link

adtrevor:
Since the package isn't tagged yet, the current file prefix URL isn't valid, causing the declarations "Go to GitHub" button to lead to a 404 page.

### DIFF
--- a/scripts/generate_docs.sh
+++ b/scripts/generate_docs.sh
@@ -17,7 +17,7 @@ set -e
 
 my_path="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 root_path="$my_path/.."
-version=$(git describe --abbrev=0 --tags || echo "0.0.0")
+version=$(git describe --abbrev=0 --tags || echo "master")
 modules=(NIOHTTPClient)
 
 if [[ "$(uname -s)" == "Linux" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/generate_docs.sh
```diff
@@ -51,13 +51,20 @@ fi
 if ! command -v jazzy > /dev/null; then
   gem install jazzy --no-ri --no-rdoc
 fi
+
+if [[ "$version" == "0.0.0" ]]; then
```
tanner0101: `0.0.0` as a default seems kind of confusing to me. What happens if/when we tag a real 0.0.0? Maybe we should change the default to `"master"` if `git describe --abbrev=0 --tags` returns nothing.

tomerd: y i think that would be a better way to fix this

adtrevor: @tanner0101 @tomerd Thanks, done! Also realised that by using `master` instead of `0.0.0` we can still use the old url prefix! 🙂

tomerd:
thanks @adtrevor! can you rebase & squash so this has a single commit with the changes

adtrevor:
@tomerd Is this different from an admin directly merging it by choosing the squash and merge option?

tomerd:
@adtrevor nope, but its hard to review the actual changes in this PR right now because of the noise, so its more to help the review

adtrevor:
I understand, thank you! Will do it when back on my computer.

adtrevor:
@tomerd I did a soft reset to #fdb9dc10e5ac47e378b0d4c2278dd5dca858d7c3 followed by a commit using [this method](https://stackoverflow.com/a/5201642/2801589)
```bash
git reset --soft fdb9dc10e5ac47e378b0d4c2278dd5dca858d7c3
git commit
```
Yet there is still another commit and I'm not sure why the GitHub diff doesn't show the changes of this PR only.

adtrevor:
This time there only is one commit remaining, but the diff still shows more than I would hope...

tomerd:
i think head is 44ddc8bae0d054bf89ec8f0958f164574254abc0

adtrevor:
@tomerd Thank you! I had used the wrong commit indeed! Done! 🙂

PR #26, opened 2019-05-02
Small README fixes

Yasumoto:
Including an update to fixup `Timeout` and `Request` after the refactor in #10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ httpClient.get(url: "https://swift.org").whenComplete { result in
 }
 ```
 
-It is important to close client instance after use to cleanly shutdown underlying NIO ```EventLoopGroup```:
-```
+It is important to close the client instance, either in a ```defer``` statement or class' ```deinit``` after use to cleanly shutdown underlying NIO ```EventLoopGroup```:
+```swift
 try? httpClient.syncShutdown()
 ```
 Alternatively, you can provide shared ```EventLoopGroup```:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,8 +49,8 @@ httpClient.get(url: "https://swift.org").whenComplete { result in
 }
 ```
 
-It is important to close client instance after use to cleanly shutdown underlying NIO ```EventLoopGroup```:
-```
+It is important to close the client instance, either in a ```defer``` statement or class' ```deinit``` after use to cleanly shutdown underlying NIO ```EventLoopGroup```:
```
tomerd: clean shutdown worth a bit more explanation and examples. i worry folks will read this as they are suppose to call `httpClient.syncShutdown()` on deinit

//cc @weissi

weissi: I agree with @tomerd . Also please don't recommend or mention to use a class' `deinit`, that's poor style and will lead to really terrible bugs. People will still do it but at least we didn't recommend it :).

README.md
```diff
@@ -61,7 +61,7 @@ In this case shutdown of the client is not neccecary.
 
 ## Usage guide
 
-Most common HTTP methods are supported out of the box. In case you need to have more control over the method, or you want to add headers or body, use ```HTTPRequest``` struct:
+Most common HTTP methods are supported out of the box. In case you need to have more control over the method, or you want to add headers or body, use the ```HTTPRequest``` struct:
```
weissi: single word quoting should be done with `` ` `` and not `` ``` `` because `` ``` `` opens a code-block which in some renderers gets its own line.

Yasumoto: 👍 I copied the style found elsewhere in this doc, I'll fixup 🚫

README.md
Yasumoto: 😬great save!

Yasumoto:
Heya folks! Friendly nudge to get this another review 🙏

PR #31, opened 2019-05-08
WIP: Pooling

adtrevor:
This is a WIP for a pooling implementation based on #25

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,11 +23,12 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.0.0"),
+        .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.0.0"),
     ],
     targets: [
         .target(
             name: "NIOHTTPClient",
-            dependencies: ["NIO", "NIOHTTP1", "NIOSSL", "NIOConcurrencyHelpers"]
+            dependencies: ["NIO", "NIOHTTP1", "NIOHTTP2", "NIOSSL", "NIOConcurrencyHelpers"]
         ),
         .testTarget(
             name: "NIOHTTPClientTests",
--- a/Sources/NIOHTTPClient/ConnectionPool.swift
+++ b/Sources/NIOHTTPClient/ConnectionPool.swift
@@ -1,0 +1,376 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIOHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Swift Server Working Group and the SwiftNIOHTTPClient project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIOHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import NIO
+import NIOConcurrencyHelpers
+import NIOHTTP1
+import NIOHTTP2
+import NIOTLS
+
+class ConnectionPool {
+    let eventLoopGroup: EventLoopGroup
+    let configuration: HTTPClient.Configuration
+    private let lock = Lock()
+
+    init(group: EventLoopGroup, configuration: HTTPClient.Configuration) {
+        self.eventLoopGroup = group
+        self.configuration = configuration
+    }
+
+    private var _connectionProviders: [Key: ConnectionProvider] = [:]
+    private subscript(key: Key) -> ConnectionProvider? {
+        get {
+            return self.lock.withLock {
+                _connectionProviders[key]
+            }
+        }
+        set {
+            self.lock.withLock {
+                _connectionProviders[key] = newValue
+            }
+        }
+    }
+
+    func getConnection(for request: HTTPClient.Request) -> EventLoopFuture<Connection> {
+        let key: Key
+        do {
+            key = try Key(url: request.url)
+        } catch {
+            return self.eventLoopGroup.next().makeFailedFuture(error)
+        }
+
+        // FIXME: See what to do with timeout
+        _ = self.configuration.timeout
+
+        if let provider = self[key] {
+            return provider.getConnection()
+        } else {
+            let protocolChoosenPromise = self.eventLoopGroup.next().makePromise(of: ConnectionProvider.self)
+            let bootstrap = ClientBootstrap.makeHTTPClientBootstrapBase(group: self.eventLoopGroup, host: request.host, port: request.port, configuration: self.configuration) { _ in
+                self.eventLoopGroup.next().makeSucceededFuture(())
+            }
+            let address = HTTPClient.resolveAddress(host: request.host, port: request.port, proxy: self.configuration.proxy)
+            return bootstrap.connect(host: address.host, port: address.port).flatMap { channel in
+                let http1PipelineConfigurator: (ChannelPipeline) -> EventLoopFuture<Void> = { pipeline in
+                    pipeline.addHTTPClientHandlers(leftOverBytesStrategy: .forwardBytes).map { _ in
+                        let http1Provider = ConnectionProvider.http1(HTTP1ConnectionProvider(group: self.eventLoopGroup, key: key, configuration: self.configuration, initialConnection: Connection(key: key, channel: channel)))
+                        self[key] = http1Provider
+                        http1Provider.closeFuture.whenSuccess {
+                            self[key] = nil
+                        }
+                        protocolChoosenPromise.succeed(http1Provider)
+                    }
+                }
+
+                let h2PipelineConfigurator: (ChannelPipeline) -> EventLoopFuture<Void> = { _ in
+                    channel.configureHTTP2Pipeline(mode: .client).map { multiplexer in
+                        let http2Provider = ConnectionProvider.http2(HTTP2ConnectionProvider(group: self.eventLoopGroup, key: key, configuration: self.configuration, channel: channel, multiplexer: multiplexer))
+                        self[key] = http2Provider
+                        http2Provider.closeFuture.whenSuccess {
+                            self[key] = nil
+                        }
+                        protocolChoosenPromise.succeed(http2Provider)
+                    }
+                }
+
+                if request.useTLS {
+                    _ = channel.pipeline.addSSLHandlerIfNeeded(for: request, tlsConfiguration: self.configuration.tlsConfiguration).flatMap { _ in
+                        channel.pipeline.configureHTTP2SecureUpgrade(h2PipelineConfigurator: h2PipelineConfigurator, http1PipelineConfigurator: http1PipelineConfigurator)
+                    }
+                } else {
+                    _ = http1PipelineConfigurator(channel.pipeline)
+                }
+
+                return protocolChoosenPromise.futureResult.flatMap { provider in
+                    provider.getConnection()
+                }
+            }
+        }
+    }
+
+    func release(_ connection: Connection) {
+        if let provider = self[connection.key] {
+            provider.release(connection: connection)
+        } else {
+            // FIXME: Is this acceptable to discard this ELF?
+            _ = connection.channel.close()
+        }
+    }
+
+    func closeAllConnections() -> EventLoopFuture<Void> {
+        return self.lock.withLock {
+            let closeFutures = _connectionProviders.values.map { $0.closeAllConnections() }
+            return EventLoopFuture<Void>.andAllSucceed(closeFutures, on: eventLoopGroup.next())
+        }
+    }
+
+    fileprivate struct Key: Hashable {
+        init(url: URL) throws {
+            switch url.scheme {
+            case "http":
+                self.scheme = .http
+            case "https":
+                self.scheme = .https
+            default:
+                if let scheme = url.scheme {
+                    throw HTTPClientError.unsupportedScheme(scheme)
+                } else {
+                    throw HTTPClientError.emptyScheme
+                }
+            }
+
+            switch url.port {
+            case .some(let specifiedPort):
+                self.port = specifiedPort
+            case .none:
+                switch self.scheme {
+                case .http:
+                    self.port = 80
+                case .https:
+                    self.port = 443
+                }
+            }
+
+            guard let host = url.host else {
+                throw HTTPClientError.emptyHost
+            }
+            self.host = host
+        }
+
+        var scheme: Scheme
+        var host: String
+        var port: Int
+
+        enum Scheme: Hashable {
+            case http
+            case https
+        }
+    }
+
+    class Connection {
+        fileprivate init(key: Key, channel: Channel) {
+            self.key = key
+            self.channel = channel
+        }
+
+        fileprivate let key: Key
+        let channel: Channel
+    }
+
+    private enum ConnectionProvider {
+        case http1(HTTP1ConnectionProvider)
+        case http2(HTTP2ConnectionProvider)
+
+        func getConnection() -> EventLoopFuture<Connection> {
+            switch self {
+            case .http1(let provider):
+                return provider.getConnection()
+            case .http2(let provider):
+                return provider.getConnection()
+            }
+        }
+
+        func release(connection: Connection) {
+            switch self {
+            case .http1(let provider):
+                return provider.release(connection: connection)
+            case .http2(let provider):
+                return provider.release(connection: connection)
+            }
+        }
+
+        func closeAllConnections() -> EventLoopFuture<Void> {
+            switch self {
+            case .http1(let provider):
+                return provider.closeAllConnections()
+            case .http2(let provider):
+                return provider.closeAllConnections()
+            }
+        }
+
+        var closeFuture: EventLoopFuture<Void> {
+            switch self {
+            case .http1(let provider):
+                return provider.closeFuture
+            case .http2(let provider):
+                return provider.closeFuture
+            }
+        }
+    }
+
+    private class HTTP1ConnectionProvider {
+        init(group: EventLoopGroup, key: ConnectionPool.Key, configuration: HTTPClient.Configuration, initialConnection: Connection) {
+            self.loopGroup = group
+            self.configuration = configuration
+            self.key = key
+            self.closePromise = group.next().makePromise(of: Void.self)
+            self.configureConnectionEvents(initialConnection)
+            self.availableConnections.append(initialConnection)
+        }
+
+        deinit {
+            assert(availableConnections.isEmpty, "Available connections should be empty before deinit")
+            assert(leased == 0, "All leased connections should have been returned before deinit")
+        }
+
+        let loopGroup: EventLoopGroup
+        let configuration: HTTPClient.Configuration
+        let key: ConnectionPool.Key
+
+        /// The curently opened, unleased connections in this connection provider
+        var availableConnections: CircularBuffer<Connection> = .init(initialCapacity: 8) {
+            didSet {
+                self.closeIfNeeded()
+            }
+        }
+
+        /// Consumers that weren't able to get a new connection without exceeding
+        /// `maximumConcurrentConnections` get a `Future<Connection>` that is
+        /// completed as soon as possible, in FIFO order.
+        var waiters: CircularBuffer<EventLoopPromise<Connection>> = .init()
+
+        var leased: Int = 0 {
+            didSet {
+                self.closeIfNeeded()
+            }
+        }
+
+        let maximumConcurrentConnections: Int = 8
+
+        /// This future is completed when the connection pool has no more
+        /// opened connections
+        var closeFuture: EventLoopFuture<Void> {
+            return self.closePromise.futureResult
+        }
+
+        let closePromise: EventLoopPromise<Void>
+
+        let lock = Lock()
+
+        /// Get a new connection from this provider
+        func getConnection() -> EventLoopFuture<Connection> {
+            return self.lock.withLock {
+                if leased < maximumConcurrentConnections {
+                    leased += 1
+                    if let connection = availableConnections.popFirst() {
+                        return self.loopGroup.next().makeSucceededFuture(connection)
+                    } else {
+                        return self.makeConnection()
+                    }
+                } else {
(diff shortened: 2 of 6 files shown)


Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+        self.eventLoopGroup = group
+        self.configuration = configuration
+    }
+    
+    var eventLoopGroup: EventLoopGroup
+    var configuration: HTTPClient.Configuration
+    private var connectionProviders: [Key:ConnectionProvider] = [:]
+    private var lock = Lock()
+    
+    private subscript(key: Key) -> ConnectionProvider? {
```
adtrevor: I felt like it was better to make a function that would always warp gets/sets in lock since it's the main data structure that needs to be protected, the goal here is to reduce risks of threading issues and making less boilerplate.

Sources/NIOHTTPClient/SwiftNIOHTTP.swift
```diff
@@ -128,6 +131,30 @@ public class HTTPClient {
...
+                addedFuture = channel.eventLoop.makeSucceededFuture(())
+            }
+            return addedFuture.flatMap {
+                let taskHandler = TaskHandler(task: task, delegate: delegate, promise: promise, redirectHandler: redirectHandler)
+                task.setChannel(channel)
+                return channel.pipeline.addHandler(taskHandler)
+            }.flatMap {
+                return channel.writeAndFlush(request)
+            }
+            }.whenFailure { error in
```
Lukasa: This indentation is not right.

Sources/NIOHTTPClient/SwiftNIOHTTP.swift
```diff
@@ -128,6 +131,30 @@ public class HTTPClient {
...
+                promise.fail(error)
+        }
+        
+        return task
+        
+        /*
```
Lukasa: Is the commented out code here for a reason?

Sources/NIOHTTPClient/Utils.swift
```diff
@@ -17,38 +17,59 @@ import NIOHTTP1
...
     }
-
+    
     public func didReceiveError(task: HTTPClient.Task<T>, _ error: Error) {
         if let handler = handleError {
             handler(error)
         }
     }
-
+
```
Lukasa: Mind cleaning up these whitespace additions?

Sources/NIOHTTPClient/Utils.swift
```diff
@@ -17,38 +17,59 @@ import NIOHTTP1
...
     }
 }
+
+extension ClientBootstrap {
+    static func makeHTPPClientBootstrapBase(group: EventLoopGroup, request: HTTPClient.Request, configuration: HTTPClient.Configuration, channelInitializer: @escaping (Channel)->EventLoopFuture<Void>) -> ClientBootstrap {
+        return ClientBootstrap(group: group)
+            .channelOption(ChannelOptions.socket(SocketOptionLevel(IPPROTO_TCP), TCP_NODELAY), value: 1)
+            .channelInitializer { channel in
+                // let encoder = HTTPRequestEncoder()
+                // let decoder = ByteToMessageHandler(HTTPResponseDecoder(leftOverBytesStrategy: .forwardBytes))
```
Lukasa: What're these here for?

adtrevor: Oops, forgot to definitively remove it after I made this HTTP version agnostic

Sources/NIOHTTPClient/SwiftNIOHTTP.swift
adtrevor: Nope, has just kept it here for reference. Will be removed.

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+class ConnectionPool {
+    
+    init(group: EventLoopGroup, configuration: HTTPClient.Configuration) {
+        self.eventLoopGroup = group
+        self.configuration = configuration
+    }
+    
+    var eventLoopGroup: EventLoopGroup
+    var configuration: HTTPClient.Configuration
+    private var connectionProviders: [Key:ConnectionProvider] = [:]
```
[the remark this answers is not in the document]
adtrevor: Thanks! I'll use the swiftformat tool  (with the existing config file present in the repo) for all of those code style issues, I think it should get everything in line with the current NIO style.

Sources/NIOHTTPClient/SwiftNIOHTTP.swift
adtrevor: Looks like Xcode auto indent had issues there, I'll se if swiftformat fixes it and will correct it manually otherwise.

Package.swift
```diff
@@ -23,11 +23,12 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.0.0"),
+        .package(url: "https://github.com/apple/swift-nio-http2.git", from: "1.0.0"),
```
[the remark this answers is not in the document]
adtrevor: Ok! Better to keep the PRs well focused on a particular change indeed!

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+            lock.withLock {
+                connectionProviders[key] = newValue
+            }
+            
+        }
+    }
+    
+    func getConnection(for request : HTTPClient.Request) -> EventLoopFuture<Connection> {
+        guard let key = makeKey(for: request.url) else {
+            fatalError("Request URL is incorrect")
```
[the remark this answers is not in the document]
adtrevor: Done, makeKey is now a throwing function with appropriate errors

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+                
+                let http1PipelineConfigurator: (ChannelPipeline) -> EventLoopFuture<Void> = { pipeline in
+                    defer {
+                        protocolChoosenPromise.succeed(())
+                    }
+                    let http1Provider = HTTP1ConnectionProvider(group: self.eventLoopGroup, configuration: self.configuration, key: key)
+                    http1Provider.release(connection: Connection(key: key, channel: channel))
+                    self[key] = .http1(http1Provider)
+                    let encoder = HTTPRequestEncoder()
+                    let decoder = ByteToMessageHandler(HTTPResponseDecoder(leftOverBytesStrategy: .forwardBytes))
```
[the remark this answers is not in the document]
adtrevor: Thanks! Didn't know about them! 🙂

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+                    let future = pipeline.addHandlers([encoder, decoder])
+                    return future
+                }
+                
+                let h2PipelineConfigurator: (ChannelPipeline) -> EventLoopFuture<Void> = { pipeline in
+                    let http2Handler = NIOHTTP2Handler(mode: .client, initialSettings: nioDefaultSettings)
+                    let multiplexer = HTTP2StreamMultiplexer(mode: .client, channel: channel)
+                    return pipeline.addHandlers([http2Handler, multiplexer]).map { _ in
```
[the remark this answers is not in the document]
adtrevor: When I saw the configurators were taking a `ChannelPipeline` and not a `Channel`, I wondered if it was right to directly modify the (captured) `Channel `, so I preferred to just use the `ChannelPipeline` directly before taking time to see if it was right or not.
Why is `addHTTPClientHandlers` on `ChannelPipeline`, but not `configureHTTP2Pipeline`?

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+        } else {
+            var protocolChoosenPromise = eventLoopGroup.next().makePromise(of: Void.self)
+            let bootstrap = ClientBootstrap.makeHTPPClientBootstrapBase(group: self.eventLoopGroup, request: request, configuration: configuration) { channel in
+                
+                let http1PipelineConfigurator: (ChannelPipeline) -> EventLoopFuture<Void> = { pipeline in
+                    defer {
+                        protocolChoosenPromise.succeed(())
+                    }
+                    let http1Provider = HTTP1ConnectionProvider(group: self.eventLoopGroup, configuration: self.configuration, key: key)
+                    http1Provider.release(connection: Connection(key: key, channel: channel))
```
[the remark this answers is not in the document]
adtrevor: Would be clearer indeed but I don't really know what name to give it, `.addConnection(...)` maybe?

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+    
+    func getConnection(for request : HTTPClient.Request) -> EventLoopFuture<Connection> {
+        guard let key = makeKey(for: request.url) else {
+            fatalError("Request URL is incorrect")
+        }
+        
+        let timeout = configuration.timeout
+        
+        if let provider = self[key] {
+            return lock.withLock {
```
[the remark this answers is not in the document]
adtrevor: When you say `getConnection` would acquire the provider specific lock, do you think it's better to have the whole `getConnection` body acquire the lock, or to have it specifically on the points where we get/set data?
In both cases that's already a finer grained locking, and knowing that many parts of the getConnection body I'd probably lock the whole body, but I would appreciate to have your opinion on this.

Sources/NIOHTTPClient/ConnectionPool.swift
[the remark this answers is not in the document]
adtrevor: Fixed!

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+                }
+                
+                if request.useTLS {
+                    return channel.pipeline.configureHTTP2SecureUpgrade(h2PipelineConfigurator: h2PipelineConfigurator, http1PipelineConfigurator: http1PipelineConfigurator)
+                } else {
+                    return http1PipelineConfigurator(channel.pipeline)
+                }
+            }
+            let address = HTTPClient.resolveAddress(request: request, configuration: self.configuration, proxy: self.configuration.proxy)
+            return bootstrap.connect(host: address.host, port: address.port).and(protocolChoosenPromise.futureResult).flatMap { _ in
```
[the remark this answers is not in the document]
adtrevor: That's a nice improvement, thank you! Done

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+        
+        enum Scheme: Hashable {
+            case http
+            case https
+        }
+    }
+    
+    struct Connection {
+        fileprivate var key: Key
+        var channel: Channel
```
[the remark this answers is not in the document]
adtrevor: Turns out they didn't need to be `var`s! Thanks

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+            return bootstrap.connect(host: address.host, port: address.port).and(protocolChoosenPromise.futureResult).flatMap { _ in
+                return self[key]!.getConnection(for: request)
+            }
+        }
+        
+    }
+    
+    func release(_ connection: Connection) {
+        if let provider = connectionProviders[connection.key] {
+            provider.release(connection: connection)
```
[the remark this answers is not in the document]
adtrevor: Oops forgot about transforming this one into `self[connection.key]`, thanks!

Sources/NIOHTTPClient/ConnectionPool.swift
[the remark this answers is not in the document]
adtrevor: Turns out the pipeline configurators are called twice (probably Happy Eyeballs?) and the old approach was only working out of luck. Working on a fix

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+                
+                let h2PipelineConfigurator: (ChannelPipeline) -> EventLoopFuture<Void> = { pipeline in
+                    let http2Handler = NIOHTTP2Handler(mode: .client, initialSettings: nioDefaultSettings)
+                    let multiplexer = HTTP2StreamMultiplexer(mode: .client, channel: channel)
+                    return pipeline.addHandlers([http2Handler, multiplexer]).map { _ in
+                        defer {
+                            protocolChoosenPromise.succeed(())
+                        }
+                        self[key] = .http2(HTTP2ConnectionProvider(group: self.eventLoopGroup, key: key, channel: channel, multiplexer: multiplexer))
+                        channel.closeFuture.whenSuccess {
```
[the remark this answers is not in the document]
adtrevor: I haven't implemented it yet, still thinking about the approach to use. For HTTP/2 I'm removing the provider whenever it's underlying connection is closed, but for HTTP/1.1 I was planning to remove it when _all_ of its concurrent connections are closed, probably decrementing a count each time a connection closes, and succeeding a `EventLoopFuture<Void>` property of the provider when it equals `0`. I'm not sure this is the right way to this, though.

Sources/NIOHTTPClient/ConnectionPool.swift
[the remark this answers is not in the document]
Lukasa: The reason `configureHTTP2Pipeline` is on the `Channel` is because configuring the multiplexer requires access to the parent `Channel`, which you cannot get from the `ChannelPipeline`.

Sources/NIOHTTPClient/ConnectionPool.swift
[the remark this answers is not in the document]
Lukasa: I'd acquire it over the whole body of the method first, and then remove the scope later. It's always a bit tricky to have a method that potentially acquires and releases a lock multiple times.

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,326 @@
...
+    private class HTTP1ConnectionProvider {
+        init(group: EventLoopGroup, key: ConnectionPool.Key, configuration: HTTPClient.Configuration, initialConnection: Connection) {
+            self.loopGroup = group
+            self.configuration = configuration
+            self.key = key
+            self.closePromise = group.next().makePromise(of: Void.self)
+            availableConnections.append(initialConnection)
+        }
+        deinit {
+            assert(leased == 0, "All leased connections should have been returned before deinit")
```
adtrevor: I'm not sure yet of what to do there

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,333 @@
...
+            provider.release(connection: connection)
+        } else {
+            //FIXME: Is this acceptable to discard this ELF?
+            _ = connection.channel.close()
+        }
+    }
+
+    fileprivate struct Key: Hashable {
+        
+        init(url: URL) throws {
```
artemredkin: why not use `HTTPClient.Request` here? It already has all necessary fields resolved and verified?

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,333 @@
...
+import NIOHTTP2
+import NIOTLS
+
+class ConnectionPool {
+    init(group: EventLoopGroup, configuration: HTTPClient.Configuration) {
+        self.eventLoopGroup = group
+        self.configuration = configuration
+    }
+
+    let eventLoopGroup: EventLoopGroup
```
artemredkin: NIT: We usually have all class/struct fields defined before the `init` (here and in other places)

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,333 @@
...
+    func getConnection(for request: HTTPClient.Request) -> EventLoopFuture<Connection> {
+        
+        let key: Key
+        do {
+            key = try Key(url: request.url)
+        } catch  {
+            return eventLoopGroup.next().makeFailedFuture(error)
+        }
+
+        //FIXME: See what to do with timeout
```
artemredkin: what's wrong with:
```swift
if let readTimeout = timeout.read {
    return channel.pipeline.addHandler(IdleStateHandler(readTimeout: readTimeout))
} else {
    return channel.eventLoop.makeSucceededFuture(())
}
```
I think we can try to remove and re-add `IdleHandler` if timeout is different. Alternatively, timeout can be part of the `Key`

Sources/NIOHTTPClient/ConnectionPool.swift
adtrevor: That's because `Key` is used in contexts where the whole Request wouldn't seem really appropriate, for instance all `Connection` objects have a `key: Key` property, but they aren't linked to any specific `Request`, same thing for the connection providers.

Sources/NIOHTTPClient/ConnectionPool.swift
adtrevor: Since the pool consumer is already the one that adds its own `TaskHandler` I wondered if it wouldn't be best for the connection pool not to add the `IdleStateHandler` associated with `configuration.timeout either`.

I think it will still add an `IdleStateHandler`, but only for its unleased connections, based on a default timeout delay or on what was set in the `Keep-Alive` header, if available.

Does this seem correct to you?

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,376 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIOHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
weissi: @adtrevor do we need the `Foundation` dependency? If it's just for the lock, you can just `import NIOConcurrencyHelpers`

adtrevor: Yes because we need the `URL` type! 🙂

weissi: @adtrevor ah right. That's an interesting case because using the `URL` type is clearly the right choice. However using anything from `Foundation` comes with the drawbacks that static linking won't work (because `Foundation` links libcurl, libxml2, and libssl). For now let's just ignore it and continue with the Foundation dependency. We can always make a `NIOHTTPClientCore` module that contains everything (and uses `String` instead of `URL`) and a `NIOHTTPClient` module which does `@_exported import NIOHTTPClientCore`. That way almost all users would just use `NIOHTTPClient` and get the `Foundation` dependency but if you really need to not have `Foundation` you can still `import NIOHTTPClientCore`. But for now, let's just ignore this all.

SwiftMetrics pulls off a [similar trick](https://github.com/apple/swift-metrics/tree/master/Sources)

adtrevor: Interesting! I didn't know that Foundation caused this. Marking this as resolved for now, should we open an issue for this?

weissi: Let’s resolve this

Sources/NIOHTTPClient/ConnectionPool.swift
```diff
@@ -0,0 +1,267 @@
...
+        private func makeConnection(for request: HTTPClient.Request) -> EventLoopFuture<Connection> {
+            let bootstrap = ClientBootstrap.makeHTPPClientBootstrapBase(group: loopGroup, request: request, configuration: configuration) { channel in
+                let encoder = HTTPRequestEncoder()
+                let decoder = ByteToMessageHandler(HTTPResponseDecoder(leftOverBytesStrategy: .forwardBytes))
+                return channel.pipeline.addHandlers([encoder, decoder])
+            }
+            let address = HTTPClient.resolveAddress(request: request, configuration: self.configuration, proxy: self.configuration.proxy)
+            return bootstrap.connect(host: address.host, port: address.port).map { channel in
+                return Connection(key: self.key, channel: channel)
+            }
```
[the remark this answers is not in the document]
adtrevor: This seems pretty hard with the latest code changes as there are quite a few parts that differ. Having two different sources of `Channel`s (the one resulting from the ALPN negotiation, and the one a connection provider directly makes) isn't as much elegant as I would appreciate, though, I'll see if I find better ways to factorise this.

weissi:
friendly ping @adtrevor . Sorry, this is conflicting now due to the rename :|. Also we're gearing up towards `1.0.0` so if you'd need some public API change in order to implement the connection pool it'd probably be a good idea to try and get at least the public API change in now.

adtrevor:
@weissi Do you know what’s the approximate deadline for the last public API changes? I probably won’t have time to implement everything but hopefully I might be able to add the needed public API.

artemredkin:
@adtrevor we want to tag before the end of this week. I've look through the code and for connection pool there seem to be no API-breaking changes. I think the only public API change we will need is the configuration option, and that could be additive change, wdyt @weissi  ?

weissi:
> @adtrevor we want to tag before the end of this week. I've look through the code and for connection pool there seem to be no API-breaking changes. I think the only public API change we will need is the configuration option, and that could be additive change, wdyt @weissi ?

@artemredkin but we'd only tag something like `1.0.0-alpha.1`, correct? We can still break the public API if necessary before `1.0.0`. We should try not to break it too badly but I think with deprecations we should be able to create a smooth path. Only from `1.0.0` on public API changes are really a no-go. Still, it would be good to get all public API changes in asap.

weissi:
and yes, so far the implementation doesn't seem to need any public API changes which is great.

adtrevor:
Thinking about it, this should indeed not require any public API change. I’ll think about it again to see if I didn’t miss something but I don’t really think I did, at worse it would probably just be improvements, but nothing mandatory.
Thank you!

adtrevor:
@weissi Closing this for now, I'm currently missing time to continue it to the end, so I prefer not to pollute the PR section and to re-open it when I have time 🙂

Issue #32, opened 2019-05-12
HTTPCookie conflicts with Foundation


adtrevor:
The `HTTPCookie` type conflicts with the one from `Foundation`:
```swift
import NIOHTTPClient
import Foundation
let cookie: HTTPCookie // Error: 'HTTPCookie' is ambiguous for type lookup in this context
```

Maybe it should be made  a subtype of `HTTPClient`, to have something such as `HTTPClient.Cookie`?

jrtibbetts:
You can qualify it with the package name to avoid ambiguity, e.g. `NIOHTTPClient.HTTPCookie`.

weissi:
@jrtibbetts Yes, you can quality but I still think it's a good goal to at least not conflict with the basic libraries. Why? Let's assume you write code that doesn't `import Foundation` so you just use `HTTPCookie` everywhere. Further down the line you might want to add `import Foundation` and suddenly you need to qualify all uses of `HTTPCookie` which isn't too hard but not great.

So I'd say from a developer ergonomics point of view, we should aim to not clash with at least the stdlib, Foundation, and NIO.

jrtibbetts:
Oh, I completely agree that it should be renamed; I was just suggesting a workaround in the meantime.

weissi:
> Oh, I completely agree that it should be renamed; I was just suggesting a workaround in the meantime.

Ah sorry, misunderstood you!

Yasumoto:
Bumped into this myself so figured I'd pitch in with a take in #36.

PR #36, opened 2019-05-21
HTTPCookie is a subtype of HTTPClient

Yasumoto:
This prevents it from colliding with Foundation's top-level type, as described in #32.

I think putting it in `HTTPHandler` matches it with `Request`, `Response,` and `Body`, but I also considered keeping it in its own file as well as in `SwiftNIOHTTP.swift`. Ideas welcome!


Sources/NIOHTTPClient/HTTPHandler.swift
```diff
@@ -94,6 +94,111 @@ public extension HTTPClient {
         public var headers: HTTPHeaders
         public var body: ByteBuffer?
     }
+
+    struct HTTPCookie {
```
artemredkin: I think having in in a separate file would be better, we can have `extension HTTPClient` there, that will prevent `HTTPHandler.swift` from becoming overloaded.

Yasumoto: Cool, works for me. Think it should be called something like `HTTPClient+HTTPCookie.swift`?

artemredkin: Yep, that will work!

Yasumoto: Lemme know what you think! 🙇‍♂️

Sources/NIOHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -0,0 +1,123 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import NIOHTTP1
+
+extension HTTPClient {
+    public struct HTTPCookie {
```
tomerd: might as well name it `Cookie` under the new structure

Yasumoto: Good call, fixed!

artemredkin:

PR #41, opened 2019-06-03
Make Task.future public

Andrew-Lees11:
When trying to implement the README example for streaming, there were a few compile warnings.

- `HTTPResponseDelegate` has been renamed to `HTTPClientResponseDelegate`
- The protocol functions now have a `task` parameter
- `HTTPRequest` has been renamed `HTTPClient.Request`
- `httpClient.execute(request:, delegate:).future` is inaccessible since future is internal.

This pull requests fixes these issues in the Readme example. 

This includes making `Task.future` public. Since future is now public, I removed `wait` and `cascade` since they are now available on the future. This would be a breaking change and existing code in the project has been updated to reflect the change.

I have two remaining questions, was there a reason why `Task.future` was originally made internal and is the README example for accessing the streaming result correct?

### DIFF
--- a/Sources/NIOHTTPClient/HTTPHandler.swift
+++ b/Sources/NIOHTTPClient/HTTPHandler.swift
@@ -207,7 +207,7 @@ internal extension URL {
 
 public extension HTTPClient {
     final class Task<Response> {
-        let future: EventLoopFuture<Response>
+        public let future: EventLoopFuture<Response>
 
         private var channel: Channel?
         private var cancelled: Bool
@@ -226,10 +226,6 @@ public extension HTTPClient {
             }
         }
 
-        public func wait() throws -> Response {
-            return try self.future.wait()
-        }
-
         public func cancel() {
             self.lock.withLock {
                 if !cancelled {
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTPClient/HTTPHandler.swift
```diff
@@ -484,6 +476,6 @@ internal struct RedirectHandler<T> {
             request.headers.remove(name: "Proxy-Authorization")
         }
 
-        return self.execute(request).cascade(promise: promise)
+        return self.execute(request).future.cascade(to: promise)
```
artemredkin: Originally, I wanted `Task` to be sort of like a future, with all methods of the future. I think those methods make using `Task` a bit simpler.

Sources/NIOHTTPClient/HTTPHandler.swift
```diff
@@ -207,7 +207,7 @@ internal extension URL {
 
 public extension HTTPClient {
     final class Task<Response> {
-        let future: EventLoopFuture<Response>
+        public let future: EventLoopFuture<Response>
```
artemredkin: This issue will be addressed here https://github.com/swift-server/swift-nio-http-client/pull/37/files#diff-4345dbdc9050b28bde080d64fa672bebR232

ianpartridge:
Thank you for the PR. We improved this in #47 - please have a look at current `master` and see what you think.

PR #48, opened 2019-06-21
Pass timeout from get/post/patch/put/delete to execute

glbrntt:
`timeout` was never being passed along to `execute` so `execute` would always fallback to the timeout from `configuration`.


artemredkin:
Thanks @glbrntt !

PR #51, opened 2019-06-24
Split out @testable tests

ianpartridge:
Reduce the use of `@testable`. Where it is required, use it from a separate testsuite so its usage is more explicit.


Tests/LinuxMain.swift
```diff
@@ -27,6 +27,7 @@ import XCTest
 
     XCTMain([
         testCase(HTTPCookieTests.allTests),
+        testCase(SwiftHTTPInternalTests.allTests),
```
ianpartridge: This is the new suite which requires `@testable`.

Tests/NIOHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -15,13 +15,21 @@
...
 import NIOHTTP1
-@testable import NIOHTTPClient
+import NIOHTTPClient
 import NIOSSL
 
 class TestHTTPDelegate: HTTPClientResponseDelegate {
     typealias Response = Void
 
-    var state = ResponseAccumulator.State.idle
+    enum State {
```
ianpartridge: `@testable` was only required because we were piggy-packing on `ResponseAccumulator.State`. As we're not using the rest of `ResponseAccumulator` I think it's better to just define the state we need here.

Tests/NIOHTTPClientTests/SwiftNIOHTTPInternalTests+XCTest.swift
```diff
@@ -0,0 +1,35 @@
...
+
+///
+/// NOTE: This file was generated by generate_linux_tests.rb
+///
+/// Do NOT edit this file directly as it will be regenerated automatically when needed.
+///
+
+extension SwiftHTTPInternalTests {
+    static var allTests: [(String, (SwiftHTTPInternalTests) -> () throws -> Void)] {
+        return [
```
ianpartridge: These are the 5 tests which require `@testable`.

weissi:
That’s a great idea, thanks @ianpartridge .

ianpartridge:
Rebased, fixed conflicts, ready for review.

Tests/NIOHTTPClientTests/SwiftNIOHTTPInternalTests+XCTest.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+import XCTest
+
+///
+/// NOTE: This file was generated by generate_linux_tests.rb
+///
+/// Do NOT edit this file directly as it will be regenerated automatically when needed.
+///
+
+extension SwiftHTTPInternalTests {
```
artemredkin: One small comment, maybe we should rename it to `HTTPInternalTests` (drop `Swift`)? wdyt?

artemredkin:
One small question but otherwise looks good @ianpartridge, thanks!

Tests/NIOHTTPClientTests/SwiftNIOHTTPInternalTests+XCTest.swift
ianpartridge: `HTTPClientInternalTests` sounds best to me :)

Issue #42, opened 2019-06-04
Percent encoding is lost in the path


Andrew-Lees11:
When you make a request using with percent encoding in the path, the encoding is lost causing the request to fail. This is because at [HTTPHandler line 200](https://github.com/swift-server/swift-nio-http-client/blob/master/Sources/NIOHTTPClient/HTTPHandler.swift#L200) the path and query are used to get the url and these unencode the URL.

Andrew-Lees11:
Closing since this was fixed in pr 52

PR #52, opened 2019-06-25
feat: Add support for percent encoded requests

Andrew-Lees11:
This is a fix for issue #42

Previously when you made HTTP requests to a route with percent encoding, the encoding would be removed causing the request to fail. 

This request adds the percent encoding back to the path when you get the uri so that it can be correctly processed.


artemredkin:

artemredkin:
Looks good, thank you!

PR #57, opened 2019-07-01
add deadlines

artemredkin:
Add deadline argument to API methods

### DIFF
--- a/Sources/NIOHTTPClient/SwiftNIOHTTP.swift
+++ b/Sources/NIOHTTPClient/SwiftNIOHTTP.swift
@@ -58,16 +58,7 @@ public class HTTPClient {
         }
     }
 
-    public func get(url: String, timeout: Timeout? = nil) -> EventLoopFuture<Response> {
-        do {
-            let request = try Request(url: url, method: .GET)
-            return self.execute(request: request, timeout: timeout)
-        } catch {
-            return self.eventLoopGroup.next().makeFailedFuture(error)
-        }
-    }
-
-    public func get(url: String, deadline: NIODeadline) -> EventLoopFuture<Response> {
+    public func get(url: String, deadline: NIODeadline? = nil) -> EventLoopFuture<Response> {
         do {
             let request = try Request(url: url, method: .GET)
             return self.execute(request: request, deadline: deadline)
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTPClient/SwiftNIOHTTP.swift
```diff
@@ -58,16 +58,16 @@ public class HTTPClient {
         }
     }
 
-    public func get(url: String, timeout: Timeout? = nil) -> EventLoopFuture<Response> {
+    public func get(url: String, timeout: Timeout? = nil, deadline: NIODeadline? = nil) -> EventLoopFuture<Response> {
```
weissi: it should be either a dealine or a timeout I think, doesn't make sense to specify both.

artemredkin: I was thinking about it, but it will basically double the number of methods... Do you think it's ok? Also, we have to deal with both timeout and deadline anyway, since client can have configuration timeouts for example.

weissi: IMHO, better 2x methods that make sense rather then 1x methods that don't make sense.
If you really hate it, make an enum that is either timeout or deadline or just get rid of the timeouts :)

weissi: @tanner0101 / @ianpartridge what are your thoughts here?

ianpartridge: I would be tempted to only do deadlines. You can make a timeout from a deadline.

artemredkin: That's actually a good idea, replaced timeouts with deadlines

artemredkin: Actually, this is not a good idea :D, timeout is composite, consisting of read and connect timeouts, so it's not that strait-forward to convert deadline to timeout...

weissi: @artemredkin could you expand here? The user only cares that the `get` request is done at a certain point in time, right? They don't care how much of that time is spent in connect or read. Am I missing something?

ianpartridge: Agreed, I would put the read and connect timeouts on the overall client configuration, then put a deadline on the individual requests.

artemredkin: Lets try just deadlines, anything else worth fixing in this PR?

weissi: @artemredkin I'd say let's get it in

Issue #49, opened 2019-06-24
Fatal error if server responds with Payload Too Large during upload streaming


t089:
I've been playing with implementing a "max body size" policy in my NIO based web-server and noticed that this can trigger a `fatalError` in the state machine of `ResponseAccumulator`.

You can reproduce this as follows:

In `HttpBinHandler` change the `.body` handler to:

```swift
var response = self.resps.removeFirst()
guard response.body?.readableBytes ?? 0 < self.maxBodySize else {
    context.close(mode: .input).whenSuccess {
        context.write(self.wrapOutboundOut(.head(HTTPResponseHead(version: HTTPVersion(major: 1, minor: 1), status: .payloadTooLarge))), promise: nil)
        context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
    }
    self.resps.prepend(response)
    return
}
response.add(body)
self.resps.prepend(response)
```

Here `self.maxBodySize` is a constant `let maxBodySize: Int = 1_024 * 1_024 // 1 MB`.

When the server receives more than the allowed body size it will close the input stream and send a `payloadTooLarge` response to the client. Not sure if this is the correct way to implement this, though?!?

Anyhow, with this change this test case will trigger a `fatalError`:

```swift
func testUploadStreamingTooMuch() throws {
    let httpBin = HttpBin()
    let httpClient = HTTPClient(eventLoopGroupProvider: .createNew)
    defer {
        try! httpClient.syncShutdown()
        httpBin.shutdown()
    }
    
    let body: HTTPClient.Body = .stream(length: nil) { writer in
        var bytesWritten: Int = 0
        let chunkSize = 1_024 * 100
        let totalSize = 1_024 * 1_024 * 10
        var buffer = ByteBufferAllocator().buffer(capacity: chunkSize)
        buffer.writeBytes([UInt8](repeating: 0, count: chunkSize))
        func step() -> EventLoopFuture<Void> {
            let f = writer.write(.byteBuffer(buffer))
            let eventLoop = f.eventLoop
            return f.flatMap {
                print("Uploaded: \(bytesWritten) / \(totalSize) (\(round(Float(bytesWritten) / Float(totalSize) * 1000) / 10)%)")
                bytesWritten += chunkSize
                if bytesWritten < totalSize {
                    return step()
                } else {
                    return eventLoop.makeSucceededFuture(())
                }
            }
        }
        
        return step()
    }
    
[...]

PR #64, opened 2019-07-08
replace preconditions with errors

artemredkin:
Will address #49 as well

Before this PR, `channel.close` was called after `Task` promise was fulfilled. This introduced a slight delay before receiving `.end` and closing the channel. This in turn could lead to a situation where we could receive `.head` right after `.end`, which is allowed by HTTP protocol (keep-alive) but is not currently supported (pending connection pool work) by `ResponseAccumulator`. In order to address this, we need to close channel immediately after receiving `.end`.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,10 +1,12 @@
 # file options
 
+--exclude .build
+
 # format options
 
---self insert 
---patternlet inline 
---stripunusedargs unnamed-only 
+--self insert
+--patternlet inline
+--stripunusedargs unnamed-only
 --comments ignore
 
 # rules
--- a/README.md
+++ b/README.md
@@ -20,8 +20,7 @@ This library provides the following:
 Add the following entry in your <code>Package.swift</code> to start using <code>HTTPClient</code>:
 
 ```swift
-// it's early days here so we haven't tagged a version yet, but will soon
-.package(url: "https://github.com/swift-server/async-http-client.git", .branch("master"))
+.package(url: "https://github.com/swift-server/async-http-client.git", from: "1.0.0-alpha.1")
 ```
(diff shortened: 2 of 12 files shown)


artemredkin:
I'm trying to add a test, but it's not that straitforward :) NIO is pretty good at detecting when you do stupid things in your `HTTPServerHandler` :D

artemredkin:
Test added

artemredkin:
@weissi see #49, we receive `head`, `end`, `head` in some cases, but it's correct in terms of http protocol I think, but `HTTPResponseAccumulator` cannot handle second response on the same connection, I will investigate why channel was not closed

artemredkin:
@weissi I think `channel.close` happens too late, we receive `.end` and succeed the `promise` but callback (see line 358 in `HTTPHandler.swift`) on it's future is not called before we receive another `channelRead`, this is how we get into this inconsistent state.

weissi:
@artemredkin cool, but that should be fixed then, no?

artemredkin:
Question is, why task futureResult callback is called so late? It seems we need to call context.close right away in all cases

artemredkin:
@weissi I reverted back to preconditions (preconditions in redirect handler will be replaced by #67), real issue was that we were closing channel too late, I propose to close channel immediately instead of relying on futureResult callback.

weissi:
@artemredkin mind fixing the PR description and adding a good explanation of when exactly this will help? Also we definitely need a test case that proves this is effective.

artemredkin:
@weissi done

artemredkin:
This is not needed at all, underlying problem is solved by NIO:
https://github.com/apple/swift-nio/pull/1093/files

weissi:
@artemredkin released as swift-nio 2.6.1

Issue #38, opened 2019-05-23
Fire `didTransmitRequestBody` only when write future is filfilled


artemredkin:

adtrevor:
Should this event be sent once `body` has been transmitted but before `end`, or should it be sent after both have been transmitted?

Lukasa:
After both. In most cases `.end` produces a zero-length write which is trivially succeeded, but in some cases it produces actual bytes that must reach the wire.

artemredkin:
Yes, after both. Right now semantic of this callback is that `request is sent`.

artemredkin:
Fixed by #65

PR #65, opened 2019-07-08
call didSendRequest after write future is fullfilled

artemredkin:
To address #38


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -345,26 +345,28 @@ internal class TaskHandler<T: HTTPClientResponseDelegate>: ChannelInboundHandler
...
+                case .success:
+                    self.state = .sent
+                    self.delegate.didSendRequest(task: self.task)
+                    promise?.succeed(())
+
+                    let channel = context.channel
+                    self.task.futureResult.whenComplete { _ in
+                        channel.close(promise: nil)
+                    }
+                case .failure(let error):
```
ianpartridge: Don't we need to fail the promise in this case?

artemredkin:
ping @weissi @Lukasa, could you guys please take a look?

PR #66, opened 2019-07-10
make API non-throwing

tomerd:
motivation: cleaner API for users

changes:
* make request initializer optional instead of throwing
* http "sugar" methods (get, post, etc) no longer need to be throwing
* fix schema logic to be non-case sensitive
* made request object version, method and url immutable
* made request object scheme and host internal
* adjusted redirect ahndler implementation to stricter request immutabllity
* adjust and add tests

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -60,33 +60,34 @@ extension HTTPClient {
     }
 
     public struct Request {
-        public var version: HTTPVersion
-        public var method: HTTPMethod
-        public var url: URL
-        public var scheme: String
-        public var host: String
+        public let version: HTTPVersion
+        public let method: HTTPMethod
+        public let url: URL
         public var headers: HTTPHeaders
         public var body: Body?
+        
+        internal let scheme: String
+        internal let host: String
 
-        public init(url: String, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) throws {
+        public init?(url: String, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) {
             guard let url = URL(string: url) else {
-                throw HTTPClientError.invalidURL
+                return nil
             }
 
-            try self.init(url: url, version: version, method: method, headers: headers, body: body)
+            self.init(url: url, version: version, method: method, headers: headers, body: body)
         }
 
-        public init(url: URL, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) throws {
-            guard let scheme = url.scheme else {
-                throw HTTPClientError.emptyScheme
+        public init?(url: URL, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) {
+            guard let scheme = url.scheme?.lowercased() else {
+                return nil
             }
 
             guard Request.isSchemeSupported(scheme: scheme) else {
-                throw HTTPClientError.unsupportedScheme(scheme)
+                return nil
             }
 
             guard let host = url.host else {
-                throw HTTPClientError.emptyHost
+                return nil
             }
 
             self.version = version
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -60,33 +60,34 @@ extension HTTPClient {
...
 
-            try self.init(url: url, version: version, method: method, headers: headers, body: body)
+            self.init(url: url, version: version, method: method, headers: headers, body: body)
         }
 
-        public init(url: URL, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) throws {
-            guard let scheme = url.scheme else {
-                throw HTTPClientError.emptyScheme
+        public init?(url: URL, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) {
+            guard let scheme = url.scheme?.lowercased() else {
```
Yasumoto: I do agree that this is much nicer as a user, but we're sacrificing the type of error for a "this worked or not" without the context on what the problem was. I think I'd prefer to keep the granular errors vs. checking for `nil` instead, but if there's already been discussion (or others have the other preference!) happy to defer.

ktoso: I didn't follow all of the discussions here recently... but would it be an option to add address the above concern by adding `static func validated(<mirror init's param list>) throws -> Request`, users get to choose then, and if it was discussed that the `nil` is "nicer" then that's indeed the default still 🤔

Yasumoto: Great call, that helps with surfacing the specific issue if someone really wants it. 🧠

t089:
Not sure if this was discussed already, but how about this: if the given `URL` is not suitable for firing off the request, instead of throwing or returning `nil`, create an appropriate `Error` and immediately fail the `Promise/Task` with this error? This way the API is very easy to use (no do/catch, no optional check) and in the rare case that for some reason the URL is incomplete it will be catched by the error handling that is anyhow necessary when doing network requests.

artemredkin:
@t089 like in #56 ?

t089:
@artemredkin Ok, yes, sorry I got confused. Actually both PRs do this already right? The only remaining question is how to represent a validated URL as an internal type, if I understand correctly?

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: I find `nil`-returning functions like this *profoundly* frustrating as they're almost impossible to debug. It's very annoying to look at something you're confident is valid data and to have the program tell you that it is not valid, without any explanation as to why.

I'd be strongly -1 on this without at least the option for a constructor that gives me better diagnostics, but frankly I'm not sure if the convenience of the `nil` return is worth it for the inevitable failure of debuggability down the line.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -526,44 +527,36 @@ internal struct RedirectHandler<T> {
...
+        
         if convertToGet {
-            request.method = .GET
-            request.body = nil
-            request.headers.remove(name: "Content-Length")
-            request.headers.remove(name: "Content-Type")
+            method = .GET
+            body = nil
+            headers.remove(name: "Content-Length")
+            headers.remove(name: "Content-Type")
```
Lukasa: What's this whitespace about?

Lukasa:
As @weissi pointed out to me, if you want to replace a function that is `throws` with one that returns `nil`, there's a keyword for that: `try?`. It is *always* possible to trivially transform a throwing function to a `nil`-returning one. However, it is not possible to get the throwing information *back*.

Right now we notify the user of what exactly we think they did wrong, and I think it's simply the wrong thing to do to hide this information.

Sources/AsyncHTTPClient/HTTPHandler.swift
tomerd: @Lukasa this PR originated as a simpler and safer way compared to #43. totally agree on the information loss point, and think we can keep the API throwing. that said, the exiting throwing code has a validation/mutability race that needs to be addressed and this PR demonstrates that part

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -60,33 +60,34 @@ extension HTTPClient {
...
-        public var scheme: String
-        public var host: String
+        public let version: HTTPVersion
+        public let method: HTTPMethod
+        public let url: URL
         public var headers: HTTPHeaders
         public var body: Body?
+        
+        internal let scheme: String
+        internal let host: String
```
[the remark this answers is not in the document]
tomerd: @artemredkin url is a let in this PR

Sources/AsyncHTTPClient/HTTPHandler.swift
tomerd: this is the throwing version: https://github.com/swift-server/async-http-client/pull/67

tomerd:
this is an older proposal, the discussion is happening in #67

PR #67, opened 2019-07-10
improve request validation

tomerd:
motivation: safer handling of request validation and mutation

changes:
* fix schema logic to be non-case sensitive
* made request version, method and url immutable
* made request scheme and host internal
* adjusted redirect ahndler implementation to stricter request immutabllity
* adjust and add tests

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -91,16 +91,13 @@ extension HTTPClient {
         public var version: HTTPVersion
         /// Request HTTP method, defaults to `GET`.
         public var method: HTTPMethod
-        /// Remote URL.
-        public var url: URL
-        /// Remote HTTP scheme, resolved from `URL`.
-        public var scheme: String
-        /// Remote host, resolved from `URL`.
-        public var host: String
         /// Request custom HTTP Headers, defaults to no headers.
         public var headers: HTTPHeaders
         /// Request body, defaults to no body.
         public var body: Body?
+        private var _url: URL!
+        private var _scheme: String!
+        private var _host: String!
 
         /// Create HTTP request.
         ///
@@ -136,7 +133,15 @@ extension HTTPClient {
         ///     - `unsupportedScheme` if URL does contains unsupported HTTP scheme.
         ///     - `emptyHost` if URL does not contains a host.
         public init(url: URL, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) throws {
-            guard let scheme = url.scheme else {
+            self.version = version
+            self.method = method
+            self.headers = headers
+            self.body = body
+            try self.setURL(url)
+        }
+
+        public mutating func setURL(_ url: URL) throws {
+            guard let scheme = url.scheme?.lowercased() else {
                 throw HTTPClientError.emptyScheme
             }
 
@@ -148,18 +153,29 @@ extension HTTPClient {
                 throw HTTPClientError.emptyHost
             }
 
-            self.version = version
-            self.method = method
-            self.url = url
-            self.scheme = scheme
-            self.host = host
-            self.headers = headers
-            self.body = body
+            self._url = url
+            self._scheme = scheme
+            self._host = host
+        }
+
+        /// Remote URL.
+        public var url: URL {
+            return self._url
+        }
+
+        /// Remote HTTP scheme, resolved from `URL`.
+        public var scheme: String {
+            return self._scheme
+        }
+
+        /// Remote host, resolved from `URL`.
+        public var host: String {
+            return self._host
         }
 
         /// Whether request will be executed using secure socket.
         public var useTLS: Bool {
-            return self.url.scheme == "https"
+            return self.scheme == "https"
         }
 
         /// Resolved port.
@@ -167,7 +183,7 @@ extension HTTPClient {
             return self.url.port ?? (self.useTLS ? 443 : 80)
         }
 
-        static func isSchemeSupported(scheme: String?) -> Bool {
+        static func isSchemeSupported(scheme: String) -> Bool {
             return scheme == "http" || scheme == "https"
         }
     }
@@ -640,7 +656,7 @@ internal struct RedirectHandler<T> {
             return nil
         }
 
-        guard HTTPClient.Request.isSchemeSupported(scheme: url.scheme) else {
+        guard HTTPClient.Request.isSchemeSupported(scheme: request.scheme) else {
             return nil
         }
 
@@ -655,18 +671,10 @@ internal struct RedirectHandler<T> {
         let originalURL = self.request.url
 
         var request = self.request
-        request.url = redirectURL
-
-        if let redirectHost = redirectURL.host {
-            request.host = redirectHost
-        } else {
-            preconditionFailure("redirectURL doesn't contain a host")
-        }
-
-        if let redirectScheme = redirectURL.scheme {
-            request.scheme = redirectScheme
-        } else {
-            preconditionFailure("redirectURL doesn't contain a scheme")
+        do {
+            try request.setURL(redirectURL)
+        } catch {
+            return promise.fail(error)
         }
 
         var convertToGet = false
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -88,16 +88,19 @@ extension HTTPClient {
     /// Represent HTTP request.
     public struct Request {
         /// Request HTTP version, defaults to `HTTP/1.1`.
-        public var version: HTTPVersion
+        public let version: HTTPVersion
         /// Request HTTP method, defaults to `GET`.
-        public var method: HTTPMethod
+        public let method: HTTPMethod
+        /// Remote URL.
+        public let url: URL
+        /// Remote HTTP scheme, resolved from `URL`.
+        public let scheme: String
+        /// Remote host, resolved from `URL`.
+        public let host: String
         /// Request custom HTTP Headers, defaults to no headers.
         public var headers: HTTPHeaders
         /// Request body, defaults to no body.
         public var body: Body?
-        private var _url: URL!
-        private var _scheme: String!
-        private var _host: String!
 
         /// Create HTTP request.
         ///
@@ -133,14 +136,6 @@ extension HTTPClient {
         ///     - `unsupportedScheme` if URL does contains unsupported HTTP scheme.
         ///     - `emptyHost` if URL does not contains a host.
         public init(url: URL, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) throws {
-            self.version = version
-            self.method = method
-            self.headers = headers
-            self.body = body
-            try self.setURL(url)
-        }
-
-        public mutating func setURL(_ url: URL) throws {
             guard let scheme = url.scheme?.lowercased() else {
                 throw HTTPClientError.emptyScheme
             }
@@ -153,24 +148,13 @@ extension HTTPClient {
                 throw HTTPClientError.emptyHost
             }
 
-            self._url = url
-            self._scheme = scheme
-            self._host = host
-        }
-
-        /// Remote URL.
-        public var url: URL {
-            return self._url
-        }
-
-        /// Remote HTTP scheme, resolved from `URL`.
-        public var scheme: String {
-            return self._scheme
-        }
-
-        /// Remote host, resolved from `URL`.
-        public var host: String {
-            return self._host
+            self.version = version
+            self.method = method
+            self.url = url
+            self.scheme = scheme
+            self.host = host
+            self.headers = headers
+            self.body = body
         }
 
         /// Whether request will be executed using secure socket.
@@ -668,14 +652,7 @@ internal struct RedirectHandler<T> {
     }
 
     func redirect(status: HTTPResponseStatus, to redirectURL: URL, promise: EventLoopPromise<T>) {
-        let originalURL = self.request.url
-
-        var request = self.request
-        do {
-            try request.setURL(redirectURL)
-        } catch {
-            return promise.fail(error)
-        }
+        let originalRequest = self.request
 
         var convertToGet = false
         if status == .seeOther, request.method != .HEAD {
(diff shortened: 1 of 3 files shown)


Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import AsyncHTTPClient
+@testable import AsyncHTTPClient
```
artemredkin: i think it's better to move this test to `HTTPClientInternalTests`, this test class is intended to test public API methods

tomerd: https://github.com/swift-server/async-http-client/pull/67/commits/8775fc10569294c853f28403f23196226ad2ee08

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -77,7 +77,15 @@ extension HTTPClient {
...
         public init(url: URL, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) throws {
-            guard let scheme = url.scheme else {
+            self.version = version
+            self.method = method
+            self.headers = headers
+            self.body = body
+            try self.setUrl(url)
+        }
+
+        public mutating func setUrl(_ url: URL) throws {
```
tomerd: not sure if the preferred casing is `setUrl` or `setURL` i can go either way

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -529,18 +545,10 @@ internal struct RedirectHandler<T> {
...
-        }
-
-        if let redirectScheme = redirectURL.scheme {
-            request.scheme = redirectScheme
-        } else {
-            preconditionFailure("redirectURL doesn't contain a scheme")
+        do {
+            try request.setUrl(redirectURL)
+        } catch {
+            return promise.fail(error)
```
tomerd: i dont love this, open to other ideas

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: I think `setURL` is the preferred casing.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -89,24 +97,32 @@ extension HTTPClient {
...
-            self.version = version
-            self.method = method
-            self.url = url
-            self.scheme = scheme
-            self.host = host
-            self.headers = headers
-            self.body = body
+            self._url = url
+            self._schema = scheme
+            self._host = host
```
Lukasa: Do we need to recalculate these? Unless we have a strong performance argument for doing so, it seems like it'd be better for `scheme` and `host` to be computed properties.

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: I think this is right.

Sources/AsyncHTTPClient/HTTPHandler.swift
tomerd: two reasons:

=> if we use computed properties they will need to force unwrap since `host` and `scheme` are optional on `URL`. this means we rely on the fact we validated they are not empty in the url setter, which we do now. so could work. having said that, in my experience this would be a fragile design as someone in the future may decide to change the validation and forget there are computed properties that make this assumption leading to runtime errors. the current implementation makes this very explicit and hard to get wrong in the future 

=> we are lowercasing the `scheme` which is used for business logic and validation downstream code. using the url's schema will not include this manipulation and can cause business logic issues. we could instead create a new url object with the lowercased schema but the current implementation seems less intrusive and expensive

Sources/AsyncHTTPClient/HTTPHandler.swift
tomerd: its "right" but not very friendly API imo

Lukasa: It's as friendly an API as you can have in Swift when something can fail. There's just really no other good way to structure this code.

Sources/AsyncHTTPClient/HTTPHandler.swift
ianpartridge: Does this need to be `public`? I understand that `redirect()` needs to use it but that doesn't mean it needs to be `public`, does it?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -62,11 +62,11 @@ extension HTTPClient {
     public struct Request {
         public var version: HTTPVersion
         public var method: HTTPMethod
-        public var url: URL
-        public var scheme: String
-        public var host: String
         public var headers: HTTPHeaders
         public var body: Body?
+        private var _url: URL!
+        private var _schema: String!
```
ianpartridge: Shouldn't this be `_scheme` not `_schema`?

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: I think it should be `public` to make it easier for users to mutate a `Request` they may be holding.

ianpartridge: OK yes I guess it makes sense given a lot of the rest of the things on `Request` are mutable. We should have doc comments on `public var url/scheme/host` saying that if you want to change the property you should use `setURL()`.

tomerd: imo, mutating the request beyond headers and body is not a common enough use to justify exposing it as a public API, hence the original version of this PR: https://github.com/swift-server/async-http-client/pull/67/commits/88704d319b7280ba2bbd4c647032434d79581b67. the one use case where this is required is redirects, but in those cases you mutate more than just the url so constructing a  new request object is not that bad

@Lukasa and myself went back and forth on this in the context of #56 and i was willing to get that up, for the sake of making progress. 

that said, my opinion has not changed, so would be interesting to get other opinions into the discussion

tomerd: /cc @tanner0101

Sources/AsyncHTTPClient/HTTPHandler.swift
tomerd: i could go either way on this one,  no strong opinion. @artemredkin @ianpartridge @tanner0101 ?

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: We can go back and forth on this, but I'll state right now at the outset that I think there is no value in the discussion. There is no meaningful difference between a `Request` with a mutable `URL` and one that does not have one as long as it remains a `struct`: literally zero.

Here is the relevant bits of the definition of `Request` from the version of @tomerd's code that was the original, immutable version from this PR:

```swift
public struct Request {
    public let version: HTTPVersion       
    public let method: HTTPMethod
    public let url: URL	
    public var body: Body?

    internal let scheme: String
    internal let host: String

    public init(url: URL, version: HTTPVersion = HTTPVersion(major: 1, minor: 1), method: HTTPMethod = .GET, headers: HTTPHeaders = HTTPHeaders(), body: Body? = nil) throws {
        guard let scheme = url.scheme?.lowercased() else {
	}
	
        guard Request.isSchemeSupported(scheme: scheme) else {
        }
        guard let host = url.host else {
        }
        self.version = version
        self.method = method
        self.url = url
        self.scheme = scheme
        self.host = host
        self.headers = headers
        self.body = body
    }
}
```

This version, as @tomerd notes, has an immutable `url`, `scheme`, and `host` field. My thesis is that there is nothing you cannot do with this code that you can do with the version we have above, and vice-versa.

Here's my demonstration: I can write the `setURL` function as an extension.

```swift
extension Request {
    mutating func setURL(_ newURL: URL) throws {
        self = try Request(url: newURL, version: self.version, method: self.method, headers: self.headers, body: self.body)
    }
}
```

This function is exactly identical to the version in this patch except that it stresses the optimiser a bit, because it does a bunch of unnecessary work. I can even use it the same way in my code.

My core argument is this: if you have a `struct` there is very little value in making any of its fields immutable when those fields can be set via an initializer, because the setter can always be synthesised by simply calling that initializer. Making a field immutable achieves only two things:

1. Making it awkward to mutate the field
2. Sending a signal that perhaps mutating the field would lead to some logical error.

In this case, neither is true: there is no risk of generating an invalid structure by mutating `url` so long as we use a specific function to achieve it, and users should not be afraid of copying the structure in this way. We don't gain anything by making it hard to do this, and we do push costs onto users by doing it.

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: Tbh I don't think it's worth spending the cycles discussing it. Let's leave it as is, fixing the typo of `schema` to `scheme`.

Sources/AsyncHTTPClient/HTTPHandler.swift
tomerd: if i understand you correctly, the difference in opinion boils down to the last statement:

> Sending a signal that perhaps mutating the field would lead to some logical error. .... In this case, neither is true...

the current code has some fragile logic tied to url, scheme and host and mutating the url does cause *logical* issues which is this PR as well as #56 try to address. the fact we are talking about this so much should be a fairly strong signal that there is an issue here. my original  suggestion was to circumvent the current logical issue and potential future ones by not allowing mutation of the url **until we have a good reason to expose it**

> we don't gain anything by making it hard to do this, and we do push costs onto users by doing it

as far as i can judge there is not a good use case to justify exposing this to users behind redirect which we already handle and discussed above, so imo there is no *pragmatic* "cost" here to push onto the users

this is why i still think its a better trade-off to delay exposing a mutable URL until there s a good reason to do so, hope this makes more sense now

Lukasa: >  my original suggestion was to circumvent the current logical issue and potential future ones by not allowing mutation of the url until we have a good reason to expose it

And my counter argument is that there is no such thing as "not allowing mutation of the URL": the example code above does so. The question is only whether you can spell it "I'd like to change the URL", or whether you have to spell it "I'd like to create a new Request whose bits all happen to be the same as the one I already have".

I think it's really important to stress that making the `url` property `let` literally does not prevent the users from doing anything that this patch allows them to do. If you can construct a single example of what the current revision of the patch allows that cannot be achieved while the property is `let`, please let me know, because my understanding of Swift is that there is absolutely nothing.

tomerd: @lukasa of course! the point is not that user can't create a new request with a different url. the point is that the internal *logical* state of the request object is easier to manage by making url immutable, since changing the url triggers logic & validation. and since there is not a clear, pragmatic use case to make it mutable, we can get away by making it immutable and save ourselves maintenance headache, at least until a good reason shows up. hope this clarifies where i am coming from

tomerd: worth mentioning again i am okay moving forward with the mutable URL option too, for the sake of making progress. so i suggest at this point we ask @artemredkin @weissi @ianpartridge and @tanner0101 to express their opinions too and just move with whatever the majority feels is the right thing, as both options are technically viable and this is more opinions

Lukasa: I think this just represents a fundamental disagreement here, because I don’t think there is any more difficulty in validating the URL via a setter function versus via the constructor. Regardless, though, I think we’re arguing about technicalities: anyone can achieve anything regardless of what construction is used here, so we’ve fundamentally just exerted far more energy here than we need to.

Given that it seems that I’m the only one arguing for a mutator function I think it’s easier for me to simply withdraw the objection than to poll for consensus. Let’s just move on.

tomerd:
@artemredkin i rebased and made this immutable again. i think its ready to go

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -89,15 +89,15 @@ extension HTTPClient {
     /// Represent HTTP request.
     public struct Request {
         /// Request HTTP version, defaults to `HTTP/1.1`.
-        public var version: HTTPVersion
```
artemredkin: Johannes suggest to drop version from request completely (see discussion in #56)

tomerd: even better! done in https://github.com/swift-server/async-http-client/pull/67/commits/73c400f69ecefedc10ed22bf49a09eefe21dc832

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -88,16 +88,14 @@ extension HTTPClient {
 
     /// Represent HTTP request.
     public struct Request {
-        /// Request HTTP version, defaults to `HTTP/1.1`.
-        public var version: HTTPVersion
         /// Request HTTP method, defaults to `GET`.
-        public var method: HTTPMethod
+        public let method: HTTPMethod
```
artemredkin: does method have to be `let`? It's not part of the url change verification, so we can keep it as `var` (maybe move closer to all other vars in the file?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -649,22 +646,7 @@ internal struct RedirectHandler<T> {
...
-        } else {
-            preconditionFailure("redirectURL doesn't contain a host")
-        }
-
-        if let redirectScheme = redirectURL.scheme {
-            request.scheme = redirectScheme
-        } else {
-            preconditionFailure("redirectURL doesn't contain a scheme")
-        }
+        let originalRequest = self.request
```
artemredkin: I think having something like:
```swift
var request = try HTTPClient.Request(url: redirectURL, method: method, headers: headers, body: body)
...
if convertToGet {
    request.method = .GET
    request.body = nil
    request.headers.remove(name: "Content-Length")
    request.headers.remove(name: "Content-Type")
}

if !originalURL.hasTheSameOrigin(as: redirectURL) {
    request.headers.remove(name: "Origin")
    request.headers.remove(name: "Cookie")
    request.headers.remove(name: "Authorization")
}

return self.execute(newRequest).futureResult.cascade(to: promise)
```
could look event simpler, wdyt?

tomerd:
@artemredkin personally i prefer this as-is. if you want prefer method to be mutable and "nicer" redirect handler we can go back to https://github.com/swift-server/async-http-client/pull/67/commits/1e3e29fcf6165827993f859d37e4dd0320f8f1a6 and make setURL private. wdyt?

artemredkin:
@tomerd no, current version works for me, let's fix formatting issue and merge

PR #72, opened 2019-07-17
Prevent crash when using past or tight deadline (see issue#71)

ldewailly:


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -166,13 +166,19 @@ public class HTTPClient {
     }
 
     private func resolve(timeout: TimeAmount?, deadline: NIODeadline?) -> TimeAmount? {
-        switch (timeout, deadline) {
+        // compute timeout from deadline
+        var deadlineTimeout: TimeAmount?
+        if case .some(let deadline) = deadline {
+            let now = NIODeadline.now()
+            deadlineTimeout = deadline > now ? deadline - now : .nanoseconds(0)
+        }
```
Lukasa: Maybe consider changing this block to:

```swift
let deadlineTimeout = deadline.map { max($0 - .now(), .nanoseconds(0)) }
```

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -166,13 +166,19 @@ public class HTTPClient {
...
-            return min(timeout, deadline - .now())
+            return min(timeout, deadline)
         case (.some(let timeout), .none):
             return timeout
         case (.none, .some(let deadline)):
-            return deadline - .now()
+            return deadline
         case (.none, .none):
             return nil
         }
```
Lukasa: This mistake seems unnecessary: should this code be using `TimeAmount` at all? It seems like the better choice might be to be using `NIODeadline` everywhere, and turning timeouts in `TimeAmount` form into `NIODeadline`s ASAP.

If we don't do that, we should consider checking for extremely short time-amounts after this call returns (or possibly in this call) and timing them out early rather than making NIO do it.

ldewailly:

Sources/AsyncHTTPClient/HTTPClient.swift
ldewailly: `$0 - .now()` will crash the same

ldewailly: I'll use `map` as you suggested though

weissi:
I fixed this bug in https://github.com/apple/swift-nio/pull/1074

ldewailly:
I'll close this PR then since it's been fixed in `NIO`.

PR #77, opened 2019-08-08
Ignore uncleanShutdown error when state is .head or .body

vkill:
Hi there,

I think we should ignore uncleanShutdown when state is `.head` or `.body`, please review it. thx.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -211,7 +211,7 @@ public class HTTPClient {
                         return channel.eventLoop.makeSucceededFuture(())
                     }
                 }.flatMap {
-                    let taskHandler = TaskHandler(task: task, delegate: delegate, redirectHandler: redirectHandler)
+                    let taskHandler = TaskHandler(task: task, delegate: delegate, redirectHandler: redirectHandler, ignoreNIOSSLUncleanShutdownError: self.configuration.ignoreNIOSSLUncleanShutdownError)
                     return channel.pipeline.addHandler(taskHandler)
                 }
             }
@@ -276,19 +276,23 @@ public class HTTPClient {
         public var timeout: Timeout
         /// Upstream proxy, defaults to no proxy.
         public var proxy: Proxy?
+        /// Enables ignore NIOSSLError.uncleanShutdown when state is .head or .body, defaults to `false`.
+        public var ignoreNIOSSLUncleanShutdownError: Bool
 
-        public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+        public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreNIOSSLUncleanShutdownError: Bool = false) {
             self.tlsConfiguration = tlsConfiguration
             self.followRedirects = followRedirects
             self.timeout = timeout
             self.proxy = proxy
+            self.ignoreNIOSSLUncleanShutdownError = ignoreNIOSSLUncleanShutdownError
         }
 
-        public init(certificateVerification: CertificateVerification, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+        public init(certificateVerification: CertificateVerification, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreNIOSSLUncleanShutdownError: Bool = false) {
             self.tlsConfiguration = TLSConfiguration.forClient(certificateVerification: certificateVerification)
             self.followRedirects = followRedirects
             self.timeout = timeout
             self.proxy = proxy
+            self.ignoreNIOSSLUncleanShutdownError = ignoreNIOSSLUncleanShutdownError
         }
     }
 
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -276,8 +276,12 @@ public class HTTPClient {
         public var timeout: Timeout
         /// Upstream proxy, defaults to no proxy.
         public var proxy: Proxy?
-        /// Enables ignore NIOSSLError.uncleanShutdown when state is .head or .body, defaults to `false`.
+        /// Ignore TLS unclean shutdown error, defaults to `false`.
         public var ignoreNIOSSLUncleanShutdownError: Bool
+
+        public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+            self.init(tlsConfiguration: tlsConfiguration, followRedirects: followRedirects, timeout: timeout, proxy: proxy, ignoreNIOSSLUncleanShutdownError: false)
+        }
 
         public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreNIOSSLUncleanShutdownError: Bool = false) {
             self.tlsConfiguration = tlsConfiguration
@@ -285,6 +289,10 @@ public class HTTPClient {
             self.timeout = timeout
             self.proxy = proxy
             self.ignoreNIOSSLUncleanShutdownError = ignoreNIOSSLUncleanShutdownError
+        }
+
+        public init(certificateVerification: CertificateVerification, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+            self.init(certificateVerification: certificateVerification, followRedirects: followRedirects, timeout: timeout, proxy: proxy, ignoreNIOSSLUncleanShutdownError: false)
         }
 
         public init(certificateVerification: CertificateVerification, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreNIOSSLUncleanShutdownError: Bool = false) {
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -211,7 +211,7 @@ public class HTTPClient {
                         return channel.eventLoop.makeSucceededFuture(())
                     }
                 }.flatMap {
-                    let taskHandler = TaskHandler(task: task, delegate: delegate, redirectHandler: redirectHandler, ignoreNIOSSLUncleanShutdownError: self.configuration.ignoreNIOSSLUncleanShutdownError)
+                    let taskHandler = TaskHandler(task: task, delegate: delegate, redirectHandler: redirectHandler, ignoreUncleanSSLShutdown: self.configuration.ignoreUncleanSSLShutdown)
                     return channel.pipeline.addHandler(taskHandler)
                 }
             }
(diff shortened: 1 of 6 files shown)


Lukasa:
I am...unsure about this. This is the only place where `.uncleanShutdown` actually does communicate useful information: specifically, that we didn't know where EOF was, and the remote server didn't use TLS to tell us, so we can't tell whether the response body got truncated or not. I don't think signalling an error here is necessary the wrong thing to do.

What it may be worth doing is making this behaviour configurable, so for users who don't care they can choose to ignore this error and always treat it as a clean EOF.

vkill:
https://github.com/swift-server/async-http-client/pull/77/files#diff-bd03a2521c79f3da4bd5825512ca9979R431

We know EOF even if we ignore `.uncleanShutdown`.

Lukasa:
No we don't, we only know that when content-length is set: that is, we know it in `testWrongContentLengthWithNIOSSLUncleanShutdown`, but `testNoContentLengthWithNIOSSLUncleanShutdown` does not error, and in my view it should.

The problem is that when content-length is _not_ set, the complete response body is signalled by shutting the connection down as discussed in [RFC 7230 § 3.3.3 Message Body Length](https://tools.ietf.org/html/rfc7230#section-3.3.3):

>  7.  Otherwise, this is a response message without a declared message body length, so the message body length is determined by the number of octets received prior to the server closing the connection.

In this case, the `.uncleanShutdown` error is warning you that a network attacker could have injected a TCP FIN to forcibly truncate the response body. In this case, async-http-client cannot be confident that this has not happened, so we warn against it.

I am open to having settings put in place to allow you to ignore this error, but defaulting to reporting this error is the correct thing to do.

vkill:
@Lukasa We can configurate it now, please continue to review.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -276,19 +276,23 @@ public class HTTPClient {
         public var timeout: Timeout
         /// Upstream proxy, defaults to no proxy.
         public var proxy: Proxy?
+        /// Enables ignore NIOSSLError.uncleanShutdown when state is .head or .body, defaults to `false`.
```
Lukasa: I think we need a better description. Maybe: "Ignore TLS unclean shutdown errors".

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -276,19 +276,23 @@ public class HTTPClient {
...
+        public var ignoreNIOSSLUncleanShutdownError: Bool
 
-        public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+        public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreNIOSSLUncleanShutdownError: Bool = false) {
```
Lukasa: This change needs to be done differently, I'm afraid. This is a breaking API change, as you are allowed to refer to functions in Swift by their complete set of labels, and this changes the set of labels here. You need to do this:

```swift
public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
    self.init(tlsConfiguration: tlsConfiguration, followRedirects: followRedirects, timeout: timeout, proxy: proxy, ignoreNIOSSLUncleanShutdownError: true)
}

public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreNIOSSLUncleanShutdownError: Bool) {
    self.tlsConfiguration = tlsConfiguration
    self.followRedirects = followRedirects
    self.timeout = timeout
    self.proxy = proxy
    self.ignoreNIOSSLUncleanShutdownError = ignoreNIOSSLUncleanShutdownError
}
```

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -276,19 +276,23 @@ public class HTTPClient {
...
             self.tlsConfiguration = tlsConfiguration
             self.followRedirects = followRedirects
             self.timeout = timeout
             self.proxy = proxy
+            self.ignoreNIOSSLUncleanShutdownError = ignoreNIOSSLUncleanShutdownError
         }
 
-        public init(certificateVerification: CertificateVerification, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+        public init(certificateVerification: CertificateVerification, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreNIOSSLUncleanShutdownError: Bool = false) {
```
Lukasa: This requires a similar change as the one above: we need two functions, not one.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -604,6 +606,12 @@ internal class TaskHandler<T: HTTPClientResponseDelegate>: ChannelInboundHandler
                 /// Some HTTP Servers can 'forget' to respond with CloseNotify when client is closing connection,
                 /// this could lead to incomplete SSL shutdown. But since request is already processed, we can ignore this error.
                 break
+            case .head, .body:
+                if self.ignoreNIOSSLUncleanShutdownError {
+                    /// we can also ignore this error like .end .
+                    break
+                }
+                fallthrough
```
Lukasa: Can I propose using a `where` clause here? That would be:

```swift
case .head where self.ignoreNIOSSLUncleanShutdownError,
     .body where self.ignoreNIOSSLUncleanShutdownError:
    break
```

That avoids the need for a fall through or the internal `if`.

Sources/AsyncHTTPClient/HTTPClient.swift
vkill: It was updated.

Sources/AsyncHTTPClient/HTTPHandler.swift
vkill: Yes, it's better to use internal if. it was updated.

Sources/AsyncHTTPClient/HTTPClient.swift
vkill: It was reverted.

Sources/AsyncHTTPClient/HTTPClient.swift
vkill: It was reverted too.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -276,19 +276,31 @@ public class HTTPClient {
         public var timeout: Timeout
         /// Upstream proxy, defaults to no proxy.
         public var proxy: Proxy?
+        /// Ignore TLS unclean shutdown error, defaults to `false`.
+        public var ignoreNIOSSLUncleanShutdownError: Bool
```
tomerd: => `ignoreNIOSSLUncleanShutdownError` is a mouthful imo, maybe `ignoreUncleanSSLShutdown`? or something even shorter if someone can come up with a suggestion?

=> what are the argument for the default to be false vs. true?

Lukasa: I’m in favour of a shorter name.

False is a good default because we already ignore it in the cases where it’s 100% safe to do so: setting this switch to true would ignore it in the cases where it potentially opens your application up to some theoretical risks.

vkill: maybe `ignoreSSLUncleanShutdown`?

tomerd: `ignoreUncleanSSLShutdown` seem more natural?

artemredkin: +1 for `ignoreUncleanSSLShutdown`

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: @Lukasa we are not tagged `1.0.0` yet, do you think we can update the signature? It's unlikely imho that someone will be referencing those particular functions, wdyt?

Lukasa: Sure, if we're not maintaining a stable API we can do that.

Sources/AsyncHTTPClient/HTTPClient.swift
vkill: @tomerd It was renamed.

artemredkin:

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -346,4 +347,141 @@ class HTTPClientTests: XCTestCase {
...
+        let httpClient = HTTPClient(eventLoopGroupProvider: .createNew,
+                                    configuration: HTTPClient.Configuration(certificateVerification: .none, ignoreUncleanSSLShutdown: true))
+
+        defer {
+            try! httpClient.syncShutdown()
+            httpBin.shutdown()
+        }
+
+        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/wrongcontentlength").wait(), "Should fail") { error in
+            guard case let error = error as? HTTPParserError, case .invalidEOFState = error else {
```
Lukasa: Use `==` instead of `guard case`.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -346,4 +347,141 @@ class HTTPClientTests: XCTestCase {
...
+        let httpClient = HTTPClient(eventLoopGroupProvider: .createNew,
+                                    configuration: HTTPClient.Configuration(certificateVerification: .none))
+
+        defer {
+            try! httpClient.syncShutdown()
+            httpBin.shutdown()
+        }
+
+        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/wrongcontentlength").wait(), "Should fail") { error in
+            guard case let error = error as? NIOSSLError, case .uncleanShutdown = error else {
```
Lukasa: Use `==` instead of `guard case`.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -346,4 +347,141 @@ class HTTPClientTests: XCTestCase {
...
+        let httpClient = HTTPClient(eventLoopGroupProvider: .createNew,
+                                    configuration: HTTPClient.Configuration(certificateVerification: .none, ignoreUncleanSSLShutdown: true))
+
+        defer {
+            try! httpClient.syncShutdown()
+            httpBin.shutdown()
+        }
+
+        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/noresponse").wait(), "Should fail") { error in
+            guard case let error = error as? NIOSSLError, case .uncleanShutdown = error else {
```
Lukasa: Use `==` instead of `guard case`.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -346,4 +347,141 @@ class HTTPClientTests: XCTestCase {
...
+        let httpClient = HTTPClient(eventLoopGroupProvider: .createNew,
+                                    configuration: HTTPClient.Configuration(certificateVerification: .none))
+
+        defer {
+            try! httpClient.syncShutdown()
+            httpBin.shutdown()
+        }
+
+        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/noresponse").wait(), "Should fail") { error in
+            guard case let error = error as? NIOSSLError, case .uncleanShutdown = error else {
```
Lukasa: Use `==` instead of `guard case`.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -346,4 +347,141 @@ class HTTPClientTests: XCTestCase {
...
+        let httpClient = HTTPClient(eventLoopGroupProvider: .createNew,
+                                    configuration: HTTPClient.Configuration(certificateVerification: .none))
+
+        defer {
+            try! httpClient.syncShutdown()
+            httpBin.shutdown()
+        }
+
+        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/nocontentlength").wait(), "Should fail") { error in
+            guard case let error = error as? NIOSSLError, case .uncleanShutdown = error else {
```
Lukasa: Use `==` instead of `guard case`.

vkill:

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
vkill: Sorry, it's my mistakes. it was updated, and I have run tests in swift 5.0 .

Lukasa:

artemredkin:
@vkill thank you!

PR #81, opened 2019-08-12
Add @discardableResult to HTTPHandler setChannel method

adtrevor:
The only place where this method is currently called uses its return
value, but that's specific to the context where it's used (inside of
chained calls to map/flatMap). Future uses might not always need to use
the return value since it is the same as what's passed as the method
argument.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -403,6 +403,7 @@ extension HTTPClient {
             }
         }
 
+        @discardableResult
         func setChannel(_ channel: Channel) -> Channel {
             precondition(self.eventLoop === channel.eventLoop, "Channel must use same event loop as this task.")
             return self.lock.withLock {


adtrevor:
This solves [an issue I have in my fork where a useless warning is emitted](https://github.com/adtrevor/async-http-client/blob/pooling/Sources/AsyncHTTPClient/HTTPClient.swift#L220).

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -403,7 +403,7 @@ extension HTTPClient {
             }
         }
 
-        func setChannel(_ channel: Channel) -> Channel {
+        @discardableResult func setChannel(_ channel: Channel) -> Channel {
```
artemredkin: could you add newline after `@discardableResult`?

adtrevor: @artemredkin Sure! Wondered what was best. Fixed it.

artemredkin:

artemredkin:

PR #84, opened 2019-08-17
rename didReceivePart to didReceiveBodyPart

### LINKED ISSUE #83: should we rename `didReceivePart` to `didReceiveBodyPart` ?
https://github.com/swift-server/async-http-client/blob/e0eeb0481c0733f7dd6323916c9f4b453ed05be5/Sources/AsyncHTTPClient/HTTPHandler.swift#L322

this is a little unclear, a part of what? I'd say we should rename to `didReceiveBodyPart`

artemredkin:
fixes #83 (JW: I made this `fixes` instead of `addresses` because that's a magic word to github which will auto close #83 once this is merged).

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ class CountingDelegate: HTTPClientResponseDelegate {
         return task.eventLoop.makeSucceededFuture(())
     }
 
-    func didReceivePart(task: HTTPClient.Task<Response>, _ buffer: ByteBuffer) -> EventLoopFuture<Void> {
+    func didReceiveBodyPart(task: HTTPClient.Task<Response>, _ buffer: ByteBuffer) -> EventLoopFuture<Void> {
         // this is executed when we receive parts of the response body, could be called zero or more times
         count += buffer.readableBytes
         // in case backpressure is needed, all reads will be paused until returned future is resolved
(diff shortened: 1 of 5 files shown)


weissi:
@artemredkin thanks that looks great. I just have one proposal that we might want or not want.

The problem with this rename is that it'll break some people's code (which obviously is cool with the SemVer rules because we're `1.0.0-alpha.*` still. However, we could put a _temporary_ compatibility shim in there and _remove_ it just before tagging `1.0.0`.

What we could do is this

```swift
/// === code in async-http-client
protocol Delegate {
    func didReceiveBodyPart(_ fake: String)

    // PROBLEM: this deprecation warning will _NOT_ be shown to users implementing the protocol but rather will be shown when async-http-client calls it (see below)
    @available(*, deprecated, message: "didReceiveBodyPart(:)")
    func didReceivePart(_ fake: String)
}

extension Delegate {
    func func didReceivePart(_ fake: String) {}

    // this calls the old didReceivePart if the user didn't override the new one
    func didReceiveBodyPart(_ fake: String) {
        // PROBLEM: this will put a (temporary) compiler warning in the async-http-client code which isn't nice but maybe acceptable.
        self.didReceivePart(fake)
    }
}

// === user's code
struct Oldie: Delegate {
    func didReceivePart(_ fake: String) {
        print("did receive \(fake)")
    }
}

// this will work
Oldie().didReceiveBodyPart("body")
```

@ianpartridge / @Lukasa / @tanner0101 / @adtrevor / @t089 how important to we think it is to not break folks now?

weissi:
UPDATE: Ignore my above comment, I did a [github search](https://github.com/search?l=&q=didReceivePart+language%3ASwift&type=Code) and I couldn't find anyone implementing that. The only hits were false positives and `async-http-client` itself... So I think we can safely merge this.

PR #86, opened 2019-08-18
add response decompression support

### LINKED ISSUE #44: Support gzip/deflate decompression
It would be great if `NIOHTTPClient` could provide the machinery to decompress (`gzip` and `deflate`) response bodies.

`swift-nio-extras` provides compression support: https://github.com/apple/swift-nio-extras/blob/master/Sources/NIOHTTPCompression/HTTPResponseCompressor.swift

artemredkin:
fixes #44


Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,140 @@
...
+
+            let rc = inflate(&self, Z_NO_FLUSH)
+            precondition(rc == Z_OK || rc == Z_STREAM_END, "decompression failed: \(rc)")
+
+            return typedOutputPtr.count - Int(self.avail_out)
+        }
+    }
+}
+
+final class HTTPResponseDecompressor: ChannelDuplexHandler, RemovableChannelHandler {
```
artemredkin: @weissi do you think this should be moved to extras?

artemredkin:
@tomerd build fails with `error: 'zlib.h' file not found`, could you help with that?

artemredkin:
@weissi @Lukasa added configurable limit and more tests

Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,151 @@
...
+            self.next_in = typedDataPtr.baseAddress!
+
+            repeat {
+                var buffer = allocator.buffer(capacity: 16384)
+                self.inflatePart(to: &buffer)
+                consumer(buffer)
+            } while self.avail_out == 0
+
+            return Int(self.avail_in)
+        }
```
Lukasa: The original code has a defer block that you _must_ use here for the sake of correctness: specifically, you must unset all the pointers on the `z_stream` structure when you're done.

Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public mutating func inflatePart(to buffer: inout ByteBuffer) {
```
Lukasa: This API **must not** be public, as it is extremely unsafe. Please make it private.

Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,151 @@
...
+import NIO
+import NIOHTTP1
+
+private enum CompressionAlgorithm: String {
+    case gzip
+    case deflate
+}
+
+extension z_stream {
+    public mutating func inflatePart(input: inout ByteBuffer, allocator: ByteBufferAllocator, consumer: (ByteBuffer) -> Void) {
```
Lukasa: This API should be internal.

Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,151 @@
...
+            let typedPtr = dataPtr.baseAddress!.assumingMemoryBound(to: UInt8.self)
+            let typedDataPtr = UnsafeMutableBufferPointer(start: typedPtr, count: dataPtr.count)
+
+            self.avail_in = UInt32(typedDataPtr.count)
+            self.next_in = typedDataPtr.baseAddress!
+
+            repeat {
+                var buffer = allocator.buffer(capacity: 16384)
+                self.inflatePart(to: &buffer)
+                consumer(buffer)
```
Lukasa: It's almost always better to return a value here rather than make an outcall, as you're running a risk that you'll crash in the face of an exclusive access violation (you hold `self` mutably here, and a recursive call into this object will re-enter and crash).

Given that the `input` is passed `inout`, it may be better to document that this API will not necessarily consume all the data from `input` and may need to be called in a loop. Indeed, having the caller provide the output buffer may also be useful.

Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,151 @@
...
+
+    public mutating func inflatePart(to buffer: inout ByteBuffer) {
+        buffer.writeWithUnsafeMutableBytes { outputPtr in
+            let typedOutputPtr = UnsafeMutableBufferPointer(start: outputPtr.baseAddress!.assumingMemoryBound(to: UInt8.self), count: outputPtr.count)
+
+            self.avail_out = UInt32(typedOutputPtr.count)
+            self.next_out = typedOutputPtr.baseAddress!
+
+            let rc = inflate(&self, Z_NO_FLUSH)
+            precondition(rc == Z_OK || rc == Z_STREAM_END, "decompression failed: \(rc)")
```
Lukasa: This should probably `throw`.

Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,151 @@
...
+            context.fireChannelRead(data)
+        }
+    }
+
+    private func initializeDecoder(encoding: CompressionAlgorithm, length: Int) {
+        self.state = .compressed(encoding, length)
+
+        self.stream.zalloc = nil
+        self.stream.zfree = nil
+        self.stream.opaque = nil
```
Lukasa: Rather than explicitly nil these out, you could simply re-initialize the encoder entirely. It won't cost much more and you'll guarantee you're safe.

Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,151 @@
...
+        let window: Int32
+        switch encoding {
+        case .gzip:
+            window = 15 + 16
+        default:
+            window = 15
+        }
+
+        let rc = CNIOExtrasZlib_inflateInit2(&self.stream, window)
+        precondition(rc == Z_OK, "Unexpected return from zlib init: \(rc)")
```
Lukasa: This should probably throw.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -310,31 +316,43 @@ public class HTTPClient {
...
+                    ignoreUncleanSSLShutdown: Bool = false,
+                    decompression: Bool = false,
+                    decompressionLimit: DecompressionLimit = .ratio(100)) {
             self.tlsConfiguration = TLSConfiguration.forClient(certificateVerification: certificateVerification)
             self.followRedirects = followRedirects
             self.timeout = timeout
             self.proxy = proxy
             self.ignoreUncleanSSLShutdown = ignoreUncleanSSLShutdown
+            self.decompression = decompression
+            self.decompressionLimit = decompressionLimit
```
tomerd: should we squash `decompression` and `decompressionLimit` into one thing? make `decompression` an emum with one options being `none` (no decompression), then other options for various limits? or something along these lines?

tomerd:
> @tomerd build fails with error: 'zlib.h' file not found, could you help with that?

@artemredkin add `apt-get install libz-dev` to docker file

Sources/AsyncHTTPClient/HTTPDecompressor.swift
```diff
@@ -0,0 +1,151 @@
...
+import CNIOExtrasZlib
+import NIO
+import NIOHTTP1
+
+private enum CompressionAlgorithm: String {
+    case gzip
+    case deflate
+}
+
+extension z_stream {
```
tomerd: @weissi this code "violates" the [sswg  min requirements](https://github.com/swift-server/sswg/blob/master/process/incubation.md#minimal-requirements) due to the use of unsafe pointers

Sources/AsyncHTTPClient/HTTPDecompressor.swift
artemredkin: I was following this:
https://github.com/apple/swift-nio-extras/blob/master/Sources/NIOHTTPCompression/HTTPResponseCompressor.swift#L193 that is following this, I think:
> The zalloc, zfree, and opaque fields in the strm structure must be initialized before calling deflateInit(). Here they are set to the zlib constant Z_NULL to request that zlib use the default memory allocation routines.

Sources/AsyncHTTPClient/HTTPDecompressor.swift
artemredkin: Good catch, fixed, thanks!

Sources/AsyncHTTPClient/HTTPDecompressor.swift
artemredkin: Fixed, thanks!

Sources/AsyncHTTPClient/HTTPDecompressor.swift
artemredkin: Fixed, thanks!

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: I don't think having it as enum would be a good choice, `decompressionLimit` is a configuration of `HTTPResponseDecompressor`, but having to deal with 'no decompression' in the settings of it is a bit unlcean, imo. But I agree that having two parameters is also not a good idea (what would decompression=false and limit=ratio mean?). What about making `decompression` an enum:
```
enum HTTPDecompression {
    case none
    case enabled(limit: HTTPDecompressionLimit)
}
```
wdyt?

Sources/AsyncHTTPClient/HTTPDecompressor.swift
artemredkin: fixed, thanks!

Sources/AsyncHTTPClient/HTTPDecompressor.swift
artemredkin: fixed, thanks!

Sources/AsyncHTTPClient/HTTPDecompressor.swift
artemredkin: re-factored

Sources/AsyncHTTPClient/HTTPDecompressor.swift
tomerd: maybe the right choice here is to extract the compression logic to its own library, so its reusable and we concentrate the unsafe into one place

Sources/AsyncHTTPClient/HTTPDecompressor.swift
tomerd: see comment above. feels to me we should move the unsafe code into a separate library and  nio relevant extension into nio extras so its reusable beyond http client and the unsafe code concentrated in one place

Sources/AsyncHTTPClient/HTTPClient.swift
tomerd: yes that is exactly how i imagined it

Sources/AsyncHTTPClient/HTTPDecompressor.swift
Lukasa: I am moderately nervous about the idea of NIO being a compression library. Is it worth considering whether we should split this out into a separate SSWG space altogether?

Sources/AsyncHTTPClient/HTTPDecompressor.swift
Lukasa: FWIW, I don't think it violates the min requirements, as it is covered by the "unless interfacing with C" exemption.

Lukasa: That said, we can and should write the interface in _this_ module to avoid exposing unsafe constructs.

Sources/AsyncHTTPClient/HTTPDecompressor.swift
tomerd: exactly. sorry if my proposal ^^ was not clear enough, what i am suggesting is:
1. create a standalone compression library, sswg seems like a good home
2. have nio-extras module that helps with http compression, that utilizes that library 

seems like the most reusability. wdyt?

Lukasa: Works for me, I think it's a good fit.

artemredkin: Two questions:
1. what timeframe should we aim for? Can we postpone it until after `1.0.0` release?
2. should standalone compression library know about `ByteBuffer`? If not, what should we expose in the API?

artemredkin: @tomerd @Lukasa ^^

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: re-factored

Sources/AsyncHTTPClient/HTTPDecompressor.swift
tomerd: @artemredkin i suggest we start with moving all the compression related code to a separate internal/private module so we get the abstraction right. then we can evaluate what it means to spin off a separate library

re `ByteBuffer`, my gut feeling is a general purpose library should probably operate on native types. @Lukasa and @weissi probably probably also have insight on this

Lukasa: The general purpose library should probably follow the pattern we use elsewhere: offer an implementation that works on `Collection where Element == UInt8` and that has a branch based on [`withContiguousStorageIfAvailable`](https://developer.apple.com/documentation/swift/sequence/3128824-withcontiguousstorageifavailable) that can do less shuffling.

That will allow `ByteBufferView` to be used here, which is sufficient for good performance with NIO.

artemredkin: @Lukasa @weissi @tomerd we actually have an issue where we cannot have API for writing to a byte buffer, since we don't want to expose `Unsafe` API on the top level I propose to move all unsafe code to `nio-extras` (where we already have unsafe zlib support). That will allow us keep `async-http-client` safe and still use `ByteBuffer` in the compression/decompression APIs. PR is here (basically a copy of the current PR):
https://github.com/apple/swift-nio-extras/pull/56

artemredkin:
@tomerd @Lukasa @weissi I've updated the PR with the latests changes from `swift-nio-extras`, we'll be able to merge this after `1.3.0` is tagged (waiting for https://github.com/apple/swift-nio-extras/pull/59).

Package.swift
```diff
@@ -23,11 +23,12 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.8.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.0.0"),
+        .package(url: "https://github.com/apple/swift-nio-extras.git", .branch("master")),
```
artemredkin: I'll update the PR as soon as `1.3.0` is tagged.

artemredkin:
@weissi are you happy with the latest changes?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -403,6 +417,14 @@ public class HTTPClient {
             return EventLoopPreference(.delegateAndChannel(on: eventLoop))
         }
     }
+
+    /// Specifies decompression settings.
+    public enum Decompression {
+        /// Decompression is disabled.
+        case disabled
+        /// Decompression is enabled.
+        case enabled(limit: NIOHTTPDecompression.DecompressionLimit)
```
ianpartridge: This means that users are required to `import NIOHTTPCompression` in order to use decompression. Are people OK with that?

You can avoid this by doing:

```swift
public typealias DecompressionLimit = NIOHTTPCompression.DecompressionLimit
```

artemredkin: I don't think they are, one can just do `enabled(limit: .size(10))`, no?

ianpartridge: Oh you're right, cool! Sorry about that.

artemredkin:
@weissi ready for review

PR #90, opened 2019-08-22
refactor proxy configuration

tomerd:
motivation: make proxy configuration follow same convention as other configuration

changes:
* nest Proxy under HTTPClient.Configuration instad of top level HTTPClient
* make host and port public and mutable, following convention of other configuration objects in this library
* add some API docs

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClientProxyHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPClientProxyHandler.swift
@@ -24,11 +24,19 @@ import NIOHTTP1
 /// If a `TLSConfiguration` is used in conjunction with `HTTPClientProxy`,
 /// TLS will be established _after_ successful proxy, between your client
 /// and the destination server.
-public extension HTTPClient {
+public extension HTTPClient.Configuration {
+    /// Proxy server configuration
     struct Proxy {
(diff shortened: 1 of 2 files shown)


Sources/AsyncHTTPClient/HTTPClientProxyHandler.swift
```diff
@@ -24,11 +24,19 @@ import NIOHTTP1
 /// If a `TLSConfiguration` is used in conjunction with `HTTPClientProxy`,
 /// TLS will be established _after_ successful proxy, between your client
 /// and the destination server.
-public extension HTTPClient {
+public extension HTTPClient.Configuration {
```
ianpartridge: The comments above refer to `HTTPClientProxy` which is out of date.  Probably the comments should be on `struct Proxy` not the extension too?

tomerd: @ianpartridge see https://github.com/swift-server/async-http-client/pull/90/commits/e0eb49647e606d86d414540aeac4a21c69816e7f wdyt?

ianpartridge: Great, thanks.

PR #93, opened 2019-08-27
nest timeout configuration type inside configuration

### LINKED ISSUE #92: Revisit settings/configuration Timeout type naming
Hi there, was skimming the APIs a little bit and found the configuration type names a bit weird.

Specifically, the fields of `Configuration` using short names looks fun, but is also quite confusing imho, e.g. Timeout:

```swift
        var config = HTTPClient.Configuration()
        config.timeout = HTTPClient.Timeout() // this is not "a" Timeout, but configuration

        // a) rename? TimeoutConfiguration
        let t = HTTPClient.TimeoutConfiguration()
        // b) move? 
        let t = HTTPClient.Configuration.Timeout()
```

The types are nested in `HTTPClient` which makes it read like "this is the http client timeout", which it is not, it is just the settings what time timeouts shall be.

Looking at types:

```swift
public struct Configuration {
    public var tlsConfiguration: TLSConfiguration?
    public var followRedirects: Bool
    public var timeout: Timeout
    public var proxy: Proxy?
}
```

Timeout should likely be called `TimeoutConfiguration`; WDYT?

Proxy is likely fine, since that indeed is the value of the proxy to be used... though tbh future proof would be to make it a Configuration as well IMHO (in case it'd get per destination proxying or not etc)

artemredkin:
Resolves #92

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -290,24 +290,6 @@ public class HTTPClient {
 
     /// `HTTPClient` configuration.
     public struct Configuration {
-        /// Timeout configuration
-        public struct Timeout {
-            /// Specifies connect timeout.
-            public var connect: TimeAmount?
-            /// Specifies read timeout.
-            public var read: TimeAmount?
-
-            /// Create timeout.
-            ///
-            /// - parameters:
-            ///     - connect: `connect` timeout.
-            ///     - read: `read` timeout.
-            public init(connect: TimeAmount? = nil, read: TimeAmount? = nil) {
-                self.connect = connect
-                self.read = read
-            }
-        }
-
         /// TLS configuration, defaults to `TLSConfiguration.forClient()`.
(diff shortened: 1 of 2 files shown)


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -290,6 +290,24 @@ public class HTTPClient {
...
+            /// Create timeout.
+            ///
+            /// - parameters:
+            ///     - connect: `connect` timeout.
+            ///     - read: `read` timeout.
+            public init(connect: TimeAmount? = nil, read: TimeAmount? = nil) {
+                self.connect = connect
+                self.read = read
+            }
+        }
```
ktoso: Could be a bit cleaner to either put the type into an extension or put it next to the other types defined here -- i.e. at the bottom of `Configuration` rather than on the top? When reading config I'm first interested in what fields it has after all 🤔

artemredkin: Good idea, thank!
Does it look like what you were thinking of?

Issue #82, opened 2019-08-15
Remove eventLoop property from Task


adtrevor:
Currently, `Task` has an `eventLoop` stored property, and its `channel` must be on this same event loop.
In practice, `eventLoop` can be used by the http client delegate in the following ways:
1) Return an immediately succeeded future in the appropriate methods when they don’t want to apply back pressure.
2) Schedule backpressure related work on this event loop and return the backpressure control future.
3) In case backpressure work is done on another event loop (which might be more often the case than #2), hop the resulting backpressure future to `eventLoop`.

In these three cases, the goal to guarantee to the http client that it can assume all futures it gets from its delegate are already on the right event loop.

### Errors on user side
When applying back pressure, users getting futures from different event loops might forget to hop them back to the right event loop. On the other hand, if the client stopped relying on the assumption that futures returned from the delegate are already on the right event loop, and instead always took care of hopping them, this source of bugs would disappear entirely for a minimal performance cost (hopTo isn’t too heavy, and in many cases the user will need to use it either way, so why force them to do it when we could do this on the library side?).

### Hard to compose with future evolutions
The `eventLoop` property might not compose very well with future evolutions of the client such as pooling (tldr: the associated event loop is hard to determine synchronously like Task needs, and Task might change of event loop during its lifetime):
- It complicates the connection generation logic on the pool side. For instance a connection pool would likely contain a method such as `getConnection(for request: HTTPClient.Request -> EventLoopFuture<Connection>`, this is logical for it to return a future because a connection might not be instantly available, but it couldn’t be used as such with the current AsyncClient public API: `execute()` synchronously returns `Task`, but `Task` needs an `eventLoop` on initialization, which couldn’t be instantaneously determined due to `getConnection` being an async method. There would of be workarounds for this, but they would complicate the connection pool logic (maybe not a really good sign when two isolated components start to interact this way) and may also force creating new connections (instead of using released ones) just to satisfy the event loop identity constraint. 
- Pooling will also need a retry logic for when a connection dies too soon, this means associating the Task with a new channel that will probably not be on the same event loop as before except if we always made a new connection for retries, but it would be a ressource waste.

## Proposed solution
Remove `eventLoop` property entirely, always hop futures provided by the delegate, additionally, update `HTTPClientResponseDelegate` backpressure controlling methods to make them return `EventLoopFuture<Void>?` instead of a non-optional `EventLoopFuture<Void>`.

This change would enable the following:
[...]

PR #96, opened 2019-08-31
Tolerate futures from arbitrary event loops

### LINKED ISSUE #95: client needs to tolerate futures from arbitrary EventLoops
For back-pressure we've got a bunch of callbacks where the user can return futures to make the client continue. Those futures, at the moment, are required to be on the 'right' EventLoop but the client should definitely not require this. Instead, the client should always `hop(to:)` user futures to the correct EL if it requires it for a certain operation.

In most cases, the user will happen to return futures from the right EL anyway so `hop(to:)` is pretty much free (it fast paths `eventLoop === future.eventLoop` and just returns).

adtrevor:
This commit should fix #95 and adds associated tests

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -227,7 +227,7 @@ internal class ResponseAccumulator: HTTPClientResponseDelegate {
         case .error:
             break
         }
-        return task.eventLoop.makeSucceededFuture(())
+        return task.currentEventLoop.makeSucceededFuture(())
     }
 
     func didReceiveBodyPart(task: HTTPClient.Task<Response>, _ part: ByteBuffer) -> EventLoopFuture<Void> {
@@ -245,7 +245,7 @@ internal class ResponseAccumulator: HTTPClientResponseDelegate {
         case .error:
             break
         }
-        return task.eventLoop.makeSucceededFuture(())
+        return task.currentEventLoop.makeSucceededFuture(())
     }
 
     func didReceiveError(task: HTTPClient.Task<Response>, _ error: Error) {
@@ -343,9 +343,9 @@ extension HTTPClientResponseDelegate {
 
     public func didSendRequest(task: HTTPClient.Task<Response>) {}
 
-    public func didReceiveHead(task: HTTPClient.Task<Response>, _: HTTPResponseHead) -> EventLoopFuture<Void> { return task.eventLoop.makeSucceededFuture(()) }
+    public func didReceiveHead(task: HTTPClient.Task<Response>, _: HTTPResponseHead) -> EventLoopFuture<Void> { return task.currentEventLoop.makeSucceededFuture(()) }
 
-    public func didReceiveBodyPart(task: HTTPClient.Task<Response>, _: ByteBuffer) -> EventLoopFuture<Void> { return task.eventLoop.makeSucceededFuture(()) }
+    public func didReceiveBodyPart(task: HTTPClient.Task<Response>, _: ByteBuffer) -> EventLoopFuture<Void> { return task.currentEventLoop.makeSucceededFuture(()) }
 
     public func didReceiveError(task: HTTPClient.Task<Response>, _: Error) {}
 }
@@ -366,15 +366,23 @@ extension HTTPClient {
     /// `EventLoopFuture<Response>` of the execution or cancellation of the execution.
     public final class Task<Response> {
         /// `EventLoop` used to execute and process this request.
-        public let eventLoop: EventLoop
-        let promise: EventLoopPromise<Response>
+        public var currentEventLoop: EventLoop {
+            return self.lock.withLock {
+                _currentEventLoop
+            }
+        }
 
+        /// The stored property used by `currentEventLoop` in combination with the `lock`
+        ///
+        /// In most cases you should use `currentEventLoop` instead
+        private var _currentEventLoop: EventLoop
+        let promise: EventLoopPromise<Response>
         private var channel: Channel?
         private var cancelled: Bool
         private let lock: Lock
 
-        public init(eventLoop: EventLoop) {
-            self.eventLoop = eventLoop
+        init(eventLoop: EventLoop) {
+            self._currentEventLoop = eventLoop
             self.promise = eventLoop.makePromise()
             self.cancelled = false
             self.lock = Lock()
@@ -405,8 +413,8 @@ extension HTTPClient {
 
         @discardableResult
         func setChannel(_ channel: Channel) -> Channel {
-            precondition(self.eventLoop === channel.eventLoop, "Channel must use same event loop as this task.")
             return self.lock.withLock {
+                self._currentEventLoop = channel.eventLoop
                 self.channel = channel
                 return channel
             }
@@ -539,9 +547,11 @@ internal class TaskHandler<T: HTTPClientResponseDelegate>: ChannelInboundHandler
             } else {
                 self.state = .head
                 self.mayRead = false
-                self.delegate.didReceiveHead(task: self.task, head).whenComplete { result in
-                    self.handleBackpressureResult(context: context, result: result)
-                }
+                self.delegate.didReceiveHead(task: self.task, head)
+                    .hop(to: context.eventLoop)
+                    .whenComplete { result in
+                        self.handleBackpressureResult(context: context, result: result)
+                    }
             }
         case .body(let body):
             switch self.state {
@@ -550,9 +560,11 @@ internal class TaskHandler<T: HTTPClientResponseDelegate>: ChannelInboundHandler
             default:
                 self.state = .body
                 self.mayRead = false
-                self.delegate.didReceiveBodyPart(task: self.task, body).whenComplete { result in
-                    self.handleBackpressureResult(context: context, result: result)
-                }
+                self.delegate.didReceiveBodyPart(task: self.task, body)
+                    .hop(to: context.eventLoop)
+                    .whenComplete { result in
+                        self.handleBackpressureResult(context: context, result: result)
+                    }
             }
         case .end:
             switch self.state {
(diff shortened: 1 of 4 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -366,15 +366,15 @@ extension HTTPClient {
     /// `EventLoopFuture<Response>` of the execution or cancellation of the execution.
     public final class Task<Response> {
         /// `EventLoop` used to execute and process this request.
-        public let eventLoop: EventLoop
+        public internal(set) var currentEventLoop: EventLoop
```
adtrevor: Now that this is a variable, we must not allow mutation from the public scope.
I thought it might be the right time to set it to `currentEventLoop` as discussed in the recent comments in #82 now that this property is a variable, but I can also make the name change in another commit if preferred.
cc @weissi @artemredkin

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -366,15 +366,15 @@ extension HTTPClient {
...
         let promise: EventLoopPromise<Response>
 
         private var channel: Channel?
         private var cancelled: Bool
         private let lock: Lock
 
-        public init(eventLoop: EventLoop) {
-            self.eventLoop = eventLoop
+        init(eventLoop: EventLoop) {
```
adtrevor: Just like changing `eventLoop` to `currentEventLoop`, I can also make this change (from `public` to `internal`) in another PR if preferred although I thought it might be acceptable here since it's closely related to the issue this PR fixes.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -405,8 +405,8 @@ extension HTTPClient {
 
         @discardableResult
         func setChannel(_ channel: Channel) -> Channel {
-            precondition(self.eventLoop === channel.eventLoop, "Channel must use same event loop as this task.")
```
adtrevor: AFAICT this isn't needed anymore now that the back-pressure futures are always hopped.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -61,23 +61,19 @@ class HTTPClientTests: XCTestCase {
         let response = try httpClient.get(url: "http://localhost:\(httpBin.port)/get").wait()
         XCTAssertEqual(.ok, response.status)
     }
-
-    func testGetWithSharedEventLoopGroup() throws {
```
adtrevor: This shouldn't be needed anymore now that there isn't anymore the threading issues. This test could be re-introduced when we add a `requires` event loop preference to check that the requirements are always met.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -61,23 +61,19 @@ class HTTPClientTests: XCTestCase {
...
+    
+    func testGetWithDifferentEventLoopBackpressure() throws {
```
adtrevor: Small question about this: without the changes (hopping the futures to the right event loop) this causes a TSan failure. I was surprised it was TSan that catched the threading issue and not a more direct check made by NIO. Am I doing something wrong or is this expected?

artemredkin:

artemredkin:
@adtrevor there are formatting issues, but otherwise looks good, thank you!

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: This looks good but we need some synchronisation here because it is mutable might be accessed from different threads now. I'd suggest to depend on `NIOConcurrencyHelpers` and store it in

    private let _currentEventLoop: AtomicBox<EventLoop>
    public var currentEventLoop: EventLoop {
        return self._currentEventLoop.load()
    }
```

File: Sources/AsyncHTTPClient/HTTPHandler.swift [hunk 6]
```diff
@@ -405,8 +405,8 @@ extension HTTPClient {
...
             return self.lock.withLock {
+                self.currentEventLoop = channel.eventLoop
```
weissi: Oh I see you do lock it here which is good (that totally works as an alternative to the atomic I proposed) but you need to also lock it when reading it, not just when setting it here.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: I reckon the failure was accessing a property in HTTP client? NIO wouldn't be able to catch those. NIO can catch some regarding accessing the `context` from the wrong EventLoop (but not all). Do you have the TSan error report handy?

adtrevor:
@weissi @artemredkin Thank you for your feedback, I'm going to fix this. Looks like the test failure is due to formatting but running `swiftformat --swiftversion 5.0 Sources/AsyncHTTPClient/*.swift` on my side doesn't change any files as if everything was alright? Should I use an additional option?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adtrevor: Strange, commenting-out `hop(to:)` now causes the failure by `precondition(self.inEventLoop, file: file, line: line)`, but I guess there is some randomness and it might happen for TSan to be the first to detect the issue before NIO can even throw?

Sources/AsyncHTTPClient/HTTPHandler.swift
adtrevor: Yes, I used the lock because it was already being used by the code prior to changes.

adtrevor: Just fixed it using `lock` since it was previously used, but it `AtomicBox` can make things clearer I could also use it instead. Tell me what you think :)

adtrevor:
Issues should now be fixed, just waiting to know what you think of the latest changes :). Formatting issues are also probably fixed now.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -366,15 +366,23 @@ extension HTTPClient {
     /// `EventLoopFuture<Response>` of the execution or cancellation of the execution.
     public final class Task<Response> {
         /// `EventLoop` used to execute and process this request.
-        public let eventLoop: EventLoop
-        let promise: EventLoopPromise<Response>
+        public var currentEventLoop: EventLoop {
+            self.lock.withLock {
```
weissi: you'll need a `return` here, omitting the `return` in properties is a Swift 5.1 feature, sorry

weissi:

weissi:

Sources/AsyncHTTPClient/HTTPHandler.swift
adtrevor: Oops, always keep forgetting about this one. Thank you, it is fixed now.

adtrevor:
@weissi 
> would you mind creating one new unit test that deliberately returns futures from the wrong eventloop just to check everything still works?
The `testGetWithDifferentEventLoopBackpressure` already tests this using `TestHTTPDelegate` which I modified to allow initialising it with `TestHTTPDelegate(backpressureEventLoop: external.next())` where `external` is a completely different event loop group! :)

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: ok, cool, thanks!

weissi: @adtrevor yes, it's possible that TSan finds it before NIO

weissi: the important thing is that for any `hop(to:)` there's at least one unit test that fails (even without TSan) if the `hop(to:)` is commented out.

weissi:
@artemredkin how does this look to you?

@tanner0101 / @ianpartridge / @ldewailly / @tomerd  any input on the API change?

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: Why do we need it to be `var`? I somehow missed it

adtrevor: It's because of `setChannel(...)`: when it updates the channel, it must also update the current event loop at the same time.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -366,15 +366,23 @@ extension HTTPClient {
...
+        public var currentEventLoop: EventLoop {
+            return self.lock.withLock {
+                _currentEventLoop
+            }
+        }
 
+        /// The stored property used by `currentEventLoop` in combination with the `lock`
+        ///
+        /// In most cases you should use `currentEventLoop` instead
+        var _currentEventLoop: EventLoop
```
t089: Hey, is this internal read/write access for this underscored var necessary? Couldn't it be `private`?

adtrevor: @t089 Thank you for noticing it. If no one sees the need for a unit test to ever access this directly then I can make it `private`.

weissi: @t089 / @adtrevor yeah, it should be `private`, the `_` is just because there's a `public var currentEventLoop` but the access to that needs to be synchronised. The actual storage is `_currentEventLoop`

adtrevor: Ok, thank you @t089, this is fixed now!

adtrevor:
All issues should now be fixed and if no one has additional comments maybe this can now be merged? cc @weissi @artemredkin

weissi:

weissi:

PR #99, opened 2019-09-06
give generic parameters nice names

weissi:
Motivation:

There's no benefit in all generic parameters being named `T`, in fact I'd argue it can be quite confusing.

Modifications:

Renamed `T` to `Delegate` and `Response` which is more readable I think.

Result:

More readable code & API.


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -681,7 +681,7 @@ internal struct RedirectHandler<T> {
         return url.absoluteURL
     }
 
-    func redirect(status: HTTPResponseStatus, to redirectURL: URL, promise: EventLoopPromise<T>) {
+    func redirect(status: HTTPResponseStatus, to redirectURL: URL, promise: EventLoopPromise<Response>) {
```
ianpartridge: Not sure this one is a good idea as we already have `public struct Response` so it's not clear at a glance that this is referring to a generic type.

weissi: we're already using `<Response>` in most places as a generic parameter:

https://github.com/swift-server/async-http-client/blob/master/Sources/AsyncHTTPClient/HTTPClient.swift#L165

weissi: `public struct Reponse` is nested in `HTTPClient` as `HTTPClient.Response` so there is a way to disambiguate and the ambiguity is restricted to the `HTTPClient` implementation itself so I thought it's fine.

But if we don't think it is, then we should find a new name and use it everywhere. Can always do `<ResponseType>`

weissi: @ianpartridge just renamed it to `<ResponseType>` for now

weissi: it was on an internal type anyway so the name `<ResponseType>` isn't actually API so it doesn't need to be consistent with other bits of the API either.

PR #102, opened 2019-09-11
EventLoop preference overhaul

### LINKED ISSUE #100: programming model questions
Right now, the delegate's methods are calling out on essentially a random `EventLoop`.
I _think_ we have the guarantee, that each delegate method will at least be called on the same `EventLoop`?

I realise that I proposed the change to have `task.currentEventLoop` (which can change) but I'm not sure anymore that was a good idea. I didn't quite realise that this means that the user has no idea on what `EventLoop` their delegate's methods will be called on.

Maybe it would be better to go back to having a fixed `EventLoop` on the task and hop over in case we get a Channel from a different `EventLoop` from a connection pool.

The other dimension where this matters is the `EventLoopPreference`, that also kind of means a random `EventLoop`. Maybe we should just have the user specify the `EventLoop` they want (if they care, random one otherwise) and use that. If we get a `Channel` from a different `EventLoop` then let's just `hop(to:)` the correct one.

This is just an idea really but I was thinking about the programming model earlier today and 'essentially a random EventLoop' sounds like it might lead to a lot of threading bugs?

Tagging the relevant people @adtrevor / @ianpartridge / @tanner0101 / @Lukasa / @artemredkin

weissi:
Motivation:

With #96 we changed `HTTPClient.Task`'s `eventLoop` property to be `currentEventLoop` because we anticipated that it might change because of the connection pool work that's ongoing.
That however made the programming model rather hard because the user didn't get any guarantee anymore on which `EventLoop` the delegate would run.
With this patch, we're bringing `eventLoop` back which is the `EventLoop` the delegate will run on. Note that it's not necessarily the `EventLoop` the `Channel` will run on. If the user requires those to be the same, the can request that with `EventLoopPreference.delegateAndChannel(on: eventLoop)` rather than `.indifferent` or `.delegate(on: eventLoop)`

Modification:

Let the user select between:

- `.indifferent`: I don't care about the `EventLoop`
- `.delegate(on: eventLoop)`: I want the delegate to be called on `eventLoop` and I would _prefer_ the `Channel` to be on `eventLoop` too but I'm happy if not
- `.delegateAndChannel(on: eventLoop)`: I want `Channel` and delegate to live on `eventLoop`

Result:

- Happier users because the programming model is easier
- fixes #100

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -205,28 +205,42 @@ public class HTTPClient {
         switch eventLoop.preference {
         case .indifferent:
             return self.execute(request: request, delegate: delegate, eventLoop: self.eventLoopGroup.next(), deadline: deadline)
-        case .prefers(let preferred):
-            precondition(self.eventLoopGroup.makeIterator().contains { $0 === preferred }, "Provided EventLoop must be part of clients EventLoopGroup.")
-            return self.execute(request: request, delegate: delegate, eventLoop: preferred, deadline: deadline)
+        case .delegate(on: let eventLoop):
+            precondition(self.eventLoopGroup.makeIterator().contains { $0 === eventLoop }, "Provided EventLoop must be part of clients EventLoopGroup.")
+            return self.execute(request: request, delegate: delegate, eventLoop: eventLoop, deadline: deadline)
+        case .delegateAndChannel(on: let eventLoop):
+            precondition(self.eventLoopGroup.makeIterator().contains { $0 === eventLoop }, "Provided EventLoop must be part of clients EventLoopGroup.")
+            return self.execute(request: request, delegate: delegate, eventLoop: eventLoop, deadline: deadline)
+        case .testOnly_exact(channelOn: let channelEL, delegateOn: let delegateEL):
+            return self.execute(request: request,
+                                delegate: delegate,
+                                eventLoop: delegateEL,
+                                channelEL: channelEL,
+                                deadline: deadline)
         }
     }
 
     private func execute<Delegate: HTTPClientResponseDelegate>(request: Request,
                                                                delegate: Delegate,
-                                                               eventLoop: EventLoop,
+                                                               eventLoop delegateEL: EventLoop,
+                                                               channelEL: EventLoop? = nil,
                                                                deadline: NIODeadline? = nil) -> Task<Delegate.Response> {
         let redirectHandler: RedirectHandler<Delegate.Response>?
         if self.configuration.followRedirects {
             redirectHandler = RedirectHandler<Delegate.Response>(request: request) { newRequest in
-                self.execute(request: newRequest, delegate: delegate, eventLoop: eventLoop, deadline: deadline)
+                self.execute(request: newRequest,
+                             delegate: delegate,
+                             eventLoop: delegateEL,
+                             channelEL: channelEL,
+                             deadline: deadline)
             }
         } else {
             redirectHandler = nil
         }
 
-        let task = Task<Delegate.Response>(eventLoop: eventLoop)
+        let task = Task<Delegate.Response>(eventLoop: delegateEL)
 
-        var bootstrap = ClientBootstrap(group: eventLoop)
+        var bootstrap = ClientBootstrap(group: channelEL ?? delegateEL)
             .channelOption(ChannelOptions.socket(SocketOptionLevel(IPPROTO_TCP), TCP_NODELAY), value: 1)
             .channelInitializer { channel in
                 let encoder = HTTPRequestEncoder()
@@ -262,9 +276,7 @@ public class HTTPClient {
             .flatMap { channel in
                 channel.writeAndFlush(request)
             }
-            .whenFailure { error in
-                task.fail(error)
-            }
+            .cascadeFailure(to: task.promise)
 
         return task
     }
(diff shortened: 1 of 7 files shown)


weissi:
CC @adtrevor

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -400,24 +404,21 @@ extension HTTPClient {
...
-                _currentEventLoop
-            }
+            return self.eventLoop
         }
 
-        /// The stored property used by `currentEventLoop` in combination with the `lock`
-        ///
-        /// In most cases you should use `currentEventLoop` instead
-        private var _currentEventLoop: EventLoop
+        /// The `EventLoop` the delegate will be executed on.
```
adtrevor: Maybe there should also be a `public` accessible property to the `channel`'s `EventLoop`?

weissi: What would the use-case be?

weissi: Happy to add now or later if we feel there’s a compelling-enough reason

adtrevor: I was thinking about the backpressure futures of the delegate: performance wise, it's best to return one from the same event loop as the channel, which wouldn't be possible anymore if there is no access to the channel event loop.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -363,9 +379,28 @@ public class HTTPClient {
 
         /// Event Loop will be selected by the library.
         public static let indifferent = EventLoopPreference(.indifferent)
+
         /// Library will try to use provided event loop if possible.
+        @available(*, deprecated, renamed: "delegate(on:)")
```
artemredkin: do we really need this?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -400,24 +404,21 @@ extension HTTPClient {
     /// Response execution context. Will be created by the library and could be used for obtaining
     /// `EventLoopFuture<Response>` of the execution or cancellation of the execution.
     public final class Task<Response> {
-        /// `EventLoop` used to execute and process this request.
+        @available(*, deprecated, renamed: "eventLoop")
```
artemredkin: same here

Sources/AsyncHTTPClient/HTTPClient.swift
weissi: @artemredkin so far, we've broken everybody in every new alpha release which is a real issue. I think we should do this and remove before 1.0.0. I'll file an issue to remove all deprecated stuff just before then 1.0.0 release

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: filed https://github.com/swift-server/async-http-client/issues/108 to remove them just before

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: I'm inclined to say that if that really matters much, the user should've chosen `.delegateAndChannel(on: ...)` and therefore the channel's EL == delegate's EL. @Lukasa what do you think?

adtrevor: The issue I see is that `.delegateAndChannel(on: ...)` is really restrictive for the connection pool and some delegates that don't need to do back-pressure (eg: `ResponseAccumulator`) would need to use it to avoid performance loss even if the `delegate`/`channel` EL doesn't matter for them.

Lukasa: I think it’s reasonable to be able to ask for the delegate loop.

adtrevor: @Lukasa The delegate `EventLoop` should already be accessible with the `eventLoop`, but I was thinking about the channel `EventLoop`.

To give a more concrete example of what worries me, if you look at what happens in the `ResponseAccumulator`, the newest changes are doing this:
```swift
public func didReceiveHead(task: HTTPClient.Task<Response>, _: HTTPResponseHead) -> EventLoopFuture<Void> {
    return task.eventLoop.makeSucceededFuture(())
}
```

But this means that in this part:
```
self.delegate.didReceiveBodyPart(task: self.task, body)
                    .hop(to: context.eventLoop)
```
The cost of hopping will be quite big especially something that's going to be called so much often.

adtrevor: On the other hand if letting public access to the channel event loop risks introducing too many errors on lib user side, maybe we could actually allow returning `nil` for the delegates that don't need backpressure?

Lukasa: Yeah, I don't think it's worth exposing the channel event loop at this time. I understand the concern, and I think we can address it in the future, but for now I don't think I'd really bother. The performance concern is not really large enough (IMO) to justify it right now.

weissi: @adtrevor if you really need that kind of performance, I'd say just use `.delegateAndChannel(on: ...)` and then the delegate and the Channel's EL are guaranteed to be the same. If we truly find that there are important use-cases to get the Channel's El we can, as Cory points out, add that at any point without breaking API.

Sources/AsyncHTTPClient/HTTPClient.swift
weissi: @artemredkin happy with that?

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: @artemredkin happy with #108 ?

artemredkin: sure, thanks!

weissi:
@artemredkin I fixed the merge conflicts now.

artemredkin:
Do you want to merge of wait for @Lukasa's approval?

PR #104, opened 2019-09-13
Remove spaces in range operators

adtrevor:
This is a question of preference, but it looks like the most commonly used format for ranges in Swift projects is what can [be found in the Swift Book](https://docs.swift.org/swift-book/LanguageGuide/BasicOperators.html) where there is no space between the values and range operators, for instance: `case 200..<300:` contrary to the current `case 200 ..< 300`.

What do you think? 🙂

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -8,5 +8,6 @@
 --patternlet inline
 --stripunusedargs unnamed-only
 --comments ignore
+--ranges nospace
 
 # rules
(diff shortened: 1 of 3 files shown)


tomerd:
no strong opinion here

adtrevor:
@ktoso Makes sense too, although personally I would prefer parenthesising or disambiguating with `TypeName.foo`.

artemredkin:
thanks, @adtrevor !

PR #107, opened 2019-09-16
preserve trailing slash in uri path

### LINKED ISSUE #101: URL.uri removes trailing slash

artemredkin:
fixes #101 by preserving trailing slash in path return by the `URL's` `path` property


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -385,14 +385,42 @@ extension HTTPClientResponseDelegate {
...
+        let url = self.absoluteString
+
+        var pathEndIndex = url.index(before: url.endIndex)
+        if let queryIndex = url.firstIndex(of: "?") {
+            pathEndIndex = url.index(before: queryIndex)
+        } else if let fragmentIndex = url.firstIndex(of: "#") {
+            pathEndIndex = url.index(before: fragmentIndex)
+        }
+
+        return url[pathEndIndex] == "/"
```
Lukasa: Writing this against the absolute string seems to be making life quite hard: why not write it against `self.path`?

artemredkin: `self.path` removes trailing slash

artemredkin:
If a server has a redirect like `path -> path/`, it is impossible to query it right now, we will be redirected to `path/` on every request, but new request will still be `path` since we lose trailing slash.

Lukasa:
It's a shame that server is broken. ☹️

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -394,16 +394,20 @@ extension URL {
...
-            pathEndIndex = url.index(before: queryIndex)
-        } else if let fragmentIndex = url.firstIndex(of: "#") {
-            pathEndIndex = url.index(before: fragmentIndex)
-        }
+            var pathEndIndex = url.index(before: url.endIndex)
+            if let queryIndex = url.firstIndex(of: "?") {
+                pathEndIndex = url.index(before: queryIndex)
+            } else if let fragmentIndex = url.suffix(from: url.firstIndex(of: "@") ?? url.startIndex).lastIndex(of: "#") {
+                pathEndIndex = url.index(before: fragmentIndex)
+            }
```
Lukasa: Do we need the `else` branch? The minimum versions above were all released in 2015, making them 4 years old this year. Can we get away without it?

If the answer is really no, then can we also add a comment to this code branch to say that the code is not 100% correct, but is considered "good enough"?

artemredkin: What do you mean by getting away without it? Adding supported platforms in `Package.swift` or a precondition? I would prefer not to have an alternative clause here as well...
@tomerd @tanner0101 @ianpartridge do you think we should limit the client to modern(er) platforms?

tomerd: given our main use case is linux, restricting older non-linux platforms in Package.swift seems fine to me

artemredkin:
@Lukasa fixed path check by using `hasDirectoryPath` instead of custom function

Package.swift
```diff
@@ -17,6 +17,7 @@ import PackageDescription
 
 let package = Package(
     name: "async-http-client",
+    platforms: [.macOS(.v10_11), .iOS(.v9), .tvOS(.v9), .watchOS(.v2)],
```
tomerd: would this potentially confuse folks to think linux is not supported? maybe worth adding a comment?

artemredkin: added a comment!

weissi: I think we need to take this out. If we don't then _every dependency_ needs to mirror this stanza.

artemredkin:
Agreed, reverted

PR #110, opened 2019-09-23
cancel should be done with outbound user event

### LINKED ISSUE #97: cancel should be done with an outbound user event
https://github.com/swift-server/async-http-client/blob/e9d0dd2cc99eb886416eb71205c2768bfacaf675/Sources/AsyncHTTPClient/HTTPHandler.swift#L398-L402
the above code would be better done with an outbound user event, so something like

```
channel.triggerOutboundUserEvent(CancelEvent)
```

as an added benefit, that returns a promise so we can know if that was successful or not. The promise will be failed automatically if that has been handled before or the channel is already closed.

artemredkin:
fixes #97

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -734,12 +734,19 @@ extension TaskHandler: ChannelDuplexHandler {
             self.state = .end
             let error = HTTPClientError.readTimeout
             self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
-        } else if (event as? TaskCancelEvent) != nil {
+        } else {
+            context.fireUserInboundEventTriggered(event)
+        }
+    }
+
+    func triggerUserOutboundEvent(context: ChannelHandlerContext, event: Any, promise: EventLoopPromise<Void>?) {
+        if (event as? TaskCancelEvent) != nil {
             self.state = .end
             let error = HTTPClientError.cancelled
             self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
+            promise?.succeed(())
         } else {
-            context.fireUserInboundEventTriggered(event)
+            context.triggerUserOutboundEvent(event, promise: promise)
         }
     }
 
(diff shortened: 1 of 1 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -476,7 +476,7 @@ extension HTTPClient {
             self.lock.withLock {
                 if !cancelled {
                     cancelled = true
-                    channel?.pipeline.fireUserInboundEventTriggered(TaskCancelEvent())
+                    channel?.triggerUserOutboundEvent(TaskCancelEvent(), promise: nil)
```
artemredkin: I think since cancellation is not guaranteed to do anything (it's best effort only), we can skip any errors, wdyt?

weissi: @artemredkin firing through the `ChannelPipeline` should come out of the lock. So something like

```
let channel = self.lock.withLock {
    if !self.cancelled {
        self.cancelled = true
        return self.channel
    }
 }
channel?.triggerUserOutboundEvent(...)
```

Regarding ignoring the promise: I'm okay with that.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -473,12 +473,14 @@ extension HTTPClient {
...
-            self.lock.withLock {
+            let channel: Channel? = self.lock.withLock {
                 if !cancelled {
                     cancelled = true
-                    channel?.pipeline.fireUserInboundEventTriggered(TaskCancelEvent())
+                    return self.channel
                 }
+                return nil
             }
+            channel?.triggerUserOutboundEvent(TaskCancelEvent(), promise: nil)
```
weissi: @artemredkin I'm pretty sure this doesn't actually work (which proves there's no test) because whatever ChannelHandler is supposed to catch this event needs to change from inbound to outbound user event.

weissi: we need to add a test that verifies cancellation works.

artemredkin: fixed! and tests pass now

PR #113, opened 2019-09-30
add support for redirect limits

### LINKED ISSUE #76: Stack overflow in promise chain when performing web scraping.
I've written a little scraper which can use both the foundation URL class and the HTTPClient classes.  Requests go through many different proxies pulled from proxy lists on the web.

The HTTPClient classes are failing.
Do you want bugs yet, or should I check back in a couple months?

artemredkin:
This PR adds support for 2 types of redirect limits - one that detects loops (by keeping history of all visited URL's for a single request), and one that just counts the number of redirects.

fixes #76

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -226,8 +226,8 @@ public class HTTPClient {
                                                                channelEL: EventLoop? = nil,
                                                                deadline: NIODeadline? = nil) -> Task<Delegate.Response> {
         let redirectHandler: RedirectHandler<Delegate.Response>?
-        switch self.configuration.redirects {
-        case .allow(let max, let allowCycles):
+        switch self.configuration.redirectConfiguration.configuration {
+        case .follow(let max, let allowCycles):
             var request = request
             if request.redirectState == nil {
                 request.redirectState = .init(count: max, visited: allowCycles ? nil : Set())
@@ -322,7 +322,7 @@ public class HTTPClient {
         ///  - `305: Use Proxy`
         ///  - `307: Temporary Redirect`
         ///  - `308: Permanent Redirect`
-        public var redirects: RedirectPolicy
+        public var redirectConfiguration: RedirectConfiguration
         /// Default client timeout, defaults to no timeouts.
         public var timeout: Timeout
         /// Upstream proxy, defaults to no proxy.
@@ -330,25 +330,25 @@ public class HTTPClient {
         /// Ignore TLS unclean shutdown error, defaults to `false`.
         public var ignoreUncleanSSLShutdown: Bool
 
-        public init(tlsConfiguration: TLSConfiguration? = nil, redirects: RedirectPolicy = .allow(max: 5, allowCycles: false), timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
-            self.init(tlsConfiguration: tlsConfiguration, redirects: redirects, timeout: timeout, proxy: proxy, ignoreUncleanSSLShutdown: false)
+        public init(tlsConfiguration: TLSConfiguration? = nil, redirectConfiguration: RedirectConfiguration? = nil, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+            self.init(tlsConfiguration: tlsConfiguration, redirectConfiguration: redirectConfiguration, timeout: timeout, proxy: proxy, ignoreUncleanSSLShutdown: false)
         }
 
-        public init(tlsConfiguration: TLSConfiguration? = nil, redirects: RedirectPolicy = .allow(max: 5, allowCycles: false), timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreUncleanSSLShutdown: Bool = false) {
+        public init(tlsConfiguration: TLSConfiguration? = nil, redirectConfiguration: RedirectConfiguration? = nil, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreUncleanSSLShutdown: Bool = false) {
             self.tlsConfiguration = tlsConfiguration
-            self.redirects = redirects
+            self.redirectConfiguration = redirectConfiguration ?? RedirectConfiguration()
             self.timeout = timeout
             self.proxy = proxy
             self.ignoreUncleanSSLShutdown = ignoreUncleanSSLShutdown
         }
 
-        public init(certificateVerification: CertificateVerification, redirects: RedirectPolicy = .allow(max: 5, allowCycles: false), timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
-            self.init(certificateVerification: certificateVerification, redirects: redirects, timeout: timeout, proxy: proxy, ignoreUncleanSSLShutdown: false)
+        public init(certificateVerification: CertificateVerification, redirectConfiguration: RedirectConfiguration? = nil, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+            self.init(certificateVerification: certificateVerification, redirectConfiguration: redirectConfiguration, timeout: timeout, proxy: proxy, ignoreUncleanSSLShutdown: false)
         }
 
-        public init(certificateVerification: CertificateVerification, redirects: RedirectPolicy = .allow(max: 5, allowCycles: false), timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreUncleanSSLShutdown: Bool = false) {
+        public init(certificateVerification: CertificateVerification, redirectConfiguration: RedirectConfiguration? = nil, timeout: Timeout = Timeout(), proxy: Proxy? = nil, ignoreUncleanSSLShutdown: Bool = false) {
             self.tlsConfiguration = TLSConfiguration.forClient(certificateVerification: certificateVerification)
-            self.redirects = redirects
+            self.redirectConfiguration = redirectConfiguration ?? RedirectConfiguration()
             self.timeout = timeout
             self.proxy = proxy
             self.ignoreUncleanSSLShutdown = ignoreUncleanSSLShutdown
@@ -430,9 +430,27 @@ extension HTTPClient.Configuration {
     }
 
     /// Specifies redirect processing settings.
-    public enum RedirectPolicy {
+    public struct RedirectConfiguration {
+        enum Configuration {
+            /// Redirects are not followed.
+            case disallow
+            /// Redirects are followed with a specified limit.
+            case follow(max: Int, allowCycles: Bool)
+        }
+
+        var configuration: Configuration
+
+        init() {
+            self.configuration = .follow(max: 5, allowCycles: false)
+        }
+
+        init(configuration: Configuration) {
+            self.configuration = configuration
+        }
+
         /// Redirects are not followed.
-        case disallow
+        public static let disallow = RedirectConfiguration(configuration: .disallow)
+
         /// Redirects are followed with a specified limit.
         ///
         /// - parameters:
@@ -440,7 +458,7 @@ extension HTTPClient.Configuration {
         ///     - allowCycles: Whether cycles are allowed.
         ///
         /// - warning: Cycle detection will keep all visited URLs in memory which means a malicious server could use this as a denial-of-service vector.
-        case allow(max: Int, allowCycles: Bool)
+        public static func follow(max: Int, allowCycles: Bool) -> RedirectConfiguration { return .init(configuration: .follow(max: max, allowCycles: allowCycles)) }
     }
 }
 
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -221,22 +221,24 @@ public class HTTPClient {
     }
 
     private func execute<Delegate: HTTPClientResponseDelegate>(request: Request,
-                                                               redirectLimit: Configuration.RedirectLimit.Next? = nil,
+                                                               redirectState: Configuration.RedirectPolicy.State? = nil,
                                                                delegate: Delegate,
                                                                eventLoop delegateEL: EventLoop,
                                                                channelEL: EventLoop? = nil,
                                                                deadline: NIODeadline? = nil) -> Task<Delegate.Response> {
         let redirectHandler: RedirectHandler<Delegate.Response>?
-        if let next = redirectLimit ?? self.configuration.followRedirects.next {
-            redirectHandler = RedirectHandler<Delegate.Response>(limit: next, request: request) { newRequest, limit in
+        switch self.configuration.redirects {
+        case .allow(let max, let allowCycles):
+            let state = redirectState ?? .init(count: max, visited: allowCycles ? nil : Set())
+            redirectHandler = RedirectHandler<Delegate.Response>(state: state, request: request) { newRequest, state in
                 self.execute(request: newRequest,
-                             redirectLimit: limit,
+                             redirectState: state,
                              delegate: delegate,
                              eventLoop: delegateEL,
                              channelEL: channelEL,
                              deadline: deadline)
             }
-        } else {
+        case .disallow:
             redirectHandler = nil
         }
 
@@ -319,7 +321,7 @@ public class HTTPClient {
         ///  - `305: Use Proxy`
         ///  - `307: Temporary Redirect`
         ///  - `308: Permanent Redirect`
-        public var followRedirects: FollowRedirects
+        public var redirects: RedirectPolicy
         /// Default client timeout, defaults to no timeouts.
         public var timeout: Timeout
         /// Upstream proxy, defaults to no proxy.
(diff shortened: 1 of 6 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -818,6 +819,24 @@ internal struct RedirectHandler<ResponseType> {
...
+        let nextLimit: HTTPClient.Configuration.RedirectLimit.Next
+        switch self.limit {
+        case .none:
+            nextLimit = .none
+        case .count(let left):
+            if left == 0 {
+                return promise.fail(HTTPClientError.redirectLimitReached)
+            }
+            nextLimit = .count(left: left - 1)
+        case .loop(let visited):
```
tomerd: would it make sense to always fail on loops, rather than make it a configuration option?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -317,31 +319,31 @@ public class HTTPClient {
...
+        public var followRedirects: FollowRedirects
         /// Default client timeout, defaults to no timeouts.
         public var timeout: Timeout
         /// Upstream proxy, defaults to no proxy.
         public var proxy: Proxy?
         /// Ignore TLS unclean shutdown error, defaults to `false`.
         public var ignoreUncleanSSLShutdown: Bool
 
-        public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
+        public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: FollowRedirects = .disabled, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
```
tomerd: should following redirect be default?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -818,6 +819,24 @@ internal struct RedirectHandler<ResponseType> {
     }
 
     func redirect(status: HTTPResponseStatus, to redirectURL: URL, promise: EventLoopPromise<ResponseType>) {
+        let nextLimit: HTTPClient.Configuration.RedirectLimit.Next
+        switch self.limit {
+        case .none:
+            nextLimit = .none
+        case .count(let left):
+            if left == 0 {
```
tomerd: guard?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -221,14 +221,16 @@ public class HTTPClient {
     }
 
     private func execute<Delegate: HTTPClientResponseDelegate>(request: Request,
+                                                               redirectLimit: Configuration.RedirectLimit.Next? = nil,
```
tomerd: another option to consider is to make the limit part of the request object (internally)

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: I think in server-side setting it makes sense to have defaults to be predictable, and in case of redirects we have potentially unexpected behaviour, but it's not a strong opinion. @Lukasa @weissi what do you think?

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: good idea, thanks, fixed

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: interesting idea, I made it an option because it requires additional memory to store all visited URL, do you think this is a premature optimization?

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: I did try that option as well, but current implementation seemed a bit better, would you like me to commit alternative version so we can compare?

tomerd: sure if you are up to it. feels better encapsulated on the surface

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -423,6 +425,41 @@ extension HTTPClient.Configuration {
...
+        /// No redirect limit.
+        public static let none = RedirectLimit(next: .none)
+        /// Request execution will be stopped if redirect loop is detected.
+        public static let detectLoop = RedirectLimit(next: .loop(visited: Set()))
+        /// Specifies maximum number of redirects for a single request.
+        public static func count(_ value: Int) -> RedirectLimit { return RedirectLimit(next: .count(left: value)) }
+    }
+
+    /// Specifies redirect processing settings.
+    public enum FollowRedirects {
```
ianpartridge: This public API seems slightly confused, to me... I would approach it as:

```swift
public enum RedirectionPolicy {
    /// Redirects are not allowed/followed.
    case disallow

    /// Redirects are allowed.
    /// limit: The maximum number of redirects allowed.
    /// detectCycles: If true, detect cycles and cancel redirection if detected.
    case allow(limit: Limit, detectCycles: Bool)

    public enum Limit {
        /// Unlimited redirection
        case unlimited

        /// A limited number redirects are allowed
        case limited(count: Int)
    }
}
```

Then, the `Set<URL>` and `count` state would be maintained elsewhere, e.g. a separate `RedirectionState` struct on the request.

In this way we can allow people to set a maximum number of redirects, and also detect cyles. I don't think that is possible in the current design?

Sources/AsyncHTTPClient/HTTPClient.swift
weissi: I would usually err on the side of 'just works' and make it redirect with a limit of 5 or something.

Regarding the parameter: `followRedirects = .disabled` doesn't read as good as `followRedirects: .no`, right?

Sources/AsyncHTTPClient/HTTPClient.swift
weissi: I think we should also to the docs that the `Set<URL>` case can balloon arbitrary amounts of data into memory so can be used as a DOS vector.

tomerd: feels like we should always set some limit here, both on count and on recursion

tomerd: similar approach to API as one from @ianpartridge

```
public enum FollowRedirects {
    /// Redirects are not followed.
    case no

    /// Redirects are followed.
    /// max: The number of redirects allowed.
    /// allowCycles: If false, detect cycles and cancel redirection if detected.
    case yes(max: Int = 50, allowCycles: Bool = false)
}
```

Sources/AsyncHTTPClient/HTTPClient.swift
Lukasa: Agreed with @weissi

Sources/AsyncHTTPClient/HTTPClient.swift
ianpartridge: My slight concern with naming the type `FollowRedirects` is that it reads like a boolean and I tend to prefer "...Strategy" or "...Policy" for this kind of enum. No big deal though.

tomerd: @ianpartridge not a strong opinion on my side

artemredkin: Great ideas, thanks! I decided to try to incorporate both suggestions - renamed FollowRedirects to RedirectPolicy and `.allow` now has both limits. wdyt?

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: Loop detection is now default

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: following redirects is now default with loop detection

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: Also, to prevent possible DOS, cycle detection is cut off by max allowed redirects (either cycle will be found or max allowed will fail)

tomerd: looks better imo. i still feel the redirect state really belongs with the request object (as private extension) rather than as peer to it

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -423,6 +427,19 @@ extension HTTPClient.Configuration {
             self.read = read
         }
     }
+
+    /// Specifies redirect processing settings.
+    public enum RedirectPolicy {
+        /// Redirects are not followed.
+        case disallow
+        /// Redirects are followed with a specified limit. Cycle detection reqiures that all visited URL's are kept in memory.
+        case allow(max: Int, allowCycles: Bool)
```
tomerd: not: "disabled" and "enabled" seem more natural than "disallow" and "allow"? since this is not some security policy?

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: I will do it next, yes

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: this struct was renamed to `Policy`, do you think `RedirectPolicy.enable` is better than `RedirectPolicy.allow`? I don't have a strong opinion here, wdyt? cc @ianpartridge

ianpartridge: In the tests it reads as `redirects: .allow(max: 5, allowCycles: false)` which seems fine (although I would prefer `redirection:` instead of `redirects:`). `.allow` makes sense to me because redirection is something that is requested by the remote HTTP server via an HTTP return code, and we decide whether to allow it or not.

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: I've updated the PR with redirect state in request

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: Moved redirect state to request

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -221,20 +221,24 @@ public class HTTPClient {
     }
 
     private func execute<Delegate: HTTPClientResponseDelegate>(request: Request,
+                                                               redirectState: Configuration.RedirectPolicy.State? = nil,
```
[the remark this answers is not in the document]
artemredkin: good idea, done

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -423,6 +427,19 @@ extension HTTPClient.Configuration {
             self.read = read
         }
     }
+
+    /// Specifies redirect processing settings.
+    public enum RedirectPolicy {
```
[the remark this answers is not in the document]
artemredkin: +1 from me, @tomerd wdyt?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -472,6 +489,7 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         case invalidProxyResponse
         case contentLengthMissing
         case proxyAuthenticationRequired
+        case redirectLimitReached
```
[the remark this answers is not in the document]
artemredkin: great idea, fixed, thanks!

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -423,6 +428,14 @@ extension HTTPClient.Configuration {
             self.read = read
         }
     }
+
+    /// Specifies redirect processing settings.
+    public enum RedirectPolicy {
+        /// Redirects are not followed.
+        case disallow
+        /// Redirects are followed with a specified limit. Cycle detection reqiures that all visited URL's are kept in memory.
```
weissi:
```suggestion
        /// Redirects are followed with a specified limit.
        ///
        /// - parameters:
        ///     - max: The maximum number of allowed redirects.
        ///     - allowCycles: Whether cycles are allowed.
        ///
        /// - warning: Cycle detection will keep all visited URLs in memory which means a malicious server could use this as a denial-of-service vector.
```

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -423,6 +428,14 @@ extension HTTPClient.Configuration {
             self.read = read
         }
     }
+
+    /// Specifies redirect processing settings.
+    public enum RedirectPolicy {
```
weissi: I don't think we should have this as a `public enum` because that means we can never add policies. I would recommend making this a private enum and wrapping it into a `public struct RedirectPolicy` with all the enum cases as

```
public struct RedirectPolicy {
    private enum Policy {
        case disallow
        ...
    }
    private var policy: Policy

    public static let disallow = RedirectPolicy(policy: .disallow)

    public static func allow(max: Int, allowCycles: Bool) -> RedirectPolicy {
        return RedirectPolicy(policy: .allow(max, allowCycles)
    }
}
```

Sources/AsyncHTTPClient/HTTPClient.swift
ianpartridge: reqiures -> requires
URL's -> URLs

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -317,33 +322,33 @@ public class HTTPClient {
...
         /// Default client timeout, defaults to no timeouts.
         public var timeout: Timeout
         /// Upstream proxy, defaults to no proxy.
         public var proxy: Proxy?
         /// Ignore TLS unclean shutdown error, defaults to `false`.
         public var ignoreUncleanSSLShutdown: Bool
 
-        public init(tlsConfiguration: TLSConfiguration? = nil, followRedirects: Bool = false, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
-            self.init(tlsConfiguration: tlsConfiguration, followRedirects: followRedirects, timeout: timeout, proxy: proxy, ignoreUncleanSSLShutdown: false)
+        public init(tlsConfiguration: TLSConfiguration? = nil, redirects: RedirectPolicy = .allow(max: 5, allowCycles: false), timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
```
ianpartridge: There's quite a lot of code duplication here, with each initializer having the same hard-coded `.allow(max: 5, allowCycles: false)` default value. We should pull that out into a separate

```swift
static let defaultRedirectPolicy: RedirectPolicy = .allow(max: 5, allowCycles: false)
```

then this can be:

```swift
public init(tlsConfiguration: TLSConfiguration? = nil, redirects: RedirectPolicy = HTTPClient.defaultRedirectPolicy, timeout: Timeout = Timeout(), proxy: Proxy? = nil) {
  ...
}
```

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -423,6 +427,19 @@ extension HTTPClient.Configuration {
...
     }
+
+    /// Specifies redirect processing settings.
+    public enum RedirectPolicy {
+        /// Redirects are not followed.
+        case disallow
+        /// Redirects are followed with a specified limit. Cycle detection reqiures that all visited URL's are kept in memory.
+        case allow(max: Int, allowCycles: Bool)
+
+        struct State {
```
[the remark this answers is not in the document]
artemredkin: good idea, done!

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -563,4 +563,32 @@ class HTTPClientTests: XCTestCase {
...
+                                    configuration: HTTPClient.Configuration(certificateVerification: .none, redirects: .allow(max: 5, allowCycles: true)))
+        defer {
+            XCTAssertNoThrow(try httpClient.syncShutdown())
+            XCTAssertNoThrow(try httpBin.shutdown())
+        }
+
+        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/redirect/infinite1").wait(), "Should fail with redirect limit") { error in
+            XCTAssertEqual(error as! HTTPClientError, HTTPClientError.redirectLimitReached)
+        }
+    }
```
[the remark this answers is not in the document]
artemredkin: do you mean when schema and path are capitalised differently?

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: Agreed :D, fixed, thanks!

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: then `defaultRedirectPolicy` would also have to be public, what if it will be optional instead?

Sources/AsyncHTTPClient/HTTPClient.swift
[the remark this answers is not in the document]
artemredkin: Done!

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: How about `RedirectConfiguration.disallow` and `RedirectConfiguration.follow`?

ianpartridge: I like `.follow` 🙂

artemredkin:
@weissi I rewrote the backpressure, could you please take a look at a comment to see if it makes sense?

weissi:
@artemredkin cool, looks good! On top of the test with actual channels, did the `EmbeddedChannel` not work?

weissi:
ping @artemredkin on the above question

artemredkin:
@weissi not yet, do you think its critical? we have "integration-style" test for backpressure for now

weissi:
@artemredkin I think it would be good to have both especially given that the integration test might turn out to not be a 100% reliable (I think it is but hard to tell for sure). I also think the integration test will be much easier to follow for people who aren't you & me because we worked on that integration test together.

artemredkin:
@weissi I filed https://github.com/swift-server/async-http-client/issues/117 to implement that test for `1.1.0`. Test is passing now and test itself is not part of this PR (redirects). wdyt, should we continue as is in order to release `1.0.0`?

weissi:
@artemredkin perfect, no, let’s not block 1.0.0

artemredkin:
@weissi @Lukasa @tomerd @ianpartridge any additional comments?

Lukasa:
None from me.

ianpartridge:
LGTM.

PR #122, opened 2019-10-30
test with thread sanitizer and check for api breakage in ci

tomerd:
motivation: better ci testing

changes:
* add api breakage script to test api breakage in ci
* add "shell" utility docker-compose task to run api breakage script
* change test task to treat warning as errors
* change test task to use thread sanitizer in 5.1

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,7 +13,7 @@ ENV LANGUAGE en_US.UTF-8
 
 # dependencies
 RUN apt-get update && apt-get install -y wget
-RUN apt-get update && apt-get install -y lsof dnsutils netcat-openbsd net-tools libz-dev # used by integration tests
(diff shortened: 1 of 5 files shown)


tomerd:
@weissi unfortunately thread sanitizer fails:

```
[563/563] Linking async-http-clientPackageTests.xctest
ThreadSanitizer:DEADLYSIGNAL
==1517==ERROR: ThreadSanitizer: SEGV on unknown address 0xffffffffff700468 (pc 0x55fad9da80a8 bp 0x000000000018 sp 0x7ffdd5b1ead0 T1517)
==1517==The signal is caused by a READ memory access.
ThreadSanitizer:DEADLYSIGNAL
ThreadSanitizer: nested bug in the same thread, aborting.
```

any ideas why? works with docker on macOS

Issue #127, opened 2019-11-06
Investigate support for nio-transport-services


artemredkin:

weissi:
cross posting from #135 

how this works is: You only need one `#if ...` and then you can

```swift
#if canImport(Network)
// Creating the "universal bootstrap" with the `NIOSSLClientTLSProvider`.
let tlsProvider = NIOTSClientTLSProvider(tlsOptions: .init())
let bootstrap = NIOTSClientTLSProvider(NIOTSEventLoopGroup(group: group), tls: tlsProvider)
#else
// TLS setup.
let configuration = TLSConfiguration.forClient()
let sslContext = try NIOSSLContext(configuration: configuration)

// Creating the "universal bootstrap" with the `NIOSSLClientTLSProvider`.
let tlsProvider = NIOSSLClientTLSProvider<ClientBootstrap>(context: sslContext, serverHostname: "example.com")
let bootstrap = NIOClientTCPBootstrap(ClientBootstrap(group: group), tls: tlsProvider)
#endif

// Bootstrapping a connection using the "universal bootstrapping mechanism"
let connection = bootstrap.enableTLS()
                     .connect(host: "example.com", port: 12345)
                     .wait()
```

For AsyncHTTPClient, you'll want to cache two bootstraps:

```
let tlsBootstrap: NIOClientTCPBootstrap = makeBootstrapForPlatform().enableTLS()
let plainTextBootstrap: NIOClientTCPBootstrap = makeBootstrapForPlatform()
```

and then you can just use the right one to bootstrap connections from depending on if that's TLS or not.

artemredkin:
This was implemented in #184

PR #135, opened 2019-11-19
Support NIO Transport Services

Yasumoto:
This is the investigation for https://github.com/swift-server/async-http-client/issues/127

This builds on https://github.com/apple/swift-nio/pull/1253 and https://github.com/apple/swift-nio-transport-services/pull/65, mostly motivated by https://github.com/swift-aws/aws-sdk-swift-core/pull/141.

Caveats:

1. I wasn't able to see how to support `ChannelOptions.maxMessagesPerRead`; we'll need this to make sure we don't block in `testUploadStreamingBackpressure`.
2. As mentioned in https://github.com/apple/swift-nio/issues/674, we're really pushing back on our users to _always_ use `NIOTSEventLoopGroup` if it's available. We should likely also add a check if we're passed in a `.shared()` EventLoop that we use the corresponding bootstrap, but figured I'd post this up here for an initial review first.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -18,6 +18,10 @@ import NIOConcurrencyHelpers
 import NIOHTTP1
 import NIOHTTPCompression
 import NIOSSL
+#if canImport(Network)
+import Network
+import NIOTransportServices
+#endif
 
 /// HTTPClient class provides API for request execution.
 ///
@@ -61,7 +65,15 @@ public class HTTPClient {
         case .shared(let group):
             self.eventLoopGroup = group
         case .createNew:
-            self.eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+            #if canImport(Network)
+                if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *) {
+                    self.eventLoopGroup = NIOTSEventLoopGroup()
+                } else {
+                    self.eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
+                }
+            #else
+                self.eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
+            #endif
         }
         self.configuration = configuration
     }
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -18,6 +18,7 @@ import NIOConcurrencyHelpers
 import NIOHTTP1
 import NIOHTTPCompression
 import NIOSSL
+
 #if canImport(Network)
 import Network
 import NIOTransportServices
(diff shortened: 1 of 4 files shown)


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -61,7 +66,15 @@ public class HTTPClient {
         case .shared(let group):
             self.eventLoopGroup = group
         case .createNew:
-            self.eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+            #if canImport(Network)
+                if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *) {
+                    self.eventLoopGroup = NIOTSEventLoopGroup()
```
artemredkin: @weissi do you think it's a good idea to default to TS on darwin?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -246,8 +259,20 @@ public class HTTPClient {
...
         let task = Task<Delegate.Response>(eventLoop: delegateEL)
 
-        var bootstrap = ClientBootstrap(group: channelEL ?? delegateEL)
-            .channelOption(ChannelOptions.socket(SocketOptionLevel(IPPROTO_TCP), TCP_NODELAY), value: 1)
+
+        #if canImport(Network)
+        var bootstrap: ClientTransportBootstrap
+
+        if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *) {
+            bootstrap = NIOTSConnectionBootstrap(group: self.eventLoopGroup)
```
artemredkin: why not `channelEL ?? delegateEL`?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -79,8 +84,22 @@ class HTTPClientTests: XCTestCase {
 
     func testGetWithDifferentEventLoopBackpressure() throws {
         let httpBin = HTTPBin()
-        let loopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
-        let external = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        let loopGroup: EventLoopGroup
+        let external: EventLoopGroup
+
+        #if canImport(Network)
```
artemredkin: maybe we should extract this to a test helper function (that returns one ELG)?

weissi:
ping @Yasumoto this is now ready :)

weissi:
how this works is: You only need one `#if ...` and then you can

```swift
#if canImport(Network)
// Creating the "universal bootstrap" with the `NIOSSLClientTLSProvider`.
let tlsProvider = NIOTSClientTLSProvider(tlsOptions: .init())
let bootstrap = NIOTSClientTLSProvider(NIOTSEventLoopGroup(group: group), tls: tlsProvider)
#else
// TLS setup.
let configuration = TLSConfiguration.forClient()
let sslContext = try NIOSSLContext(configuration: configuration)

// Creating the "universal bootstrap" with the `NIOSSLClientTLSProvider`.
let tlsProvider = NIOSSLClientTLSProvider<ClientBootstrap>(context: sslContext, serverHostname: "example.com")
let bootstrap = NIOClientTCPBootstrap(ClientBootstrap(group: group), tls: tlsProvider)
#endif

// Bootstrapping a connection using the "universal bootstrapping mechanism"
let connection = bootstrap.enableTLS()
                     .connect(host: "example.com", port: 12345)
                     .wait()
```

Yasumoto:
Oh heck yah 💪

Sources/AsyncHTTPClient/HTTPClient.swift
weissi: Yes

adam-fowler:
Hey guys,
I have a version of this working on another branch. There are still a few issues: a couple of tests fail and TLS testing with NIOTS fails but that looks like it is a problem with the TLS testing, not the code. 

I have tested it with [aws-sdk-swift](https://github.com/swift-aws/aws-sdk-swift) and it works on both macOS and iOS. Should I open another PR?

fabianfett:
> I have tested it with [aws-sdk-swift](https://github.com/swift-aws/aws-sdk-swift) and it works on both macOS and iOS. Should I open another PR?

+1 the sooner we get the special Darwin treatment out of `aws-sdk-swift` the better. I can only imagine @Yasumoto would be supporting this.

Yasumoto:
@adam-fowler amazing! If you’ve already got a branch ready for review let’s close this out and transition to a new PR 🙇

PR #136, opened 2019-11-19
Adds a performance benchmark.

kylebrowning:
This adds performance benchmarks so we can see connection pooling and other various PRs and how they affect network efficiency.

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientPerformanceBenchmarkTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientPerformanceBenchmarkTests.swift
@@ -17,7 +17,7 @@ import AsyncHTTPClient
 import Foundation
 import XCTest
 
-private class NIOTestOperation: AsynchronousOperation {
+private class AsyncHTTPClientTestOperation: AsynchronousOperation {
     private let httpClient: HTTPClient
     
     init(httpClient: HTTPClient) {
@@ -41,19 +41,11 @@ private class URLSessionTestOperation: AsynchronousOperation {
     }
 
     override func execute() {
-
         let url = URL(string: "https://swift.org")!
-
         var request = URLRequest(url: url)
         request.httpMethod = "GET"
-
-        let task = session.dataTask(with: request) { [weak self] (data, response, error) in
-
-            guard data != nil else {
-                return
-            }
-
-            self?.finish()
+        let task = session.dataTask(with: request) { (_, _, _) in
+            self.finish()
         }
         task.resume()
     }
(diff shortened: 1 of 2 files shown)


Tests/AsyncHTTPClientTests/HTTPClientPerformanceBenchmarkTests.swift
```diff
@@ -0,0 +1,117 @@
...
+private class AsyncHTTPClientTestOperation: AsynchronousOperation {
+    private let httpClient: HTTPClient
+    
+    init(httpClient: HTTPClient) {
+        self.httpClient = httpClient
+        super.init()
+    }
+    
+    override func execute() {
+        httpClient.get(url: "https://swift.org").whenComplete { _ in
```
weissi: we shouldn't reach out to the internet during benchmarks, you'd be benchmarking mostly the network and not the library then.

Tests/AsyncHTTPClientTests/HTTPClientPerformanceBenchmarkTests.swift
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+final class PerformanceTests: XCTestCase {
+
+    var httpClient: HTTPClient!
+    var session: URLSession!
+
+    private let operationQueue = OperationQueue()
+    private let processingQueue = DispatchQueue(label: "TestOperationQueue", qos: .utility, attributes: .concurrent)
```
weissi: let's not use `.concurrent` dispatch queues. They're really slow on Linux and also usually not the right choice on Darwin.

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -490,6 +490,95 @@ extension ByteBuffer {
...
+open class AsynchronousOperation: Operation {
+
+    // MARK: - Properties
+    private let stateQueue = DispatchQueue(
+        label: "com.asynchttpclient.operation.state",
+        attributes: .concurrent)
+
+    private var rawState = OperationState.ready
+
+    @objc private dynamic var state: OperationState {
```
weissi: can we remove all the `@objc` stuff? That won't work on Linux I don't think. Also is the KVO/KVC? That won't work either I don't think.

Tests/AsyncHTTPClientTests/HTTPClientPerformanceBenchmarkTests.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    private struct K {
+        static let numberOfOperations = 2
+    }
+
+    override func setUp() {
+        operationQueue.maxConcurrentOperationCount = 1
```
weissi: if we're doing 1 concurrent operation, why do we need the `OperationQueue` thing? Can't we just do

```
let httpClient = ...
for _ in 0..<numberOfRequests {
    httpClient.get(...).wait()
}
```

and let that be the whole benchmark?

PR #137, opened 2019-11-23
Codable Support

kyleishie:
This adds support for Codable types both when making requests and when handling responses.

Created protocols Encoder and Decoder to allow encoders and decoder to be swapped out with third party modules, e.g., XML, YAML, etc.  The existing Foundation coders conform to these protocols.

Added a static method for creating a HTTPClient.Body from an Encodable type.

Added HTTPClient.Response computed property for retrieving body as Data instead of ByteBuffer.

Extended EventLoopFuture<HTTPClient.Response> with methods for easy inline body decoding.  I implemented two versions of this method; one for use cases in which a response with no body throws an error and the other for use cases that allow optional bodies.

### DIFF
--- a/Sources/AsyncHTTPClient/Body+Encodable.swift
+++ b/Sources/AsyncHTTPClient/Body+Encodable.swift
@@ -1,0 +1,31 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Swift Server Working Group and the AsyncHTTPClient project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+
+import Foundation
+
+extension HTTPClient.Body {
+    
+    public static func encodable<T : Encodable>(_ encodable: T, encoder: Encoder = JSONEncoder()) throws -> HTTPClient.Body {
+        return HTTPClient.Body.data(try encoder.encode(encodable))
+    }
+    
+}
+
+public protocol Encoder {
+    func encode<T>(_ value: T) throws -> Data where T : Encodable
+}
+
+extension JSONEncoder : Encoder {}
+extension PropertyListEncoder : Encoder {}
--- a/Sources/AsyncHTTPClient/Response+Decodable.swift
+++ b/Sources/AsyncHTTPClient/Response+Decodable.swift
@@ -1,0 +1,80 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Swift Server Working Group and the AsyncHTTPClient project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import NIO
+
+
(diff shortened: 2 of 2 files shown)


Sources/AsyncHTTPClient/Response+Decodable.swift
```diff
@@ -0,0 +1,80 @@
...
+    }
+    
+    /// Decode the response body as T using the given decoder.
+    ///
+    /// - parameters:
+    ///     - type: The type to decode.  Must conform to Decoable.
+    ///     - decoder: The decoder used to decode the reponse body.  Defaults to JSONDecoder.
+    /// - returns: A future decoded type.
+    /// - throws: BodyError.noBodyData when no body is found in reponse.
+    public func decode<T : Decodable>(as type: T.Type, decoder: Decoder = JSONDecoder()) -> EventLoopFuture<T> {
```
Lukasa: Is there any reason these functions need to be extensions on `EventLoopFuture` instead of on `HTTPClient.Response`?

Sources/AsyncHTTPClient/Body+Encodable.swift
```diff
@@ -0,0 +1,31 @@
...
+extension HTTPClient.Body {
+    
+    public static func encodable<T : Encodable>(_ encodable: T, encoder: Encoder = JSONEncoder()) throws -> HTTPClient.Body {
+        return HTTPClient.Body.data(try encoder.encode(encodable))
+    }
+    
+}
+
+public protocol Encoder {
+    func encode<T>(_ value: T) throws -> Data where T : Encodable
```
ianpartridge: In Foundation `JSONEncoder` deliberately doesn't implement this method via a protocol, it just has a method for it with no protocol involved: https://github.com/apple/swift-corelibs-foundation/blob/2195121bf063e2db7c71080fcae7e76e8a43950e/Foundation/JSONEncoder.swift#L234

AFAIK this was a deliberate design choice in Foundation.

It seems to me that if this protocol is going to exist it really belongs in Foundation.

kyleishie: I agree that this belongs in Foundation.  I simply added it here to support the interchange of encoders and decoders.  However, I also agree with @Lukasa that this functionality is probably not needed.  If we want I can remove these protocol and the encoder and decoder interchange api.

Sources/AsyncHTTPClient/Response+Decodable.swift
kyleishie: This is simply convenience to allow the user to do something like this. 
`httpClient.get(url: url).decode(as: Something.self)`

Sources/AsyncHTTPClient/Body+Encodable.swift
kyleishie: With that being said I implemented this functionality with two things in mind.
1.  The ability to inject coders for other protocols, e.g., XML, YAML.
2.  The ability to inject a customized json coder, e.g., a JSONDecoder where keyDecodingStrategy = .convertFromSnakeCase.

Again this may be overkill for the API but I felt it was a nice optional parameter to have.

Sources/AsyncHTTPClient/Response+Decodable.swift
Lukasa: Yeah, I think I'd rather have the visual noise of requiring that this appear in a `map`.

kyleishie: Ok.  Whats the thinking with that decision?  The only reason I even thought to create this extension was because I found myself ~~writing~~ copy/pasting the same boilerplate map function to all of my requests.  I can see cases where one would want a one-off, custom decode handler but that seems like a poor reason to require it to be written out ever time.

Sources/AsyncHTTPClient/Body+Encodable.swift
kyleishie: @ianpartridge what do you think about my comments above?  Should I remove this functionality?

ianpartridge: I think so, personally.

ktoso:

PR #139, opened 2019-11-29
Bugfix HTTPS SNI and IP Address

StrangeDays:
Motivation:

Solving the SNI Bug

Modifications:

Added an internal extension on String for checking if the hostname is an IP Address -- see the private extension on SNI. Additionally using the IPv4Address and IPv6Address Function from Network above 10.14 as protecting with #availabe.
Adding the test for HTTPS and IP in as hostname

Result:

We get results with an IP as Hostname

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -275,7 +275,7 @@ public class HTTPClient {
                     let taskHandler = TaskHandler(task: task, delegate: delegate, redirectHandler: redirectHandler, ignoreUncleanSSLShutdown: self.configuration.ignoreUncleanSSLShutdown)
                     return channel.pipeline.addHandler(taskHandler)
                 }
-        }
+            }
 
         if let timeout = self.resolve(timeout: self.configuration.timeout.connect, deadline: deadline) {
             bootstrap = bootstrap.connectTimeout(timeout)
@@ -285,12 +285,11 @@ public class HTTPClient {
         bootstrap.connect(host: address.host, port: address.port)
             .map { channel in
                 task.setChannel(channel)
-        }
-        .flatMap { channel in
-            channel.writeAndFlush(request)
-        }
-        .cascadeFailure(to: task.promise)
-
+            }
+            .flatMap { channel in
+                channel.writeAndFlush(request)
+            }
+            .cascadeFailure(to: task.promise)
         return task
     }
 
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -275,7 +275,7 @@ public class HTTPClient {
                     let taskHandler = TaskHandler(task: task, delegate: delegate, redirectHandler: redirectHandler, ignoreUncleanSSLShutdown: self.configuration.ignoreUncleanSSLShutdown)
                     return channel.pipeline.addHandler(taskHandler)
                 }
-            }
+        }
 
         if let timeout = self.resolve(timeout: self.configuration.timeout.connect, deadline: deadline) {
             bootstrap = bootstrap.connectTimeout(timeout)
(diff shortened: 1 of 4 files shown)


Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -32,3 +36,36 @@ public final class HTTPClientCopyingDelegate: HTTPClientResponseDelegate {
...
+            }
+        } else {
+            var ipv4Addr = in_addr()
+            var ipv6Addr = in6_addr()
+
+            return self.withCString { ptr in
+                return inet_pton(AF_INET, ptr, &ipv4Addr) == 1 ||
+                       inet_pton(AF_INET6, ptr, &ipv6Addr) == 1
+            }
+        }
```
Lukasa: The fallback block is common across both operating systems, so it may be better to factor that out into a helper function that we can call directly to save repetition.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -14,6 +14,10 @@
 
 import NIO
 import NIOHTTP1
+#if os(Linux)
+#else
+import Network
+#endif
```
Lukasa: It might be better to guard this behind `#if canImport(Network)`.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -285,11 +285,11 @@ public class HTTPClient {
...
-            }
-            .flatMap { channel in
-                channel.writeAndFlush(request)
-            }
-            .cascadeFailure(to: task.promise)
+        }
+        .flatMap { channel in
+            channel.writeAndFlush(request)
+        }
+        .cascadeFailure(to: task.promise)
```
Lukasa: Can we remove these whitespace changes?

StrangeDays: I ran now with swiftformat ... i think they back to normal now ...

StrangeDays:
anyone else want to review ? to fix that bug?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -831,7 +844,7 @@ class HTTPClientTests: XCTestCase {
             XCTAssertNoThrow(try web.writeOutbound(.head(.init(version: .init(major: 1, minor: 0),
                                                                status: .ok,
                                                                headers: HTTPHeaders([("connection",
-                                                                                      i % 2 == 0 ? "close" : "keep-alive")])))))
+                                                                                     i % 2 == 0 ? "close" : "keep-alive")])))))
```
Lukasa: Mind removing this extraneous change?

StrangeDays: no prob -- I'll check asap (basically I'm using swiftlint)

StrangeDays: i checked & I've use the` swiftformat --swiftversion 5 .` before push

Lukasa:
@artemredkin or @weissi want to take a look here as well?

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -15,6 +15,32 @@
...
+            if IPv4Address(self) != nil || IPv6Address(self) != nil {
+                return true
+            }
+            return false
+        }
+    }
+
+#else
+    internal extension String {
+        var isIPAddress: Bool {
```
weissi: @Lukasa should we add smth like this to NIO? This is unrelated to this PR but we could make a 'good-first-issue'

Lukasa: Yeah, I wouldn't object, though we'd be extending a type we don't own so we'd have to prefix the method name, which is kinda gross.

Lukasa: And actually don't we have this already in the form of `SocketAddress(ipAddress:port:)`?

weissi: @Lukasa ish, need to give a port too. But maybe that's not too bad?

```
extension String {
    var isIPAddress: Bool {
        return SocketAddress(ipAddress: self, port: 0) != nil
   }
}
```

?

Lukasa: I suspect it's fine, and the only reason not to use it here is that you do have to heap-allocate for that code.

weissi:

weissi:

weissi:

PR #145, opened 2019-12-17
use Swift 5.1.3 in CI

weissi:


docker/docker-compose.1804.51.yaml
```diff
@@ -7,7 +7,7 @@ services:
     build:
       args:
         ubuntu_version: "bionic"
-        swift_version: "5.1"
+        swift_version: "5.1.3"
```
tomerd: this means we would need to keep this up to date as minor revisions are released

weissi: @tomerd Yes. Is there another way to force the CI to pull the latest 5.1?

tomerd: the issue here is that docker will catch the 5.1 images, so the only way would be to forcibly invalidate the cache which is not ideal

weissi: Right. I think manually bumping is fine for now. Or we implement something where it’d kill the cache once a week or so?

tomerd: yes we can do that. we can also create a CI job that will nuke the relevant cache for a specific version. this may be part of the checklist when we release new versions?

PR #151, opened 2020-01-19
Support UNIX Domain Sockets

krzyzanowskim:
Adds support for UNIX Domain Socket requests.

Usage:
```swift
let socketURL = URL(string: "unix:///var/run/docker.sock")!
let req = try HTTPClient.Request(url: URL(string: "/users/list", relativeTo: socketURL)!, method: .GET)
let response = try httpClient.execute(request: req).wait()



```


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -89,24 +89,47 @@ extension HTTPClient {
...
+            }
+        }
 
         struct RedirectState {
             var count: Int
             var visited: Set<URL>?
         }
 
         var redirectState: RedirectState?
+        private var value: HTTPRequest
```
Lukasa: I don't love using a protocol existential here. I think the splitting of basic request types is a good idea, but I'd rather use an enumeration that we can switch over than the protocol existential.

artemredkin: It seems the only difference between implementations is how they handle scheme? Why not just add a `unix://` to a list of supported schemes?

artemredkin:
Two questions:
1. What is the use-case you have in mind? Never personally encountered it, just curious
2. Cory, could ssl negotiation be run on top of the unix socket? Is there anything tcp/ip specific? Is it something potentially useful?

krzyzanowskim:
> What is the use-case you have in mind? 

My use case is local Docker server over unix socket

artemredkin:
Point 2 would probably not be useful since we running it on the same machine...

Sources/AsyncHTTPClient/HTTPHandler.swift
krzyzanowskim: > Why not just add a unix:// to a list of supported schemes?

slightly different asserts in Request constructor

krzyzanowskim:
> could ssl negotiation be run on top of the unix socket?

in general yes. I don't know if nio can handle it though. Looks like eg. postgres can utilise that

Lukasa:
> Cory, could ssl negotiation be run on top of the unix socket? Is there anything tcp/ip specific? Is it something potentially useful?

You can run TLS over the unix socket, it's mostly not useful though it can be if the process on the other side of the unix socket is a TCP proxy. It's worth keeping it as a thing we could potentially support in future. 

> in general yes. I don't know if nio can handle it though.

NIO's TLS implementation is 100% composable, so it can be run over literally any stream transport. :wink:

Sources/AsyncHTTPClient/HTTPHandler.swift
krzyzanowskim: >  I'd rather use an enumeration that we can switch over than the protocol existential.

I'm not a fan. Do you think about if'ing in every `Request` property?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -146,22 +164,26 @@ extension HTTPClient {
...
 
-            guard let host = url.host else {
-                throw HTTPClientError.emptyHost
+                self.host = host
+            } else if Kind.unixSocket.isSchemeSupported(scheme: scheme) {
+                self.kind = .unixSocket
+                self.host = ""
+            } else {
+                throw HTTPClientError.unsupportedScheme(scheme)
             }
```
Lukasa: This block can be rewritten to be substantially cleaner by defining some helpers:

```swift
extension Kind {
    private static var hostSchemes = ["http", "https"]
    private static var unixSchemes = ["unix"]

    init(forScheme scheme: String) throws {
        if Kind.hostSchemes.contains(scheme) {
            self = .host
        } else if Kind.unixSchemes.contains(scheme) {
            self = .unixSocket
        } else {
            throw HTTPClientError.unsupportedScheme
        }
    }

    func hostFromURL(_ url: URL) throws -> String {
        switch self {
        case .host:
            guard let host = url.host else {
            }
            return host
        case .unixSocket:
            return ""
        }
    }
}
```

This code then becomes:

```swift
self.kind = try Kind(scheme: scheme)
self.host = try self.kind.hostFromURL(url)
```

We can then also rewrite `kind.isSchemeSupported` in terms of our statics. This should put most of the complexity into the enum definition, which helps keep the code elsewhere a lot nicer.

krzyzanowskim: thanks for the feedback. Applied as requested.

weissi:

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -88,14 +88,57 @@ extension HTTPClient {
...
+        public var method: HTTPMethod
         /// Remote URL.
-        public let url: URL
+        public var url: URL
         /// Remote HTTP scheme, resolved from `URL`.
-        public let scheme: String
+        public var scheme: String
         /// Remote host, resolved from `URL`.
-        public let host: String
+        public var host: String
```
artemredkin: why do we need those properties to be mutable? If someone changes their value we have to re-check all our invariants (scheme, host, kind, etc)

krzyzanowskim: we don't. I changed it when it was computed value, and didn't change it back when reverted it to the property

weissi:
sorry @krzyzanowskim , this repo is set up with an formating requirements :|

```
=> Checking format... formatting issues!
diff --git a/Sources/AsyncHTTPClient/HTTPClient.swift b/Sources/AsyncHTTPClient/HTTPClient.swift
index bb93908..dd50433 100644
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -290,12 +290,12 @@ public class HTTPClient {
         }
 
         eventLoopChannel.map { channel in
-              task.setChannel(channel)
-            }
-            .flatMap { channel in
-                channel.writeAndFlush(request)
-            }
-            .cascadeFailure(to: task.promise)
+            task.setChannel(channel)
+        }
+        .flatMap { channel in
+            channel.writeAndFlush(request)
+        }
+        .cascadeFailure(to: task.promise)
 
         return task
     }
diff --git a/Sources/AsyncHTTPClient/HTTPHandler.swift b/Sources/AsyncHTTPClient/HTTPHandler.swift
index eb86ee2..fc83eac 100644
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -88,7 +88,6 @@ extension HTTPClient {
 
     /// Represent HTTP request.
     public struct Request {
-
         /// Represent kind of Request
         enum Kind {
             /// Remote host request.
@@ -111,13 +110,13 @@ extension HTTPClient {
 
             func hostFromURL(_ url: URL) throws -> String {
                 switch self {
-                    case .host:
-                        guard let host = url.host else {
-                            throw HTTPClientError.emptyHost
-                        }
-                        return host
-                    case .unixSocket:
-                        return ""
+                case .host:
+                    guard let host = url.host else {
+                        throw HTTPClientError.emptyHost
+                    }
+                    return host
+                case .unixSocket:
+                    return ""
                 }
             }
 
@@ -844,7 +843,7 @@ internal struct RedirectHandler<ResponseType> {
             return nil
         }
 
-        guard request.kind.supports(scheme: self.request.scheme) else {
+        guard self.request.kind.supports(scheme: self.request.scheme) else {
             return nil
```

mind running SwiftFormat on it?

lou-lan:
I am test it, got  some error:
```
Fatal error: Error raised at top level: HTTPClientError.unsupportedScheme("file"): file /BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1100.8.280/swift/stdlib/public/core/ErrorType.swift, line 200
2020-01-27 21:46:57.444788+0800 Demo[30365:399706] Fatal error: Error raised at top level: HTTPClientError.unsupportedScheme("file"): file /BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1100.8.280/swift/stdlib/public/core/ErrorType.swift, line 200
(lldb) 
```

weissi:

Sources/AsyncHTTPClient/HTTPHandler.swift
krzyzanowskim: reverted 535303b

weissi:
@lou-lan try `unix:///path/to/file` instead of `file:///...`

lou-lan:
@weissi I got the same error message：
```
Fatal error: Error raised at top level: HTTPClientError.unsupportedScheme("file"): file /BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1100.8.280/swift/stdlib/public/core/ErrorType.swift, line 200
2020-01-27 22:01:34.073115+0800 Demo[31194:417905] Fatal error: Error raised at top level: HTTPClientError.unsupportedScheme("file"): file /BuildRoot/Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-1100.8.280/swift/stdlib/public/core/ErrorType.swift, line 200
```

```
import Foundation
import AsyncHTTPClient

let socketURL = URL(fileURLWithPath: "unix:///var/run/docker.sock")
let req = try HTTPClient.Request(url: URL(string: "/_version", relativeTo: socketURL)!, method: .GET)
let response = try httpClient.execute(request: req).wait()
print(response)
```

截屏2020-01-2722 05 03

krzyzanowskim:
@lou-lan I updated example in description. Don't use `fileURLWithPath`, it overrides your URL schema

lou-lan:
@krzyzanowskim Success, thank you very much, very cool PR.

PR #170, opened 2020-02-26
Close idle pool connections

### LINKED ISSUE #168: connection pool: idle connection never closed
Currently, the pool keeps idle connections open indefinitely (or until the server closes). That is okay from a correctness point of view but may lead to a very large number of open connections when for example implementing a crawler.

adtrevor:
Motivation: Pooled connections should close at some point (see #168)

Changes:
- Add new `maximumAllowedIdleTimeInConnectionPool ` property to `HTTPClient.Configuration`, its default value is currently `.seconds(60)`, and it can be set to `nil` if one wishes to disable this timeout.
- Add relevant unit test
- Closes #168

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -228,11 +228,9 @@ final class ConnectionPool {
         func removeIdleConnectionHandlersForLease() -> EventLoopFuture<Connection> {
             return self.channel.eventLoop.flatSubmit {
                 self.removeHandler(IdleStateHandler.self).flatMap { () -> EventLoopFuture<Bool> in
-                    self.channel.pipeline.handler(type: IdlePoolConnectionHandler.self).flatMap { idleHandler -> EventLoopFuture<Bool> in
-
-                        self.channel.pipeline.removeHandler(idleHandler).flatMapError { error in
-                            fatalError("\(error)")
-                            return self.channel.eventLoop.makeSucceededFuture(())
+                    self.channel.pipeline.handler(type: IdlePoolConnectionHandler.self).flatMap { idleHandler in
+                        self.channel.pipeline.removeHandler(idleHandler).flatMapError { _ in
+                            self.channel.eventLoop.makeSucceededFuture(())
                         }.map {
                             idleHandler.timeoutClosed.load()
                         }
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -235,6 +235,8 @@ final class ConnectionPool {
                             idleHandler.timeoutClosed.load()
                         }
                     }.flatMapError { error in
+                        // These handlers are only added on connection release, they are not added
+                        // when a connection is made to be instantly leased, so we ignore this error
                         if let channelError = error as? ChannelPipelineError, channelError == .notFound {
                             return self.channel.eventLoop.makeSucceededFuture(!self.channel.isActive)
                         } else {
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -227,10 +227,24 @@ final class ConnectionPool {
 
         func removeIdleConnectionHandlersForLease() -> EventLoopFuture<Connection> {
             return self.channel.eventLoop.flatSubmit {
-                self.removeHandler(IdleStateHandler.self).flatMap {
-                    self.removeHandler(IdlePoolConnectionHandler.self)
-                }.flatMap {
-                    if self.channel.isActive {
+                self.removeHandler(IdleStateHandler.self).flatMap { () -> EventLoopFuture<Bool> in
+                    self.channel.pipeline.handler(type: IdlePoolConnectionHandler.self).flatMap { idleHandler -> EventLoopFuture<Bool> in
+
+                        self.channel.pipeline.removeHandler(idleHandler).flatMapError { error in
+                            fatalError("\(error)")
+                            return self.channel.eventLoop.makeSucceededFuture(())
+                        }.map {
+                            idleHandler.timeoutClosed.load()
+                        }
+                    }.flatMapError { error in
+                        if let channelError = error as? ChannelPipelineError, channelError == .notFound {
+                            return self.channel.eventLoop.makeSucceededFuture(!self.channel.isActive)
+                        } else {
+                            return self.channel.eventLoop.makeFailedFuture(error)
+                        }
+                    }
+                }.flatMap { timeoutClosed in
+                    if timeoutClosed == false {
                         return self.channel.eventLoop.makeSucceededFuture(self)
                     } else {
                         return self.channel.eventLoop.makeFailedFuture(IdleCloseError())
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -232,7 +232,7 @@ final class ConnectionPool {
                         self.channel.pipeline.removeHandler(idleHandler).flatMapError { _ in
                             self.channel.eventLoop.makeSucceededFuture(())
                         }.map {
-                            idleHandler.timeoutClosed.load()
+                            idleHandler.timeoutClosed.load() || !self.channel.isActive
                         }
                     }.flatMapError { error in
                         // These handlers are only added on connection release, they are not added
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -224,6 +224,22 @@ final class ConnectionPool {
         fileprivate var closePromise: EventLoopPromise<Void>
 
         var closeFuture: EventLoopFuture<Void>
+
+        func removeIdleConnectionHandlersForLease() -> EventLoopFuture<Connection> {
+            return self.channel.eventLoop.flatSubmit {
+                self.removeHandler(IdleStateHandler.self).flatMap {
+                    self.removeHandler(IdlePoolConnectionHandler.self)
+                }.flatMap {
+                    if self.channel.isActive {
+                        return self.channel.eventLoop.makeSucceededFuture(self)
+                    } else {
+                        return self.channel.eventLoop.makeFailedFuture(IdleCloseError())
+                    }
+                }
+            }
+        }
+
+        struct IdleCloseError: Error {}
     }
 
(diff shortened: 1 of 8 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -1008,3 +1023,12 @@ internal struct RedirectHandler<ResponseType> {
         }
     }
 }
+
+class IdlePoolConnectionHandler: ChannelInboundHandler, RemovableChannelHandler {
+    typealias InboundIn = NIOAny
+    func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        if let idleEvent = event as? IdleStateHandler.IdleStateEvent, idleEvent == .write {
+            context.close(promise: nil)
+        }
```
weissi: you need to forward all other events here in an `else` branch

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -1631,4 +1631,17 @@ class HTTPClientTests: XCTestCase {
...
+    func testPoolClosesIdleConnections() {
+        let httpBin = HTTPBin()
+        let httpClient = HTTPClient(eventLoopGroupProvider: .createNew, configuration: .init(poolingTimeout: .seconds(1)))
+        defer {
+            XCTAssertNoThrow(try httpBin.shutdown())
+            XCTAssertNoThrow(try httpClient.syncShutdown(requiresCleanClose: true))
+        }
+        XCTAssertNoThrow(try httpClient.get(url: "http://localhost:\(httpBin.port)/get").wait())
+        XCTAssertEqual(httpBin.activeConnections, 1)
+        Thread.sleep(forTimeInterval: 2)
```
weissi: can we reduce this to something significantly less than 2 seconds? Maybe 100 ms. I think you can also remove the `XCTAssertEqual(httpBin.activeConnections, 1)` check because that's technically racy...

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -408,6 +409,8 @@ public class HTTPClient {
         public var redirectConfiguration: RedirectConfiguration
         /// Default client timeout, defaults to no timeouts.
         public var timeout: Timeout
+        /// Timeout of pooled connections
+        public var poolingTimeout: TimeAmount?
```
adtrevor: Is this a good name for this setting?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -497,13 +497,15 @@ extension HTTPClient {
         var cancelled: Bool
         let lock: Lock
         let id = UUID()
+        let poolingTimeout: TimeAmount?
 
-        init(eventLoop: EventLoop) {
+        init(eventLoop: EventLoop, poolingTimeout: TimeAmount? = nil) {
```
adtrevor: Let me know if you think this isn't correct. Since the `Task` has the code that removes / adds back the handlers on release, it needs to know the timeout, so the only other solution I see to avoid passing it to the `Task` would be to refactor the code to move this responsibility outside of `Task`

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -418,30 +421,68 @@ public class HTTPClient {
...
+                    redirectConfiguration: RedirectConfiguration? = nil,
+                    timeout: Timeout = Timeout(),
+                    proxy: Proxy? = nil,
+                    ignoreUncleanSSLShutdown: Bool = false,
+                    decompression: Decompression = .disabled) {
+            self.init(
+                tlsConfiguration: tlsConfiguration,
+                redirectConfiguration: redirectConfiguration,
+                timeout: timeout,
+                poolingTimeout: .seconds(60),
```
adtrevor: What do you think of the default 60 seconds timeout?

Sources/AsyncHTTPClient/HTTPClient.swift
weissi: I'd make it a bit longer and more descriptive. Maybe `maximumAllowedIdleTimeInConnectionPool` or so? @Lukasa ?

Sources/AsyncHTTPClient/HTTPClient.swift
weissi: I'm okay with this, @Lukasa ?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adtrevor: Thanks, done!

Sources/AsyncHTTPClient/HTTPHandler.swift
adtrevor: Thank you

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: I think currently there's a race risk here. What if

T1: decides to pick a certain `Connection` from the pool
T2 (EventLoop associated with the connection's `Channel`): timeout fires, connection closes
T1: tries to send a request on the connection's `Channel`.

However, I _think_ we can fix this. If we do the following logic:

- select a `Connection` to use
- hop over to to the connection's `Channel.eventLoop` and in _there_ check that the channel `isActive` and remove the handlers
- if the channel was active, continue, if not, go back and select another connection

adtrevor: You are right, I will attempt this approach, thanks

adtrevor: @weissi Tell me what you think of https://github.com/swift-server/async-http-client/pull/170/commits/2044699c3b2d0e6713b687eee6ab6f4da71d5a88 🙂

Sources/AsyncHTTPClient/HTTPClient.swift
adtrevor: Ok, I set it to it for now, thanks.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -224,6 +224,22 @@ final class ConnectionPool {
         fileprivate var closePromise: EventLoopPromise<Void>
 
         var closeFuture: EventLoopFuture<Void>
+
+        func removeIdleConnectionHandlersForLease() -> EventLoopFuture<Connection> {
+            return self.channel.eventLoop.flatSubmit {
+                self.removeHandler(IdleStateHandler.self).flatMap {
```
weissi: this is _almost_ correct but there's still a tiny little race left unfortunately (which I didn't think of before): It's possible that the `IdlePoolConnectionHandler` has already issued a `close` but the `close` might not actually have happened yet. Why is that? It's because the `SSLHandler` actually has to actually do a "send a `CloseNotify`/wait for `CloseNotify`" dance to close a connection. So with an `SSLHandler` in the pipeline, a `close` unfortunately won't complete synchronously. So it's possible that a `Channel` is both unusable and yet `channel.isActive == true`.

So I propose two things:
1. let's write a test that highlights this (I'll help with that one in a second)
2. fix the race. The only way I can think of to fix this is to add _one more_ handler at the front which is a `ChannelDuplexHandler`. It holds the state whether the channel is usable. It does this by: listening for `channelInactive` (when received: unusable) and also offering an outbound user event `TearDownChannelEvent` or so. When it receives `TearDownChannelEvent`, this handler will a) mark the channel as unusable b) call `context.channel.close()`.

Then, the only thing that's left is to query the `TearDownChannelEvent` handler's state before deciding whether this channel is usable or not.

Does this make some sense? If not, please ping on Discord or so.

weissi: @adtrevor https://github.com/adtrevor/async-http-client/pull/4

adtrevor: @weissi Thank you!

adtrevor: @weissi Currently the pool waits for a connection to actually close before creating the new connection, so  this means in your test `doActualCloseNowPromise.succeed(())` cannot come after request execution, I modified for now, but then it makes it non-deterministic 😕

I think we can either consider that the code is ok as is and in such case use only non-deterministic tests, or if we consider the current behaviour is wrong re-factor the code to avoid waiting on the connection to close.

adtrevor: The latest commit should fix the race 🙂

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -224,6 +224,34 @@ final class ConnectionPool {
...
         var closeFuture: EventLoopFuture<Void>
+
+        func removeIdleConnectionHandlersForLease() -> EventLoopFuture<Connection> {
+            return self.channel.eventLoop.flatSubmit {
+                self.removeHandler(IdleStateHandler.self).flatMap { () -> EventLoopFuture<Bool> in
+                    self.channel.pipeline.handler(type: IdlePoolConnectionHandler.self).flatMap { idleHandler in
+                        self.channel.pipeline.removeHandler(idleHandler).flatMapError { _ in
+                            self.channel.eventLoop.makeSucceededFuture(())
+                        }.map {
+                            idleHandler.timeoutClosed.load()
```
weissi: you need to _also_ check `self.channel.isActive` here. The close could've come from either us (checked with `idleHandler.timeoutClosed`) or the other side (to be checked with `self.channel.isActive`)

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -224,6 +224,34 @@ final class ConnectionPool {
...
+                        }
+                    }.flatMapError { error in
+                        if let channelError = error as? ChannelPipelineError, channelError == .notFound {
```
weissi: when does this happen? Don't we always add that handler? If there's a legit case where this would happen, please add a comment when that is

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: @adtrevor looks great! One small issue but then we can merge

Sources/AsyncHTTPClient/ConnectionPool.swift
adtrevor: Thanks, fixed it!

Sources/AsyncHTTPClient/ConnectionPool.swift
adtrevor: Fixed

PR #171, opened 2020-02-26
test closing connections whilst syncShutdown

weissi:
Motivation:

There's an edge case where the connection pool won't tolerate closing
connections during syncShutdown.

Modification:

Add test case.

Result:

better coverage


adtrevor:
Just merged this into #169

PR #177, opened 2020-03-05
use 3 EventLoops for tests

weissi:
3 makes it less likely that correlated Channels share an EventLoop so it
should shake out more threading bugs...


ktoso:

PR #180, opened 2020-03-12
Fix leaked TLS handshake promise

### LINKED ISSUE #179: Leaking promise created at ConnectionPool.swift line 373
Hi,

I just checked out the current master commit `9cdd1dc` and wanted to test it with my [lambda-runtime](https://github.com/fabianfett/swift-lambda-runtime).

When building for macOS in DEBUG, I ran immediately into a fatalError:

```
Fatal error: leaking promise created at (file: "/Users/xxx/Library/Developer/Xcode/DerivedData/xxx-brfbdomqotikxpeqafasrajmegus/SourcePackages/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool.swift", line: 373): file /Users/xxx/Library/Developer/Xcode/DerivedData/xxx-brfbdomqotikxpeqafasrajmegus/SourcePackages/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool.swift, line 373
```

This happens on the first invocation of `execute`.

I haven't found a similar issue here, so that's why I created one.

adtrevor:
motivation:
TLS handshake promise was leaked in some cases of failure (fixes #179)

changes:
- Avoid leaking promise
- Clearer completion flow for related futures
- Add testAvoidLeakingTLSHandshakeCompletionPromise test

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -383,32 +383,33 @@ final class ConnectionPool {
             }
 
             return channel.flatMap { channel -> EventLoopFuture<ConnectionPool.Connection> in
-                channel.pipeline.addSSLHandlerIfNeeded(for: self.key, tlsConfiguration: self.configuration.tlsConfiguration, handshakePromise: handshakePromise).flatMap {
+                channel.pipeline.addSSLHandlerIfNeeded(for: self.key, tlsConfiguration: self.configuration.tlsConfiguration, handshakePromise: handshakePromise)
+                return handshakePromise.futureResult.flatMap {
                     channel.pipeline.addHTTPClientHandlers(leftOverBytesStrategy: .forwardBytes)
                 }.map {
                     let connection = Connection(key: self.key, channel: channel, parentPool: self.parentPool)
                     connection.isLeased = true
                     return connection
                 }
-            }.flatMap { connection in
-                handshakePromise.futureResult.map {
-                    self.configureCloseCallback(of: connection)
-                    return connection
-                }.flatMapError { error in
-                    connection.closePromise.succeed(())
-                    let action = self.parentPool.connectionProvidersLock.withLock {
-                        self.stateLock.withLock {
-                            self.state.failedConnectionAction()
-                        }
+            }.map { connection in
+                self.configureCloseCallback(of: connection)
+                return connection
+            }.flatMapError { error in
+                // This promise may not have been completed if we reach this
+                // so we fail it to avoid any leak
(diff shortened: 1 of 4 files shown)


Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -383,32 +383,33 @@ final class ConnectionPool {
             }
 
             return channel.flatMap { channel -> EventLoopFuture<ConnectionPool.Connection> in
-                channel.pipeline.addSSLHandlerIfNeeded(for: self.key, tlsConfiguration: self.configuration.tlsConfiguration, handshakePromise: handshakePromise).flatMap {
+                channel.pipeline.addSSLHandlerIfNeeded(for: self.key, tlsConfiguration: self.configuration.tlsConfiguration, handshakePromise: handshakePromise)
```
adtrevor: The promise *and* returned future were redundant IMO so this only makes use of `handshakePromise` now

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -1656,4 +1656,17 @@ class HTTPClientTests: XCTestCase {
             Thread.sleep(forTimeInterval: 0.01 + .random(in: -0.05...0.05))
         }
     }
+
+    func testAvoidLeakingTLSHandshakeCompletionPromise() {
+        let httpClient = HTTPClient(eventLoopGroupProvider: .createNew)
+        defer {
+            XCTAssertNoThrow(try httpClient.syncShutdown(requiresCleanClose: true))
+        }
+        XCTAssertThrowsError(try httpClient.get(url: "http://localhost/").wait()) { error in
```
weissi: this test isn't great. It relies on nothing being bound to port 80 on `localhost`. Can I suggest a slightly different strategy: Let's bind a random port with a server socket, then immediately close it and then connect that port instead and we have pretty high confidence that this port won't be bound by something else

adtrevor: @weissi Thank you! In https://github.com/swift-server/async-http-client/commit/cfba7e762d7faef23e42f66a438b538436ba973c, the  `refusesConnections: true` option is now set on `HTTPBin`, I think it gives the behaviour we want?

adtrevor: Oops actually I don't think it does, I must still close it because the connection is still accepted it seems

adtrevor: @weissi Latest push should be ok 🙂

PR #181, opened 2020-03-13
Add testProxyStreamingNoDeadlock test

adtrevor:
Because of there is a maximum number of concurrent connections per host
it is possible for the client to deadlock itself when making requests
with a streamed body if the body data is itself requested from the same
host.

If we reach the maximum number of concurrent requests for a certain
host, and if all active tasks use a streamed body that obtains its data
from requests to the same host, then the bodies will never be obtained as
the requests they need to make will be put on a queue whose ability to
dequeue them depends on their very result -> deadlock.


ktoso:

PR #183, opened 2020-03-27
Implement asynchronous shutdown

### LINKED ISSUE #126: no asynchronous shutdown method
async-http-client only has a synchronous shutdown method but it'd be great to have an asynchronous one too.

artemredkin:
Motivation:
Most NIO-based client implementation provide asynchronous shutdown method, we need to provide one as well. Closes #126 

Modifications:
1. Converted all `close` and `prepareClose` methods in pool and connection providers to be asynchronous.
2. Removed checks for not running `close` and `prepareClose` in pool and connection providers on event loop.
3. Converted shutdown to use NIO asynchronous event loop shutdown method.
4. Implemented synchronous shutdown methods on top of asynchronous ones.

Result:
Client of the library can now use asynchronous shutdown when required.


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -93,54 +93,105 @@ public class HTTPClient {
...
+            }
+        }
+    }
 
-        self.pool.prepareForClose()
+    /// Shuts down the client and event loop gracefully. This function is clearly an outlier in that it uses a completion
+    /// callback instead of an EventLoopFuture. The reason for that is that NIO's EventLoopFutures will call back on an event loop.
+    /// The virtue of this function is to shut the event loop down. To work around that we call back on a DispatchQueue
+    /// instead.
+    public func shutdown(_ callback: @escaping (Error?) -> Void) {
```
weissi: Please don’t default to `.global` and take a DispatchQueue instead. A user who wants async shutdown will know what queue they want to be called on and it’s very likely not going to be `.global

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -93,54 +93,105 @@ public class HTTPClient {
...
+        if let eventLoop = MultiThreadedEventLoopGroup.currentEventLoop {
+            preconditionFailure("""
+            BUG DETECTED: syncShutdown() must not be called when on an EventLoop.
+            Calling syncShutdown() on any EventLoop can lead to deadlocks.
+            Current eventLoop: \(eventLoop)
+            """)
+        }
+        let errorStorageLock = Lock()
+        var errorStorage: Error?
+        let continuation = DispatchWorkItem {}
```
weissi: I’d use a fresh, single use serial queue here instead of `.global`

weissi:
CC @adtrevor , mind taking a look at the new async shutdown?

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: sure, one question is though: should `EventLoopGroup` do the same?

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: done, thanks!

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: (done)

artemredkin:

artemredkin:

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -93,20 +93,23 @@ final class ConnectionPool {
...
+    func close(on eventLoop: EventLoop) -> EventLoopFuture<Void> {
+        let connectionProviders = self.connectionProvidersLock.withLock {
+            self.connectionProviders.values
         }
-        self.connectionProvidersLock.withLock {
-            assert(self.connectionProviders.count == 0, "left-overs: \(self.connectionProviders)")
+
+        return EventLoopFuture.andAllComplete(connectionProviders.map { $0.close() }, on: eventLoop).map {
+            self.connectionProvidersLock.withLock {
+                assert(self.connectionProviders.count == 0, "left-overs: \(self.connectionProviders)")
```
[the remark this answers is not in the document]
artemredkin: Ok, I'll look into that when I'll get to those issues, thanks!

Issue #175, opened 2020-03-04
flaky test: HTTPClientTests.testRaceNewRequestsVsShutdown: Invalid number of leased connections (-1)


weissi:
```
Test Case 'HTTPClientTests.testRaceNewRequestsVsShutdown' started at 2020-02-28 02:37:12.372
Assertion failed: Invalid number of leased connections (-1): file /code/Sources/AsyncHTTPClient/ConnectionPool.swift, line 467
Current stack trace:
0    libswiftCore.so                    0x00007f0b72f00830 _swift_stdlib_reportFatalErrorInFile + 115
1    libswiftCore.so                    0x00007f0b72e3b94c <unavailable> + 3463500
2    libswiftCore.so                    0x00007f0b72e3ba3e <unavailable> + 3463742
3    libswiftCore.so                    0x00007f0b72c3729a <unavailable> + 1348250
4    libswiftCore.so                    0x00007f0b72e0fa32 <unavailable> + 3283506
5    libswiftCore.so                    0x00007f0b72c36569 <unavailable> + 1344873
6    async-http-clientPackageTests.xctest 0x000055ff350c3b23 <unavailable> + 588579
7    async-http-clientPackageTests.xctest 0x000055ff350c3b67 <unavailable> + 588647
8    async-http-clientPackageTests.xctest 0x000055ff350bcf8e <unavailable> + 561038
9    async-http-clientPackageTests.xctest 0x000055ff350bc73a <unavailable> + 558906
10   async-http-clientPackageTests.xctest 0x000055ff350c8751 <unavailable> + 608081
11   async-http-clientPackageTests.xctest 0x000055ff350bd13a <unavailable> + 561466
12   async-http-clientPackageTests.xctest 0x000055ff350c879b <unavailable> + 608155
13   async-http-clientPackageTests.xctest 0x000055ff3551da3b <unavailable> + 5151291
14   async-http-clientPackageTests.xctest 0x000055ff350bc642 <unavailable> + 558658
15   async-http-clientPackageTests.xctest 0x000055ff350c2931 <unavailable> + 583985
16   async-http-clientPackageTests.xctest 0x000055ff350bd13a <unavailable> + 561466
17   async-http-clientPackageTests.xctest 0x000055ff350c297b <unavailable> + 584059
18   async-http-clientPackageTests.xctest 0x000055ff3551da3b <unavailable> + 5151291
[...]

PR #184, opened 2020-03-28
Support NIO Transport Services - part 2

adam-fowler:
This is the continuation of the good work of @Yasumoto and @weissi in #135 

The following code adds support for NIO Transport services. When the `ConnectionPool` asks for a connection bootstrap it is returned a `NIOClientTCPBootstrap` which wraps either a `NIOTSConnectionBootstrap` or a `ClientBootstrap` depending on whether the EventLoop we are running on is `NIOTSEventLoop`.

If you initialize an `HTTPClient` with `eventLoopGroupProvider` set to `.createNew` then if you are running on iOS, macOS 10.14 or later it will provide a `NIOTSEventLoopGroup` instead of a `EventLoopGroup`.

Currently a number of tests are failing. 4 of these are related to the NIOSSLUncleanShutdown error the others all seem related to various race conditions which are being dealt with on other PRs. I have tested this code with [aws-sdk-swift](https://github.com/swift-aws/aws-sdk-swift/tree/niots-async-http-client) and it is working on both macOS and iOS.

Things look into:
- The aws-sdk-swift NIOTS HTTP client had issues with on Mojave. We should check if this is the case for async-http-client as well.


adam-fowler:
Here's a list of all the failing tests. Some of them fail both with TS and without.

With Transport Services

- [ ] All the TLS tests. Issue with tests, can't get Network.framework to verify SSL certificates.
- [ ] `testAvoidLeakingTLSHandshakeCompletionPromise` throws connectTimeout instead of NIO.ConnectionError
- [] `testStressGetClose`

Without

- [] `testProxyTLS` crashes inside the server code (this happens with TS as well)
- [] `testUDS...` are crashing with a leaked promise

adam-fowler:
fyi Set Environment variable "ENABLE_TS_TESTS" to "true" to test NIOTS.

Package.swift
```diff
@@ -17,18 +17,20 @@ import PackageDescription
 
 let package = Package(
     name: "async-http-client",
+    platforms: [.iOS(.v12), .tvOS(.v12)],
     products: [
         .library(name: "AsyncHTTPClient", targets: ["AsyncHTTPClient"]),
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.13.1"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.4.1"),
```
artemredkin: we need to update `nio-ssl` version to `2.7.0` to get `NIOSSLClientTLSProvider`

Package.swift
```diff
@@ -17,18 +17,20 @@ import PackageDescription
 
 let package = Package(
     name: "async-http-client",
+    platforms: [.iOS(.v12), .tvOS(.v12)],
```
artemredkin: why do we need this? older platforms should be covered by `#if canImport(Network)` and `if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)`

Package.swift
```diff
@@ -17,18 +17,20 @@ import PackageDescription
...
     products: [
         .library(name: "AsyncHTTPClient", targets: ["AsyncHTTPClient"]),
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.13.1"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.4.1"),
         .package(url: "https://github.com/apple/swift-nio-extras.git", from: "1.3.0"),
+        .package(url: "https://github.com/adam-fowler/swift-nio-transport-services.git", .branch("master")),
```
artemredkin: these changes need to be upstreamed

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -368,11 +370,57 @@ final class ConnectionPool {
...
+            let tlsProvider = try NIOSSLClientTLSProvider<ClientBootstrap>(context: sslContext, serverHostname: (key.scheme == .unix || key.host.isIPAddress) ? nil : key.host)
+            return NIOClientTCPBootstrap(ClientBootstrap(group: eventLoop), tls: tlsProvider)
+        }
+        
+        /// create a TCP Bootstrap based off what type of `EventLoop` has been passed to the function.
+        private func makeBootstrap(on eventLoop: EventLoop) throws -> NIOClientTCPBootstrap {
+            let bootstrap: NIOClientTCPBootstrap
+            #if canImport(Network)
+            if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *), eventLoop is NIOTSEventLoop {
+                let tlsProvider = NIOTSClientTLSProvider(tlsOptions: .init())
```
artemredkin: it seems we'll need some kind of translation from `TLSConfiguration` to `NWProtocolTLS.Options`, @Lukasa what do you think will be the right approach?

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -46,26 +46,6 @@ public final class HTTPClientCopyingDelegate: HTTPClientResponseDelegate {
     }
 }
 
-extension ClientBootstrap {
```
artemredkin: what are the reasons for moving this out of `Utils`?

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -19,6 +19,17 @@ import NIOConcurrencyHelpers
 import NIOHTTP1
 import NIOHTTPCompression
 import NIOSSL
+import NIOTransportServices
+
+func getDefaultEventLoopGroup(numberOfThreads: Int) -> EventLoopGroup {
+    #if canImport(Network)
+    if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *),
+    ProcessInfo.processInfo.environment["ENABLE_TS_TESTS"] == "true" {
```
artemredkin: @weissi do you think we should just default to testing with NIOTS on darwin?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -368,11 +370,57 @@ final class ConnectionPool {
             }
         }
 
+        private func makeNonTSBootstrap(on eventLoop: EventLoop) throws -> NIOClientTCPBootstrap {
```
artemredkin: I would recommend moving it back to `Utils`

Sources/AsyncHTTPClient/ConnectionPool.swift
Lukasa: Right now I think we'll probably need to hand-roll this here. In an ideal world we would define a shim module, but conditional dependency isn't real right now so we can't do that nicely.

Package.swift
adam-fowler: There is no valid code path for iOS 11.0 or less. If we don't have this it'll try to use a non TS path on iOS 11 which isn't valid.

Package.swift
adam-fowler: There is a conversation about his already on https://github.com/apple/swift-nio-transport-services/pull/74. Looks like we'll be changing how this is done

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: The function creates a NIOClientTCPBootstrap now. It isn't something specific to ClientBootstrap. I could move the functions back to Utils, but it would make more sense as an extension of NIOClientTCPBootstrap

artemredkin: Sure, sounds good, I just want to make sure that client pool file is not too big and filled with bunch of stuff

Sources/AsyncHTTPClient/ConnectionPool.swift
adam-fowler: Maybe we should move the whole flow of creating a NIOClientTCPBootstrap to Utils. Doesn't make sense we have one half in ConnectionPool.swift and the other in Utils

Package.swift
Lukasa: However, it's no more broken than it is today, and adding these constraints is a semver major.

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: 👍

Package.swift
adam-fowler: done

Package.swift
adam-fowler: removed

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
```diff
@@ -217,7 +230,9 @@ class HTTPClientInternalTests: XCTestCase {
             func didFinishRequest(task: HTTPClient.Task<Response>) throws {}
         }
 
-        let httpClient = HTTPClient(eventLoopGroupProvider: .createNew)
+        // cannot test with NIOTS as `maxMessagesPerRead` is not supported
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
artemredkin: cc @weissi 
@adam-fowler if we create ELG here we need to make sure to stop it

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -632,7 +637,10 @@ class HTTPClientTests: XCTestCase {
     }
 
     func testResponseFutureIsOnCorrectEL() throws {
-        let group = MultiThreadedEventLoopGroup(numberOfThreads: 4)
+        let group = getDefaultEventLoopGroup(numberOfThreads: 4)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
```
artemredkin: shouldn't this be executed by the tests teardown?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -720,32 +728,32 @@ class HTTPClientTests: XCTestCase {
     }
 
     func testLoopDetectionRedirectLimit() throws {
-        let httpBin = HTTPBin(ssl: true)
```
artemredkin: I think this change needs to be reverted

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -720,32 +728,32 @@ class HTTPClientTests: XCTestCase {
...
 
-        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/redirect/infinite1").wait(), "Should fail with redirect limit") { error in
-            XCTAssertEqual(error as! HTTPClientError, HTTPClientError.redirectCycleDetected)
+        XCTAssertThrowsError(try httpClient.get(url: "http://localhost:\(httpBin.port)/redirect/infinite1").wait(), "Should fail with redirect limit") { error in
+            XCTAssertEqual(error as? HTTPClientError, HTTPClientError.redirectCycleDetected)
         }
     }
 
     func testCountRedirectLimit() throws {
-        let httpBin = HTTPBin(ssl: true)
```
artemredkin: same here

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: In this case the eventLoopGroup has been created inside the test because it required a different number of threads

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: I could create duplicates. One with SSL and one without

Sources/AsyncHTTPClient/ConnectionPool.swift
adam-fowler: done

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: done

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
adam-fowler: done

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -380,9 +384,12 @@ class HTTPClientTests: XCTestCase {
     }
 
     func testProxyTLS() throws {
+        XCTFail("Disabled test as it crashes");
```
artemredkin: Looking into it, for some reason HTTPBin is not setup correctly

artemredkin: ok, so this test actually catches one very specific behaviour that is broken by this PR. TLS proxy works as following: we create a pipeline (without SSL handlers), then send `CONNECT` in plaintext, remove proxy handler and add SSL handlers. In this PR we enable TLS right from the start, meaning that `CONNECT` is sent encrypted. cc @weissi @Lukasa is it possible to bypass TLS or delay its initialization?

weissi: @artemredkin yes, just add the SSL handler at exactly the point from which you want encryption from.

weissi: here's an example for STARTTLS in IMAP: We add the SSL handler after sending the `STARTTLS` command: https://github.com/apple/swift-nio-examples/blob/master/NIOSMTP/NIOSMTP/SendEmailHandler.swift#L132-L147

artemredkin: @weissi this is what we do right now, question is what to do about NIOTS? SSL is handled by TS...

weissi: @artemredkin sadly Network.framework doesn't support this. The only way to get this to work is to tell Network.framework (and therefore NIOTS) to _not_ use TLS and then add the the SSL handler from NIO SSL :). NIO SSL's SSL handler is compatible with all implementations.

If you check the NIOSMTP example, this time at https://github.com/apple/swift-nio-examples/blob/4bdfd14f416df123261714cdddbc8d812fa10acb/NIOSMTP/NIOSMTP/ViewController.swift#L132-L159 you'll also see that's exactly what we do there

weissi: For regular TLS, we bundle `NIOTSConnectionBootstrap` with its regular `NIOTSClientTLSProvider`

```
            bootstrap = NIOClientTCPBootstrap(NIOTSConnectionBootstrap(group: group),
                                              tls: NIOTSClientTLSProvider())
```

but for STARTTLS, we bundle `NIOTSConnectionBootstrap` with the TLS implementation from `swift-nio-ssl` (`NIOSSLClientTLSProvider`):

```
            bootstrap = try NIOClientTCPBootstrap(NIOTSConnectionBootstrap(group: group),
                                                  tls: NIOSSLClientTLSProvider(context: sslContext,
                                                                               serverHostname: hostname))
```

artemredkin: @adam-fowler we need to do three things:
1. modify how we create bootstrap to something like this:
```
fileprivate static func makeBootstrap(on eventLoop: EventLoop, host: String, requiresTLS: Bool, configuration: HTTPClient.Configuration) throws -> NIOClientTCPBootstrap {
        let bootstrap: NIOClientTCPBootstrap
        #if canImport(Network)
        if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *), eventLoop is NIOTSEventLoop {
            if configuration.proxy != nil, requiresTLS {
                let tlsConfiguration = configuration.tlsConfiguration ?? TLSConfiguration.forClient()
                let sslContext = try NIOSSLContext(configuration: tlsConfiguration)
                let hostname = (!requiresTLS || host.isIPAddress) ? nil : host
                bootstrap = try NIOClientTCPBootstrap(NIOTSConnectionBootstrap(group: eventLoop), tls: NIOSSLClientTLSProvider(context: sslContext, serverHostname: hostname))
            } else {
                let tlsProvider = NIOTSClientTLSProvider(tlsOptions: .init())
                bootstrap = NIOClientTCPBootstrap(NIOTSConnectionBootstrap(group: eventLoop), tls: tlsProvider)
            }
        } else {
            bootstrap = try ClientBootstrap.makeBootstrap(on: eventLoop, host: host, requiresTLS: requiresTLS, configuration: configuration)
        }
        #else
        bootstrap = try ClientBootstrap.makeBootstrap(on: eventLoop, host: host, requiresTLS: requiresTLS, configuration: configuration)
        #endif

        if requiresTLS, configuration.proxy == nil {
            return bootstrap.enableTLS()
        }
        return bootstrap
    }
```
2. we need to revert `handshake` promise and add `addSSLHandlerIfNeeded` back
3. in `addSSLHandlerIfNeeded` we need to add `SSLHandler` if `eventLoopGroup` is not TS or if TLS & proxy enabled
If you have questions feel free to ping me on swiftpm/swift-server slack

adam-fowler: Sorry missed that case where the SSL handler should be applied after the proxy handler is processed. You changes all makes sense. I should be able to get them working.

adam-fowler: fixed

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: reverted

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: reverted

adam-fowler:
@weissi @Lukasa @artemredkin I think this is ready for another review. All the tests pass except the ones which have race conditions (and have caused issues on the non NIOTS branch) and the SSL Unclean shutdown ones. 

The race condition errors I have a local change which fixes all but `testStressGetClose()` but @adtrevor is working on a related PR which has a similar fix and @artemredkin has also posted a PR so left the fix out to avoid merge hell. 

I'm not sure what to do with the SSL unclean shutdown errors. This is an error coming from NIOSSL. Is this something that needs caught on NIOTS as well or does it deal with these issues internally.

I added a `TLSConfiguration` to `NWProtocolTLS.Options` conversion. This can set min/max TLS version and disable certificate verification. I'm not sure what other elements of this will be needed, so didn't implement anything else. Like I could implement the cipher suite but its a fair amount of code and I'm not sure it is worth it. I call `preconditionFailure` on anything that isn't supported.  

Error Handling: I created a DNS Error which is an enum with all the DNS errors listed in dnssd. Finding this file took me ages so thought an enum would help here. I had similar issues with TLS but there are so many TLS errors I left it as the `OSStatus` value. If you have a list of common TLS errors we could make an enum of those. The posix error type uses the already available `POSIXErrorCode` enum. Although I just thought we could extend this enum to be the error.

Lukasa:
> Is this something that needs caught on NIOTS as well or does it deal with these issues internally.

Sadly it deals with them internally at the moment.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,127 @@
...
+        
+        // key log callback
+        if let _ = self.keyLogCallback {
+            preconditionFailure("TLSConfiguration.keyLogCallback is not supported")
+        }
+        
+        // private key
+        if let _ = self.privateKey {
+            preconditionFailure("TLSConfiguration.privateKey is not supported")
+        }
```
Lukasa, line 95: Sadly, supporting this and `certificateChain` in a cross-platform way is really hard, I think. NWProtocolTLS wants a `sec_identity_t`. I know of no way to create a `sec_identity_t` from data in memory: it requires adding the private key to the keychain. We could try to do that along with a label that allowed us to try to find the same private key again to avoid duplicating it, but it's a lot of work. Leaving it for another day seems reasonable.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,127 @@
...
+        
+        // renegotiation support key is unsupported
+        
+        // trust roots
+        if let trustRoots = self.trustRoots {
+            guard case .default = trustRoots else {
+                preconditionFailure("TLSConfiguration.trustRoots != .default is not supported")
+            }
```
Lukasa, line 103: This one is also reasonably do-able, if we configure a custom trust challenge. For now, though, let's leave it as-is.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -47,21 +53,85 @@ public final class HTTPClientCopyingDelegate: HTTPClientResponseDelegate {
...
+        requiresTLS: Bool,
+        configuration: HTTPClient.Configuration
+    ) throws -> NIOClientTCPBootstrap {
+        let bootstrap: NIOClientTCPBootstrap
+        #if canImport(Network)
+        // if eventLoop is compatible with NIOTransportServices create a NIOTSConnectionBootstrap
+        if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *),
+            var tsBootstrap = NIOTSConnectionBootstrap(validatingGroup: eventLoop) {
+            
+            tsBootstrap = tsBootstrap.channelOption(NIOTSChannelOptions.waitForActivity, value: false)
```
Lukasa, line 84: Why are we setting this to `false` unconditionally?

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -47,21 +53,85 @@ public final class HTTPClientCopyingDelegate: HTTPClientResponseDelegate {
...
+            let tlsConfiguration = configuration.tlsConfiguration ?? TLSConfiguration.forClient()
+            // if we have a proxy and require TLS then use NIOSSL tls support
+            if configuration.proxy != nil, requiresTLS {
+                let sslContext = try NIOSSLContext(configuration: tlsConfiguration)
+                let hostname = (!requiresTLS || host.isIPAddress) ? nil : host
+                bootstrap = try NIOClientTCPBootstrap(tsBootstrap, tls: NIOSSLClientTLSProvider(context: sslContext, serverHostname: hostname))
```
Lukasa, line 90: Why are we adding a `TLSProvider` here if we're not enabling it? Proxy support doesn't use the `TLSProvider` interface so far as I can see.

adam-fowler:
> > Is this something that needs caught on NIOTS as well or does it deal with these issues internally.
> 
> Sadly it deals with them internally at the moment.

So I should ignore these tests? In general I am getting other errors back, apart from `testNoContentLengthForSSLUncleanShutdown` which doesn't throw any error.

Lukasa:
I suspect the tests should simply be skipped: as there's no unclean shutdown reporting, these tests are kind of moot.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,127 @@
...
+        if self.applicationProtocols.count > 0 {
+            preconditionFailure("TLSConfiguration.applicationProtocols is not supported")
+        }
+        /*for applicationProtocol in self.applicationProtocols {
+            applicationProtocol.utf8.withContiguousStorageIfAvailable { buffer in
+                guard let opaquePointer = OpaquePointer(buffer.baseAddress) else { return }
+                let int8Pointer = UnsafePointer<Int8>(opaquePointer)
+                sec_protocol_options_add_tls_application_protocol(options.securityProtocolOptions, int8Pointer)
+            }
+        }*/
```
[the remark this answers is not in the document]
adam-fowler: It was commented out because I wanted to talk to you about it first. I wasn't sure I was doing the right thing as I had no way to test it. Ok will change to `withCString`

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,127 @@
...
+    ///
+    /// - Parameter queue: Dispatch queue to run `sec_protocol_options_set_verify_block` on.
+    /// - Returns: Equivalent NWProtocolTLS Options
+    func getNWProtocolTLSOptions() -> NWProtocolTLS.Options {
+        let options = NWProtocolTLS.Options()
+        
+        // minimum TLS protocol
+        if #available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *) {
+            sec_protocol_options_set_min_tls_protocol_version(options.securityProtocolOptions, self.minimumTLSVersion.nwTLSProtocolVersion)
+        }
```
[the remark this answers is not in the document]
adam-fowler: Why would we use the deprecated version over the new one?

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,127 @@
...
+        
+        // maximum TLS protocol
+        if let maximumTLSVersion = self.maximumTLSVersion {
+            if #available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *) {
+                sec_protocol_options_set_max_tls_protocol_version(options.securityProtocolOptions, maximumTLSVersion.nwTLSProtocolVersion)
+            } else {
+                precondition(self.maximumTLSVersion != nil, "TLSConfiguration.maximumTLSVersion is not supported")
```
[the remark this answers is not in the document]
adam-fowler: see above

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
Lukasa: We wouldn't, we'd use it as a fallback if the `#available` returns `false`.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,127 @@
...
+        }*/
+
+        // the certificate chain
+        if self.certificateChain.count > 0 {
+            preconditionFailure("TLSConfiguration.certificateChain is not supported")
+        }
+        
+        // cipher suites
+        if self.cipherSuites.count > 0 {
+            //preconditionFailure("TLSConfiguration.cipherSuites is not supported")
```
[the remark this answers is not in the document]
adam-fowler: I did have an attempt at this, and got quite far but then hit an issue which would have taken time to get around and I wasn't sure it was worth the development time.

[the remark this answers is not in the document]
Lukasa: apple/swift-nio-ssl#207.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
adam-fowler: I assume this would be done with `sec_protocol_options_set_verify_block`. There doesn't seem to be any documentation on what the default verify block would involve, outside of evaluating the trust (Does it also verify server address?). So wasn't keen on setting one up outside the very specific case of having no certificate verification.

Lukasa: That's correct. You start out with a trust that's appropriately configured (it's passed in, I believe) and you can modify it as needed.

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: Thinking about this, I'm wondering if there is a case for a change in NIOTS. The reason I set it to `false` was DNS errors were not being returned, I would get a timeout instead. But if you are getting a DNS error I assume you havent reached the server and NIOTS should close the connection at that point instead of waiting.

Lukasa: We can consider trying to provide a more expressive timeout error in NIOTS, but in general I think waiting for connectivity is the right choice. As I noted in our discussion over on the NIOTS repository, DNS errors are not final: they may change as network state changes.

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: This was suggested by @weissi and @artemredkin. For proxies if needed we enable the TLS at a later point using `channel.pipeline.addSSLHandlerIfNeeded`. Although we maybe creating two `NIOSSLContext`, one to place in the `NIOClientTCPBootstrap` and then the one that is actually used. @weissei Should I replace the one in `NIOClientTCPBootstrap` with a `NIOInsecureNoTLS`

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
adam-fowler: ok fair point

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: That isn't what NIOSSL does though, connection errors due to DNS return almost immediately

Sources/AsyncHTTPClient/Utils.swift
Lukasa: I think the need to add the later TLS in `addSSLHandlerIfNeeded` is fine, but we can't use the abstract bootstrap protocol to do it, so it seems unnecessary to add a TLS provider here.

Sources/AsyncHTTPClient/Utils.swift
Lukasa: Regular NIO doesn't do it because it can't: it doesn't have sufficient insight into the network to know. In principle it will start to do things like this, though, because if we add more support for streaming service discovery mechanisms those mechanisms may not necessarily have a definitive final "this service does not exist" error state, much as DNS does not.

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: Yes you are correct

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: Ok so timeout is considered the correct error for a DNS fail?

adam-fowler: If I revert this back to `true`. I don't get the DNS error as expected, but I also don't get ECONNREFUSED when I attempt to connect to closed port.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
adam-fowler: The deprecated fallback uses a type SSLProtocol which includes v1, v2, v3 which doesn't really map to `TLSVersion` which v1.0 - v1.3.

Sources/AsyncHTTPClient/Utils.swift
Lukasa: > Ok so timeout is considered the correct error for a DNS fail?

It's substantially more complex than that. I think providing more-detailed timeouts would be useful: if we're in `state` `.waiting(error)` then we should provide that error as part of the information when we fire that timeout.

> If I revert this back to true. I don't get the DNS error as expected, but I also don't get ECONNREFUSED when I attempt to connect to a closed port.

Correct, NIOTS hides all the lower-level information. In some ways, though, that information is not super useful. The result is just "I couldn't connect, sorry". Again, though, I'm happy for us to add more detail to the error NIOTS returns.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
Lukasa: It also includes `.tlsProtocol1`, `.tlsProtocol11`, `.tlsProtocol12`, and `.tlsProtocol13`.

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: I think that would be useful. It could inform whether a user tries to connect again.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
adam-fowler: Ah, I looked at the docs and just looked at the top of the page.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
adam-fowler: Would this use ssl_create_cipher_list(). Are the cipher ids consistent across libraries?

Sources/AsyncHTTPClient/Utils.swift
Lukasa: So the way this is _intended_ to be used is that the user shouldn't try to connect again: they should set a timeout as long as they are willing to wait for a connection. The analogy here is that trying to connect, looking at the error, and maybe trying again is essentially analogous to polling the network. A better model is to try to be reactive: let the OS make decisions about when it is potentially worth retrying a connection attempt, and let it do so.

adam-fowler: ok, now you need to make NIOSSL do the same :-)
I'll change NIOTS code to reflect this

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
Lukasa: It can't, that function is private. It would have to create a throwaway `SSL_CTX` and call `SSL_CTX_set_cipher_list` followed by `SSL_CTX_get_ciphers`. For BoringSSL you want to get the cipher _value_, not the cipher _id_, because the cipher value is the UInt16 value associated with that cipher in the wire protocol. That is consistent across libraries.

Sources/AsyncHTTPClient/Utils.swift
Lukasa: Note that this is not NIOSSL's job: it's _NIO_'s job. And we cannot currently make NIO do this very thoroughly because the OS doesn't tell us these things.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
adam-fowler: sorted

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
adam-fowler: done

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
adam-fowler: done

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: removed setting `waitForActivity` to `false`

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: When running a proxy, I've setup the `NIOClientTCPBootstrap` with `tls` of `NIOInsecureNoTLS`

adam-fowler:
> I suspect the tests should simply be skipped: as there's no unclean shutdown reporting, these tests are kind of moot.

I'm ignoring these tests if running NIOTS

Package.swift
adam-fowler: Using swift-nio-transport-services 1.5.1

adam-fowler:
@Lukasa I think that's all the changes you've requested dealt with.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,140 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2020 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2020 Apple Inc. and the AsyncHTTPClient project authors
```

as this is a brand new file

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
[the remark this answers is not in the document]
Lukasa: I'd rather not have the commented out code: can we just add a TODO with a link to the relevant issue?

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+            preconditionFailure("TLSConfiguration.certificateChain is not supported")
+        }
+        
+        // cipher suites
+        if self.cipherSuites.count > 0 {
+            //preconditionFailure("TLSConfiguration.cipherSuites is not supported")
+        }
+        
+        // key log callback
+        if let _ = self.keyLogCallback {
```
Lukasa, line 101:
```suggestion
        if self.keyLogCallback != nil {
```

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+            preconditionFailure("TLSConfiguration.keyLogCallback is not supported")
+        }
+        
+        // private key
+        if let _ = self.privateKey {
```
Lukasa, line 106:
```suggestion
        if self.privateKey != nil {
```

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -47,21 +53,86 @@ public final class HTTPClientCopyingDelegate: HTTPClientResponseDelegate {
...
+        #else
+        if let clientBootstrap = ClientBootstrap(validatingGroup: eventLoop) {
+            bootstrap = try clientBootstrap.makeClientTCPBootstrap(host: host, requiresTLS: requiresTLS, configuration: configuration)
+        } else {
+            preconditionFailure("Cannot create bootstrap for the supplied EventLoop")
+        }
+        #endif
+        
+        // don't enable TLS if we have a proxy, this will be enabled later on
+        if requiresTLS, configuration.proxy == nil {
```
Lukasa, line 112:
```suggestion
        if requiresTLS && configuration.proxy == nil {
```

Only use comma if conditional binding is involved.

artemredkin:

artemredkin:
@adam-fowler some tests are missing from linux test definitions:
```
diff --git a/Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift b/Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift
index c6a8699..0ffeb42 100644
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift
@@ -96,7 +96,7 @@ extension HTTPClientTests {
             ("testPoolClosesIdleConnections", testPoolClosesIdleConnections),
             ("testRacePoolIdleConnectionsAndGet", testRacePoolIdleConnectionsAndGet),
             ("testAvoidLeakingTLSHandshakeCompletionPromise", testAvoidLeakingTLSHandshakeCompletionPromise),
-            ("testAsyncShutdown", testAsyncShutdown)
+            ("testAsyncShutdown", testAsyncShutdown),
         ]
     }
 }
diff --git a/Tests/LinuxMain.swift b/Tests/LinuxMain.swift
index 93e594a..de6695f 100644
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -28,6 +28,7 @@ import XCTest
     XCTMain([
         testCase(HTTPClientCookieTests.allTests),
         testCase(HTTPClientInternalTests.allTests),
+        testCase(HTTPClientNIOTSTests.allTests),
         testCase(HTTPClientTests.allTests),
         testCase(RequestValidationTests.allTests),
     ])
```
could you please add them? you can do it by running:
```
scripts/generate_linux_tests.rb
```
thanks!

adam-fowler:
> @adam-fowler some tests are missing from linux test definitions:
> 
The only tests I added were specific to NIOTS so aren’t really valid on Linux. Do you still want me to add them?

artemredkin:
Files and methods are still visible to swift compiler, so we have to add them, yes

adam-fowler:
@artemredkin Even though they have a #if canImport(Network) around all of them.

artemredkin:
Ah! Thats a very good point. Our linux tests are generated automatically and tests will not succeed if automatically generated linux test descriptions do not match those in the commit. But it seems that our script does not understand `#canImport` macro... Correct solution would be to put those `#canImport` statements inside all NIO-specific methods.

adam-fowler:
> Correct solution would be to put those `#canImport` statements inside all NIO-specific methods.

done

artemredkin:

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
```diff
@@ -0,0 +1,172 @@
...
+
+#if canImport(Network)
+
+    import Network
+    import NIO
+    import NIOHTTP1
+    import NIOTransportServices
+
+    /// Enum containing all the DNS errors
+    public enum NWDNSError: Error {
```
weissi, line 23: do these need to be public? If yes, they should probably get an `HTTPClient*` or something prefix. `NWDNSError` is not unlikely to clash in the future

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
```diff
@@ -0,0 +1,172 @@
...
+        case badKey
+        case transient
+        case serviceNotRunning /* Background daemon not running */
+        case NATPortMappingUnsupported /* NAT doesn't support PCP, NAT-PMP or UPnP */
+        case NATPortMappingDisabled /* NAT supports PCP, NAT-PMP or UPnP, but it's disabled by the administrator */
+        case noRouter /* No router currently configured (probably no network connectivity) */
+        case pollingMode
+        case timeout
+        case defunctConnection /* Connection to daemon returned a SO_ISDEFUNCT error result */
+        case other(DNSServiceErrorType)
```
weissi, line 57: `enum`s with an `other` case usually should really be `struct`s because equality/hashability/extensibility is a problem for enums.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -47,21 +53,85 @@ public final class HTTPClientCopyingDelegate: HTTPClientResponseDelegate {
...
+    fileprivate static func makeBootstrap(
+        on eventLoop: EventLoop,
+        host: String,
+        requiresTLS: Bool,
+        configuration: HTTPClient.Configuration
+    ) throws -> NIOClientTCPBootstrap {
+        let bootstrap: NIOClientTCPBootstrap
+        #if canImport(Network)
+            // if eventLoop is compatible with NIOTransportServices create a NIOTSConnectionBootstrap
+            if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *),
```
weissi, line 85: this doesn't look quite right. If somebody passes you a `MultiThreadedEventLoopGroup` on macOS, even if Network.framework is supported, you'll have to use `ClientBootstrap`.

In other words, here we should just try all the options, one after the other using `validatingGroup:`. Alternatively, you can store what kind of ELG it is and switch over it here.

The problem is that the user can do (on new macOSes):

```
let elg = MultiThreadedELG(...)
HTTPClient(eventLoopGroup: .shared(elg))
```

and that needs to remain valid.

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
adam-fowler: After speaking with @lukasa the needs for the DNS error enum is much reduced. Given NIOTS will be handling them. I could possibly remove this completely

weissi: That'd be even better ofc :)

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: err, we are trying each option one after the other. If attempts to create an `NIOTSConnectionBootstrap` and if that fails it attempts to create a `ClientBootstrap`

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
adam-fowler: I'll remove it

Sources/AsyncHTTPClient/Utils.swift
weissi: sorry @adam-fowler ! I didn't see that the `let tsBootstrap = ...` is part of the `if` above! Ignore this

weissi: maybe an extra empty line after `let tsBootstrap = ...` ?

```
                let tsBootstrap = NIOTSConnectionBootstrap(validatingGroup: eventLoop) {
                let tlsConfiguration = configuration.tlsConfiguration ?? TLSConfiguration.forClient()
```

looks like two normal `let`s...

weissi: didn't see the `{` and the `,` :)

adam-fowler: I'll stick the { on a newline

weissi: @adam-fowler you might get into trouble with SwiftFormat :P. If it works, then cool, if not, let's just disregard this comment.

adam-fowler: I'll run swift format on it and see. I think one of the versions of swift format was actually encouraging this. Maybe the Apple one, which you don't use ..?!?

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
adam-fowler: The other two errors, shall i stick them inside HTTPClient?

Sources/AsyncHTTPClient/Utils.swift
adam-fowler: Yep swift format didn't like it. I've re-formatted it slightly. Also moved the tlsConfiguration down slightly as it doesn't need initialised if we have a proxy.

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
weissi: That sounds like a good plan to me!

adam-fowler: done

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
adam-fowler: this is no longer an issue as I deleted it

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -19,6 +19,22 @@ import NIOConcurrencyHelpers
 import NIOHTTP1
 import NIOHTTPCompression
 import NIOSSL
+import NIOTransportServices
+
+/// Are we testing NIO Transport services
+func isTestingNIOTS() -> Bool {
+    return ProcessInfo.processInfo.environment["ENABLE_TS_TESTS"] == "true"
```
weissi, line 26: @adam-fowler one last question: Could we flip the default here and have the tests run on TS by default? Otherwise I'm sure nobody will ever run them :|

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
weissi: yes, sorry, missed this. Commented elsewhere.

weissi:

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
adam-fowler: Done

weissi:
hit #175

weissi:
retrying

weissi:

adam-fowler:
@weissi thank you.

I do have a fix for #175, but @adtrevor has his own version of the fix in one of his branches I believe. Do you want me to put up a PR to compare?

adtrevor:
@adam-fowler I think the already opened PR #192 should fix #175 while also making the code simpler

adam-fowler:
@adtrevor as I mentioned in #175 if it protects against releaseAssociatedConnection being called multiple times then we're cool.

PR #186, opened 2020-03-30
fix location header lookup

### LINKED ISSUE #185: Headers should be case insensitive
I have run into an issue where the library did not follow up on a `301`. This was when I accessed the website from *wired*: [https://www.wired.com/story/the-science-of-this-pandemic-is-moving-at-dangerous-speeds](url) 

It turns out that this website returns the location header as:

```log
location: https://www.wired.com/story/the-science-of-this-pandemic-is-moving-at-dangerous-speeds/
```
Note that the location is written as `location` and not as `Location`.

In the redirect handler this is managed as:

```swift
guard let location = headers.first(where: { $0.name == "Location" }) else {
     return nil
}
```

According to the RFC: [https://ietf.org/rfc/rfc7230.html#section-3.2](url) the headers are case insensitive. Hence I believe this code should become:

```swift
guard let location = headers.first(where: { $0.name.lowercased() == "location" }) else {
     return nil
}
```

I did not check but I assume there this may be true in other areas as well. The above changed fixed it for me.

I  used version `1.1.0` of the package.

artemredkin:
Fix `Location` header case-sensitive lookup, closes #185 

Motivation:
HTTP Header names are case-insensitive.

Modifications:
Use `.first` method provided by `HTTPHeaders`
Remove `cookieHeaders` method and instead use getter provided by `HTTPHeaders`

Result:
Redirects will now be processed correctly.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
@@ -140,12 +140,8 @@ extension HTTPClient {
 }
 
 extension HTTPClient.Response {
-    var cookieHeaders: [HTTPHeaders.Element] {
-        return headers.filter { $0.name.lowercased() == "set-cookie" }
-    }
-
     /// List of HTTP cookies returned by the server.
     public var cookies: [HTTPClient.Cookie] {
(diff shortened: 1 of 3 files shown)


artemredkin:

PR #189, opened 2020-03-31
HTTPRequest without body: Content-Length shall not be send

fabianfett:
This pr ensures that we don't send the HTTPHeader `Content-Length` if we don't have a body.

The RFC states:
> The presence of a message-body in a request is signaled by the inclusion of a Content-Length or Transfer-Encoding header field in the request's message-headers.
https://tools.ietf.org/html/rfc2616#section-4.3


weissi:

fabianfett:
close in favor of #191

Issue #146, opened 2019-12-18
Delete Request not including request body


prafsoni:
I am seeing that even though I create delete `Request` with body but the request received by server doesn't have the body. Is this not supported ? if so what is the rationale behind it?

weissi:
CC @artemredkin

weissi:
From [RF 7231]( https://tools.ietf.org/html/rfc7231#section-4.3.5)

>    A payload within a DELETE request message has no defined semantics;
>  sending a payload body on a DELETE request might cause some existing
>   implementations to reject the request.

That seems like we do actually comply to the spec because a payload on DELETE has no defined semantics. However, it might be better to pass the body on or to fail the request...

CC @Lukasa

Lukasa:
"No defined semantics" doesn't mean "cannot have", just means that REST doesn't imply anything about what such a payload means. It's totally fine to have one, it's just up to the server and client to agree on what it means.

The actual bug here is in NIOHTTP1. I captured the TCP stream of such a request in Wireshark, and here's what we sent (I just picked a random body I had lying around in a sample project):

```
DELETE /delete HTTP/1.1
Content-Type: application/x-www-form-urlencoded; charset=utf-8
Host: httpbin.org
Connection: close

grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&assertion=ddd
```

This occurs because NIOHTTP1's `sanitizeTransportHeaders` is stripping the content-length header that async-http-client set. The switch statement in `HTTPMethod.hasRequestBody` is overbroad: we should be only forbidding bodies in `TRACE`. `HEAD` and `DELETE` _may_ have bodies, they just have no defined meaning, so they should go in `.unlikely`. That will fix this issue we see here.

weissi:
@Lukasa my reading of “no defined semantics” was that server & client can do whatever they want. Therefore I was assuming that not sending it is spec compliant (and so is sending it).
I agree that the spec definitely doesn’t say it must not have a body.

Lukasa:
Not sending it is spec compliant but not useful, as there are definitely use-cases that require sending it.

weissi:
Agreed, hence suggesting to either sending it (more useful) or failing it (less useful). I’ll file a NIOHTTP1 bug.

prafsoni:
So it had to do purely with `swift-nio`. This is awesome I really appreciate the fix, as there are scenarios/apis where we require body in `delete`

[...]

PR #191, opened 2020-03-31
Verify header field names

fabianfett:
This pr builds on top of #189 and will need a rebase once the former is merged.

### Motivation

Having spent about four hours today debugging AsyncHTTPClient just to find out I had a whitespace instead of a dash in my http header field name, I decided I want to save somebody else's time – probably future me.

`User Agent` vs `User-Agent`

### Changes

- Added a method: `HTTPHeaders.validateFieldNames() throws`
- Added an error `.invalidHeaderFieldNames([String])`

### Result

Hours of time saved. Karma for me.


Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -29,28 +29,35 @@ extension HTTPHeaders {
...
-                throw HTTPClientError.chunkedSpecifiedMultipleTimes
-            }
+        try self.validateFieldNames()
+
+        guard let body = body else {
+            // https://tools.ietf.org/html/rfc2616#section-4.3
+            //
+            // The presence of a message-body in a request is signaled by the inclusion of a
+            // Content-Length or Transfer-Encoding header field in the request's message-headers.
+            return
```
Lukasa: RFC 2616 is dead, let's stop citing it. It's RFC 7230 section 3.3. _However_, this change is not cautious enough: if there's no body, but the request verb is `POST`, "Content-Length: 0" is mandatory. Content-Length: 0 is also harmless on all request types that do not define a body. A patch that users this escape hatch on only `GET`, `HEAD`, `DELETE`, `CONNECT`, and `TRACE` would be acceptable though, so long as it sends C-L 0 everywhere else.

fabianfett:
@Lukasa Compile errors are fixed. There are some failing tests though (nothing I touched).

artemredkin:
@fabianfett it seems that most tests fail due to change in behaviour, `Content-Length: 0` was expected in those tests, but now it's missing

fabianfett:
@artemredkin ow shoot. fixed.

artemredkin:
awesome, thanks! there is one failing test, but this one is flaky (#175), I'll try to run tests again

artemredkin:

Tests/AsyncHTTPClientTests/RequestValidationTests.swift
```diff
@@ -0,0 +1,95 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
weissi, line 5: @fabianfett can we fix the copyright here? should be `2020` because it's all new code

weissi:
@fabianfett From what I can tell, we're good but would you mind checking if we have a test case which covers this https://github.com/swift-server/async-http-client/issues/146 ? If yes, we're all good, if no we might want to add it here?

Tests/AsyncHTTPClientTests/RequestValidationTests.swift
```diff
@@ -0,0 +1,107 @@
...
+            ("Transfer-Encoding", "chunked"),
+        ])
+        var buffer = ByteBufferAllocator().buffer(capacity: 200)
+        buffer.writeBytes([UInt8](repeating: 12, count: 200))
+        XCTAssertThrowsError(try headers.validate(method: .TRACE, body: .byteBuffer(buffer))) {
+            XCTAssertEqual($0 as? HTTPClientError, .traceRequestWithBody)
+        }
+    }
+
+    func testGET_HEAD_DELETE_CONNECTRequestCanHaveBody() {
```
fabianfett, line 74: @weissi this is for you.

Tests/AsyncHTTPClientTests/RequestValidationTests.swift
fabianfett: fixed

artemredkin:

artemredkin:

artemredkin:

Tests/AsyncHTTPClientTests/RequestValidationTests.swift
weissi: thanks so much!

artemredkin:
thank you @fabianfett !

Issue #188, opened 2020-03-31
HTTPClient.Task.fail can release its associated connection more than once 


adam-fowler:
This occurs on the NIOTransportServices PR #184. The testStressGetClose() test fails. 

It is related to NIOTS returning two errors when it tries to read off a closed channel. HTTPClient.Task.fail is called twice. This then causes connection.release() being called twice which messes up with the connection lease count.

PR #192, opened 2020-04-03
refactor pool

### LINKED ISSUE #175: flaky test: HTTPClientTests.testRaceNewRequestsVsShutdown: Invalid number of leased connections (-1)

artemredkin:
Ideas for connection pool refactoring. Main idea is that we need to simplify connection lifecycle. Right now its spread over many callbacks, that return some kind of potential action. I think this part can be simpler. There are basically two operations, polling a connection and releasing the connection. All magic should as much as possible be contained in those two methods.
One thing I'm still figuring out is how to simplify idle connection handling.
Closes #175 and #176

There are two things missing at the moment:
1. removing connections from pool and removing providers when they are idle (this is caught by the failing test)
2. there are some verifications needed for event loop preference (and possibly tests?)

Other things to check:
1. what about inflight connection requests that are not finished yet (+test?)
  we assert on `openedConnectionCount` on deinit, so we good there, only question is should we cancel those requests on shutdown or just let them fail when we stop `eventLoop`
2. Code handling idle pool connections is tough to follow, but I haven't found a way to simplify it yet
3. Same for code handling when channel is inactive
4. we probably need to move `IdlePoolConnectionHandler` stuff to connection pool file, otherwise we add idle handler in one place and remove in another


Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -265,114 +234,132 @@ final class ConnectionPool {
...
-                            self.state.defaultEventLoop
+            let promise = self.eventLoop.makePromise(of: Connection.self)
+
+            self.lock.withLockVoid {
+                if let connection = self.availableConnections.popFirst() {
+                    connection.isLeased = true
+                    // check if we can vend this connection to caller
+                    connection.cancelIdleTimeout().flatMapError { error in
+                        if error is Connection.InactiveChannelError {
+                            return self.makeConnection(on: preference.bestEventLoop ?? self.eventLoop)
```
weissi, line 307: are we _sure_ this doesn't call out immediately? If anything in there succeeds or fails a future, this isn't valid as we'd call out with the lock held.

artemredkin:
cc @adtrevor

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: Same question applies to _all_ other calls we make with the lock held. I think the idea of the `struct State` was to have a place where all the computation goes (under a lock). The computation would then return what the next correct step is that needs to be performed (not under a lock).

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -385,314 +372,62 @@ final class ConnectionPool {
...
-                let waitersCopy = self.state.waiters
-                self.state.waiters.removeAll()
-                let waitersPromises = waitersCopy.map { $0.promise }
-                let waitersFutures = waitersPromises.map { $0.futureResult }
-                waitersPromises.forEach { $0.fail(HTTPClientError.cancelled) }
-                let closeFutures = self.state.availableConnections.map { $0.close() }
-                return (waitersFutures, closeFutures)
+        func close() {
+            self.lock.withLockVoid {
+                self.waiters.forEach { $0.promise.fail(HTTPClientError.cancelled) }
```
weissi, line 398: this is definitely incorrect. Here we call out under a lock and that must not be the case (and if the test suite should pass we need to add a test here). If anyone were to re-call into AHC from the future that is failed here, we'd get a dead-lock.

If possible, I'd organise the code in a way that it's clear that we're never doing anything to promises whilst the lock is held. A state machine is probably the right tool to make sure that we can safely (under the lock) be transitioned, then returns the action, and finally performs the action (not under a lock).

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -385,314 +372,62 @@ final class ConnectionPool {
...
-                    }
-                }
-        }
-
-        func close() -> EventLoopFuture<Void> {
-            let availableConnections = self.stateLock.withLock { () -> CircularBuffer<ConnectionPool.Connection> in
-                assert(self.state.activity == .closing)
-                return self.state.availableConnections
+            self.lock.withLock {
+                self.availableConnections.forEach { $0.close() }
```
weissi, line 403: `$0.close()` might also be calling out

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -385,314 +372,62 @@ final class ConnectionPool {
...
-                case .indifferent:
-                    return (self.defaultEventLoop, false)
-                case .delegate(let el):
-                    return (el, false)
-                case .delegateAndChannel(let el), .testOnly_exact(let el, _):
-                    return (el, true)
-                }
-            }
-
-            fileprivate enum ConnectionGetAction {
```
weissi, line 648: These "actions" seem like the right tool to return so that outside of the lock we can then perform the actions.

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: I did hit that (calling out and then deadlocking, see line 328), so yes, I probably will revert some of that state/action stuff...

weissi: Yeah, I know it's annoying but it's the only way to retain sanity: Separate the state from the actions, and then it'll be cool.

A good way to spot bugs is: If an action is returned but the state is _not_ changed, it's probably a bug. Because someone else could then race in and get the same action before the first action has run (and advanced the state with requiring the lock). By the time the second action is then executing the state has then changed under its feet and problems happen.

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: you are right, they are indeed right tool, but I think that State struct maybe not that useful (give a few more hours :D)

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: ah, I did miss that, thanks, I'll think about the state machine here, for now I moved stuff out of the locks

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: same as above

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: Yep, they are indeed

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -65,18 +65,12 @@ final class ConnectionPool {
     func getConnection(for request: HTTPClient.Request, preference: HTTPClient.EventLoopPreference, on eventLoop: EventLoop, deadline: NIODeadline?) -> EventLoopFuture<Connection> {
         let key = Key(request)
 
-        let provider: HTTP1ConnectionProvider = self.connectionProvidersLock.withLock {
-            if let existing = self.connectionProviders[key] {
-                existing.stateLock.withLock {
-                    existing.state.pending += 1
```
adtrevor, line 71: What will be replacing this `pending` property?

adtrevor: One way (but I don't know if that's still applicable with the refactor) is to expose the get connection action outside of the provider so that there doesn't need to be an additional property indicating wether or not there are still pending connections.
Like many other places I would have liked to be able to hold a lock for just a bit longer but IIRC that wasn't possible for the same potential deadlock issues you discussed with @weissi

adtrevor:
@artemredkin Thank you very much for this!
> 1. what about inflight connection requests that are not finished yet (+test?)
we assert on openedConnectionCount on deinit, so we good there, only question is should we cancel those requests on shutdown or just let them fail when we stop eventLoop

I think we can't rely on the event loop correctly notifying remaining futures that it is shutdown currently, IIRC this was the related issue: https://github.com/apple/swift-nio/issues/1309, the best option would be to fail the requests on shutdown IMO, it also gives a more consistent behaviour between client owned EL and user provided EL I guess?

> 2. Code handling idle pool connections is tough to follow, but I haven't found a way to simplify it yet
> 3. Same for code handling when channel is inactive

Tell me if something doesn't make sense, I'll hopefully still remember the reasoning behind them 🙂 

> 4. we probably need to move IdlePoolConnectionHandler stuff to connection pool file, otherwise we add idle handler in one place and remove in another

 Agreed, that's not clean enough

adtrevor:
I hit this assertion in `testRaceNewRequestsVsShutdown`:

> Test Suite 'Selected tests' started at 2020-04-04 05:25:47.352
> Test Suite 'AsyncHTTPClientTests.xctest' started at 2020-04-04 05:25:47.354
> Test Suite 'HTTPClientTests' started at 2020-04-04 05:25:47.354
> Test Case '-[AsyncHTTPClientTests.HTTPClientTests testRaceNewRequestsVsShutdown]' started.
> Fatal error: ioOnClosedChannel: file /Users/adtrevor/Development/AsyncHTTPClient/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift, line 525
> 2020-04-04 05:25:48.502101+0200 xctest[1407:29744] Fatal error: ioOnClosedChannel: file /Users/adtrevor/Development/AsyncHTTPClient/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift, line 525

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -265,114 +234,132 @@ final class ConnectionPool {
...
-                    switch result {
-                    case .success(let connection):
-                        promise.succeed(connection)
-                    case .failure(let error):
-                        promise.fail(error)
+            self.lock.withLock {
+                if connection.isActiveEstimation { // If connection is alive, we can give to a next waiter
+                    if let waiter = self.waiters.popFirst() {
+                        // TODO:
+                        connection.channel.eventLoop.execute {
```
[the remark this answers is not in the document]
weissi: Yes, this is the correct solution. You could do it only ever 100 calls or something if you wanted but just `.execute` sounds good for now. Do we have a test that covers this?

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: this property is replaced by `openedConnectionsCount`, we increase it every time we start a new connection and decrease it when connection goes away. This allows to achieve two things:
1. is easy to check for how many connections we have - either to check that we do not exceed the limit or if we closed everything when shutting down
2. this allows us to use `availableConnections` as a pool - we pop active connections for usage and offer them when we are done, I think it might be a liittle bit cleaner that iterating over connections to find "available". Additionally, we can implement different vending options - FIFO or LIFO

Sources/AsyncHTTPClient/ConnectionPool.swift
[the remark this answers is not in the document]
artemredkin: Two followup questions:
1. by "every 100 calls" do you mean I should track it manually? or there is a way to get this info from NIO?
2. Test would be good, I'm just not sure how can I test this?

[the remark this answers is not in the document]
weissi: @artemredkin 

1. yes, manually. NIO doesn't give you that at the moment, we could add a feature to measure the stack depth. We already got an issue for that. At the moment you'd need to track manually, I'd just `.execute` for now

2. couldn't you just open 3000 connections or so, then verify that it crashes without the `.execute`. Then we know we have a test and then we can add the `.execute` back.

Sources/AsyncHTTPClient/ConnectionPool.swift
adtrevor: @artemredkin  It's great `openedConnectionsCount` completely avoids all close callbacks, I think that alone may have been the cause of some of the bugs found in the flaky tests.

Note there might be a potential issue with `getConnection` (ConnectionPool.swift, L65) once the provider removal is implemented, because in:
```swift
let provider: HTTP1ConnectionProvider = self.lock.withLock {
    if let existing = self.providers[key] {
        return existing
    } else {
        let http1Provider = HTTP1ConnectionProvider(key: key, eventLoop: eventLoop,
configuration: self.configuration, pool: self)
        self.providers[key] = http1Provider
        return http1Provider
    }
}
return provider.getConnection(preference: preference)
```

The issue is that once we stop holding the lock and use `provider` (on the `return` line) there is nothing synchronising it anymore, so before we call the `getConnection` method on it it doesn't know there is a new tasks that asked it for a connection, and if at this point in time it were to be removed from the pool from another thread, we would be getting our connection on an already removed provider, that was what `pending` was attempting to avoid, but that wasn't elegant at all.

artemredkin: Yep, you are correct, to handle that I've added `active: Bool` property

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -73,7 +73,6 @@ public class HTTPClient {
     }
 
     deinit {
-        assert(self.pool.connectionProviderCount == 0)
```
artemredkin, line 76: I wonder how important this is, we call `cancel` on all active tasks but do not wait for it to propagate to provider's `delete`...

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -265,434 +237,255 @@ final class ConnectionPool {
...
-                    case .success(let connection):
-                        promise.succeed(connection)
-                    case .failure(let error):
-                        promise.fail(error)
+            case .lease(let connection, let waiter):
+                // check if we can vend this connection to caller
+                connection.cancelIdleTimeout().flatMapError { error in
+                    // if connection is already inactive, we create a new one.
+                    if error is Connection.InactiveChannelError {
+                        self.openedConnectionsCount += 1
```
weissi, line 320: This is missing the lock, no? Could we have a test case that fails under TSan, showing that this isn’t right?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -265,434 +237,255 @@ final class ConnectionPool {
...
-                case .none:
-                    break
+        func getConnection(preference: HTTPClient.EventLoopPreference) -> EventLoopFuture<Connection> {
+            let waiter = Waiter(promise: self.eventLoop.makePromise(), preference: preference)
+
+            let action: Action = self.lock.withLock {
+                if let connection = self.availableConnections.popFirst() {
+                    return .lease(connection, waiter)
+                } else if self.openedConnectionsCount < self.maximumConcurrentConnections {
+                    self.openedConnectionsCount += 1
```
weissi, line 340: This seems racy. We take a decision under the lock but then return without changing the state, so before we actually do the action, the circumstances may change. Or is this race okay?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -85,37 +79,34 @@ final class ConnectionPool {
     }
 
     func release(_ connection: Connection) {
-        let connectionProvider = self.connectionProvidersLock.withLock {
-            self.connectionProviders[connection.key]
+        let connectionProvider = self.lock.withLock {
```
weissi, line 82: Almost certainly, this is racy. I’m pretty sure release could be called twice (eg pool shutdown vs ordinary connection release). Both could then call release with the same connection and that probably leads to bad things. Am I missing something?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -265,434 +237,255 @@ final class ConnectionPool {
...
+            return channel.flatMap { channel -> EventLoopFuture<Connection> in
+                channel.pipeline.addSSLHandlerIfNeeded(for: self.key, tlsConfiguration: self.configuration.tlsConfiguration, handshakePromise: handshakePromise)
+                return handshakePromise.futureResult.flatMap {
+                    channel.pipeline.addHTTPClientHandlers(leftOverBytesStrategy: .forwardBytes)
+                }.map {
+                    let connection = Connection(key: self.key, channel: channel, pool: self.pool)
+
+                    channel.closeFuture.whenComplete { _ in
+                        self.delete(connection: connection)
                     }
```
weissi, line 429: I don’t think this is enough. We might never call `delete` if we fail before setting up this close handler. Basically if we skip the whole `map` because the future is failed already.

We need a test for this. A connection that is closed before this runs should do it

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: what do you mean by not changing state? we do increase the counter if we will be creating another connection. We only skip state change if we reuse connection

artemredkin: Actually, we do change:
1. if we reuse - we pop from available connections,
2. if we create - we increase the connection count
3. if we cannot reuse or create - we append to waiters

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: not sure I understand, if connection is closed before this we do not create the connection at all and there is no need to delete it. This callback is only needed to clean up the `Connection`, not the `Channel`

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: good catch, I think so, yes

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: I will test this

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: Sorry @artemredkin , not sure what I was thinking (or I can't see the problem I meant anymore) :)

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: @artemredkin what if it fails between `channel.flatMap` and us adding `channel.closeFuture.whenComplete`? So if adding the SSL handler fails or so? Is there no cleanup to do then?

Sources/AsyncHTTPClient/HTTPClient.swift
weissi: @artemredkin don't the providers reference the `HTTPClient`?

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: done

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: this will not be called twice, since we are not calling it on pool shutdown, on pool shutdown we call close on connections and tasks, and they as part of their shutdown call release, does that make sense?

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: I think no. There are two cleanup steps we need to do if connection bootstap fails:
1. decrease `openConnectionCount` - is executed always in `flatMapError`, so it will be called in the scenario you describe
2. delete the `Connection` object, but this is only needed if it was added.
Does this sound reasonable?

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: No, providers only reference the pool, not the client

artemredkin: Ok, I got it.

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: 👍

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: sounds good, yes

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -163,73 +152,57 @@ final class ConnectionPool {
...
 
         /// The `Channel` of this `Connection`
         ///
         /// - Warning: Requests that lease connections from the `ConnectionPool` are responsible
         /// for removing the specific handlers they added to the `Channel` pipeline before releasing it to the pool.
         let channel: Channel
-
-        /// Wether the connection is currently leased or not
-        var isLeased: Bool = false
+        let closeFuture: EventLoopFuture<Void>
```
weissi, line 172: isn't this just

```swift
var closeFuture: ELF<Void> {
    return self.channel.closeFuture
}
```

?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -163,73 +152,57 @@ final class ConnectionPool {
...
+        /// Release this `Connection` to its associated `HTTP1ConnectionProvider` in the parent `ConnectionPool`
+        ///
+        /// - Warning: This only releases the connection and doesn't take care of cleaning handlers in the `Channel` pipeline.
+        func release() {
+            self.pool.release(self)
+        }
 
-        var closeFuture: EventLoopFuture<Void>
+        func close() -> EventLoopFuture<Void> {
+            self.channel.close(promise: nil)
```
weissi, line 201: why not `return self.channel.close()` ?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -266,114 +239,187 @@ final class ConnectionPool {
...
-        fileprivate let stateLock = Lock()
+        /// - Warning: This lock should always be acquired *after* `ConnectionPool`s `lock` if used in combination with it.
+        private let lock = Lock()
 
         /// The maximum number of concurrent connections to a given (host, scheme, port)
         private let maximumConcurrentConnections: Int = 8
 
-        /// The pool this provider belongs to
-        private let parentPool: ConnectionPool
+        private var active: Bool = true
```
weissi, line 272: This needs a comment and should probably be longer than just `active`. There are so many moving things here, what exactly does this `active` mean?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -266,114 +239,187 @@ final class ConnectionPool {
...
 
         var description: String {
-            return "HTTP1ConnectionProvider { key: \(self.key), state: \(self.state) }"
+            return "HTTP1ConnectionProvider { key: \(self.key) }"
         }
 
-        func getConnection(preference: HTTPClient.EventLoopPreference) -> EventLoopFuture<Connection> {
-            self.activityPrecondition(expected: [.opened])
-            let action = self.stateLock.withLock { self.state.connectionAction(for: preference) }
+        var isActive: Bool {
```
weissi, line 310: I think this is confusing:

- `self.active` when already locked
- `self.isActive` when currently not locked?

Also, as mentioned above: what does "active" mean in this context?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -266,114 +239,187 @@ final class ConnectionPool {
...
+            case .lease(let connection, let waiter):
+                // check if we can vend this connection to caller
+                connection.cancelIdleTimeout().flatMapError { error in
+                    // if connection is already inactive, we create a new one.
+                    if error is Connection.InactiveChannelError {
+                        self.lock.withLockVoid {
+                            self.openedConnectionsCount += 1
                         }
-                        return self.makeConnection(on: preference.bestEventLoop ?? defaultEventLoop)
+                        return self.makeConnection(on: waiter.preference.bestEventLoop ?? self.eventLoop)
```
weissi, line 326: Is this really not racy? If we shut the whole thing down, could this create a new connection after we think we've shut everything?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -405,315 +451,79 @@ final class ConnectionPool {
                     #endif
                     return eventLoop.makeSucceededFuture(())
                 }.map {
-                    let connection = Connection(key: self.key, channel: channel, parentPool: self.parentPool)
-                    connection.isLeased = true
+                    let closePromise = channel.eventLoop.makePromise(of: Void.self)
```
weissi, line 454: why not use `channel.closeFuture`?

artemredkin:

artemredkin:

artemredkin:
This also closes #188

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -153,567 +133,757 @@ final class ConnectionPool {
...
+        enum State {
+            case active
+            case closed
         }
 
-        func release(connection: Connection) {
-            self.activityPrecondition(expected: [.opened, .closing])
-            let action = self.parentPool.connectionProvidersLock.withLock {
-                self.stateLock.withLock { self.state.releaseAction(for: connection) }
+        #if DEBUG
```
weissi, line 247: this won't work if you run tests in release mode, I think we should just remove it.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -153,567 +133,757 @@ final class ConnectionPool {
...
 
-        func close() -> EventLoopFuture<Void> {
-            let availableConnections = self.stateLock.withLock { () -> CircularBuffer<ConnectionPool.Connection> in
-                assert(self.state.activity == .closing)
-                return self.state.availableConnections
+        mutating func drop(connection: Connection) {
+            switch self.state {
+            case .active:
+                self.leasedConnections.remove(connection)
+            default:
```
weissi, line 435: I think we should remove all `default:`s in the state machine, that way you get compile errors anywhere you need to fix stuff when you add a case. [Here](https://github.com/apple/swift-nio-examples/blob/master/backpressure-file-io-channel/Sources/BackpressureChannelToFileIO/FileIOChannelWriteCoordinator.swift#L265-L275)'s an example of who this can still be very concise.

I'd also recommend adding a function like

```
    private func illegalTransition(_ function: String = #function) -> Never {
        preconditionFailure("illegal transition \(function) in \(self)")
    }
```

to get really good error messages. The function name is the "input" to the state machine, therefore you get exactly what you need to debug: The current state + the input :)

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -153,567 +133,757 @@ final class ConnectionPool {
...
+        case closeProvider
+        case park(Connection)
+        case none
+        case fail(Waiter, Error)
+        case cancel(Connection, Bool)
+        indirect case closeAnd(Connection, Action)
+        indirect case parkAnd(Connection, Action)
+    }
+
+    struct ConnectionsState {
```
weissi, line 241: Should we maybe move the state machine to a separate file?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -153,567 +133,757 @@ final class ConnectionPool {
...
-                        promise.fail(error)
-                    }
-                }
+        private var state: State = .active
 
-            case .none:
-                break
-            }
+        /// Opened connections that are available.
+        private var availableConnections: CircularBuffer<Connection> = .init(initialCapacity: 8)
```
weissi, line 264: I think these variables should eventually move into the `State` but I'm happy if we do this in a followup.

For example if `state = .inactive`, what if `leasedConnections` isn't empty. Or `pending != 0` or something like that?

Here's an example state machine like that: https://github.com/apple/swift-nio-examples/blob/master/backpressure-file-io-channel/Sources/BackpressureChannelToFileIO/FileIOChannelWriteCoordinator.swift#L44-L81

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -61,69 +62,48 @@ final class ConnectionPool {
...
+        self.lock.withLockVoid {
+            self.providers[provider.key] = nil
         }
     }
 
-    func prepareForClose(on eventLoop: EventLoop) -> EventLoopFuture<Void> {
-        let connectionProviders = self.connectionProvidersLock.withLock {
-            self.connectionProviders.values
+    func close(on eventLoop: EventLoop) -> EventLoopFuture<Bool> {
+        let providers = self.lock.withLock {
```
weissi, line 90: shouldn't `self.providers` be emptied here? Ideally we'd transition to a state that doesn't have providers but that's maybe for a followup too?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -61,69 +62,48 @@ final class ConnectionPool {
...
-                assert(self.connectionProviders.count == 0, "left-overs: \(self.connectionProviders)")
-            }
-        }
+        return EventLoopFuture.reduce(true, providers.map { $0.close() }, on: eventLoop) { $0 && $1 }
     }
 
     var connectionProviderCount: Int {
-        return self.connectionProvidersLock.withLock {
-            self.connectionProviders.count
+        return self.lock.withLock {
```
weissi, line 98: do we need this `connectionProviderCount` property? Seems racy so I'd just remove it.

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: all providers will delete themselves (and we assert that in client's `deinit`)

Tests/AsyncHTTPClientTests/ConnectionPoolTests.swift
```diff
@@ -0,0 +1,1437 @@
...
+class ConnectionPoolTests: XCTestCase {
+    struct TempError: Error {}
+
+    func testPending() {
+        let eventLoop = EmbeddedEventLoop()
+
+        var state = HTTP1ConnectionProvider.ConnectionsState(eventLoop: eventLoop)
+
+        var snapshot = state.testsOnly_getInternalState()
+        XCTAssertEqual(0, snapshot.availableConnections.count)
```
weissi, line 35: they look all great! Quick question: Why are we asserting on the (private) state and not the `Action`s that the state machine returns? Examples here: https://github.com/apple/swift-nio-examples/blob/master/backpressure-file-io-channel/Tests/BackpressureChannelToFileIOTests/StateMachineTest.swift#L267-L309

For the example above, I extended the `Action` to have lots of functions (one for each of the actions) like

```
    @discardableResult
    func assertOpenFile(_ check: (String) throws -> Void,
                        file: StaticString = #file,
                        line: UInt = #line) -> Self {
        if case .openFile(let path) = self.main {
            XCTAssertNoThrow(try check(path))
        } else {
            XCTFail("action \(self) not \(#function)", file: file, line: line)
        }
        return self
    }
```

And then you can just write

```
self.stateMachine.someInput()
    .assertActionIsX()
    .assertWhateverElse()
```

and we wouldn't need to expose the state. But again, I think that's something we can do in the future.

The benefit of asserting on the `Action` is that we can change the state machine representation to say an `enum` without rewriting all the tests.

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: we use it one place - client deinit to assert that all providers are deleted (so no decisions based on it are made)

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: I don't think you can run our tests in release mode at all, at least most of them, due to `@testable` import, no? but, yeah, I'll remove it, I just wanted to make sure those methods cannot be used in the "real" code... maybe it was a bit of an overreaction 😄

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: @artemredkin ok, what prevents a second call to `close(on:)`?

weissi: Also, wouldn't it be better to just remove them here? What's the benefit of them deleting themselves?

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: @artemredkin in `deinit` we don't need the lock. So we can just do

```
assert(self.providers.count == 0)
```

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: Yes, you can :) `swift build -c release -Xswiftc -enable-testing`. The NIO release tools run them for every project in the following confugurations:

- debug
- ASan in _release_ mode (ASan works MUCH better in Swift when run in release mode)
- TSan in debug mode
- in docker in debug mode

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: second call to close is prevented by the client, and since providers already delete themselves when they are empty (which is basically what close means, we close/fail all connections waiters and suddenly provider is empty so it deletes itself)

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: 👍

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: huh, did not know that, don't usually run tests from command line 😆

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: I though about that, but decided to leave it out for now, as this already taken way too much time, there are other things I want to improve here as well...

weissi: @artemredkin yes, good call. We can followup

Tests/AsyncHTTPClientTests/ConnectionPoolTests.swift
artemredkin: we do assert on actions as well (not in this test specifically, but in other tests), I want to test not only action but the way state is changing. I test before the action to "show" what state I was expecting in this test and after to see if we arrived at it, also serves as a documentation of sorts for the test case

weissi: Hmm, but the tests shouldn't care how the state machine achieves its goal, assuming it does, right? When we refactor, we'll lose all the 'private state tests'.
The state machine is fully deterministic, so there isn't anything you can't test with just inputs & actions. The state is fully determined by the inputs that have been fed.

As long as most of the tests do only action assertions, I think there's no harm caused also asserting on private state but I think it's a bit of an anti-pattern.

artemredkin: I agree, but it can be a bit easier to test that when we `aqcuire` a connection that it is inside the waiters for example, than to get a series of transformations to `aquire` and then release some other connection and then test that our promise was succeeded

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: done

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: good idea, done, thanks!

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: done

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: done

Tests/AsyncHTTPClientTests/ConnectionPoolTests.swift
weissi: Github’s broken, i was meaning to thumbs UP only but fat fingered a DOWN too which I can’t remove now 🤪

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -61,69 +62,46 @@ final class ConnectionPool {
...
                 return existing
             } else {
-                let http1Provider = HTTP1ConnectionProvider(key: key, eventLoop: eventLoop, configuration: self.configuration, parentPool: self)
-                self.connectionProviders[key] = http1Provider
-                http1Provider.stateLock.withLock {
-                    http1Provider.state.pending += 1
-                }
-                return http1Provider
+                let provider = HTTP1ConnectionProvider(key: key, eventLoop: eventLoop, configuration: self.configuration, pool: self)
+                self.providers[key] = provider
```
Lukasa, line 75: Are we not missing a call to `enqueue` here?

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -153,567 +131,429 @@ final class ConnectionPool {
...
-        ///     - parentPool: The pool this provider belongs to
-        init(key: ConnectionPool.Key, eventLoop: EventLoop, configuration: HTTPClient.Configuration, parentPool: ConnectionPool) {
-            self.eventLoop = eventLoop
-            self.configuration = configuration
-            self.key = key
-            self.parentPool = parentPool
-            self.state = State(eventLoop: eventLoop, parentPool: parentPool, key: key)
+    /// Sets idle timeout handler and channel inactivity listener.
+    func setIdleTimeout(timeout: TimeAmount?) {
+        _ = self.channel.pipeline.addHandler(IdleStateHandler(writeTimeout: timeout)).flatMap { _ in
```
Lukasa, line 197: Idle handlers should usually go first in the pipeline.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -153,567 +131,429 @@ final class ConnectionPool {
...
+        return self.removeHandler(IdleStateHandler.self).flatMap { _ in
+            self.removeHandler(IdlePoolConnectionHandler.self)
         }
+    }
+}
 
-        var description: String {
-            return "HTTP1ConnectionProvider { key: \(self.key), state: \(self.state) }"
-        }
+extension Connection: Hashable {
```
Lukasa, line 210: Minor nit, but I'd rather not conform something to `Hashable`/`Equatable` if its definition of hash value and equality is identity.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -153,567 +131,429 @@ final class ConnectionPool {
...
+            }
+        case .create(let waiter):
+            self.makeChannel(preference: waiter.preference).whenComplete { result in
+                switch result {
+                case .success(let channel):
+                    self.connect(using: channel, waiter: waiter)
+                case .failure(let error):
+                    self.connectFailed(error: error, waiter: waiter)
+                }
+            }
```
Lukasa, line 298: This block is repeated a few times, it seems like we could factor it out a bit.

Sources/AsyncHTTPClient/ConnectionPool.swift
Lukasa: Ok, after reading through the entire PR I finally found out why we aren't. Can we add a comment here that states that providers begin at `+1`?

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: Why? I what would you recommend instead?

Lukasa: Users explicitly testing for identity and explicitly using `ObjectIdentifier` as a hash key.

The main reason I object is that Swift pretty pervasively separates the idea of equality and identity. Indeed, the [`Equatable` docs call this out directly](https://developer.apple.com/documentation/swift/equatable#2847781). In particular, being equatable implies that two objects are substitutable. If two objects are only equal if they are literally the same object, we have a way to express that already and do not need to bring equatability into the mix.

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: Even if the timeout handling handler is added immediately after it? I don't think there are any handlers that can (or should) handle timeouts in the pipeline at this point...

Lukasa: The reason they should go first in the pipeline is so that they observe the TCP writes/reads, rather than something more structured that may be delayed. Otherwise slow connections risk being killed for being idle, when in practice there's data flow occurring and just not being seen due to intermediate buffering.

This function is potentially callable at any time, so we'll risk fewer bugs by explicitly calling out that the idle state handler should go first. I'm fine for the actual timeout handler to go wherever.

artemredkin: Ah! That makes sense, done

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: Good idea, done, thanks!

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: Done!

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: I don't think there is a good way to use `ObjectIdentifier` is a `Set`, while I can add and test for `contains` and remove, since `ObjectIdentifier` is not generic, I cannot say `leased.map { $0.close() }`

Lukasa: Do you need a `Set`? Why not just `Array`?

artemredkin: Main usage - is testing if given connection is in `leased` set

Lukasa: Ok, so let's step back. We have three options:

1. Make `Connection` `Hashable`.
2. Put `ObjectIdentifier` into the `leased` `Set`.
3. Change the `leased` structure from `Set` to `Array`.

My inclination is that either (2) or (3) is preferable to (1).

(2) is acceptable if you never need to iterate the `leased` `Set` to grab `Connection`s from it. If you do, then `Array` works just fine. In general a linear `Array` search is _very_ fast, and likely to be faster than hashing for small numbers of elements. The only reason it'd be a concern is if you are having to add-and-remove elements a lot, and even there if `leased` is small then I think it's really not a big problem.

weissi: @artemredkin you can always make a wrapper struct that implements identity hashable/equals

weissi: (an ad-hoc one that you just use here)

weissi: or use a dictionary `[ObjectIdentifier: Connection]`.

weissi: @Lukasa and 4) (which I think may be best):

```
fileprivate struct LeasedConnectionKey: Hashable {
    private let connection: Connection

   func ==(...) {}
   func hash...
}
```

Lukasa: Sure, works for me.

artemredkin: I do add and remove connections constantly, its added on every new http call and removed when call is finished. This is a result from a simple benchmark (adds, removes, searches) on a small (8 element) array/set:
```
name   time       std         iterations  
----------------------------------------
array  3208.0 ns  ±  29.53 %  425528      
set    1677.0 ns  ±  52.07 %  827306 
```
In this case set is almost twice as fast...

I think I'll go with 4

weissi: Yeah, the problem with Array will be the removals of elements that are not the last...

artemredkin: done

PR #201, opened 2020-04-13
Enable clients to call URLs that include %2F as an escaped backslash

iainsmith:
Hi 👋, thanks for all the great work on this project.

I have a package that calls the Travis API, which uses %2F as part of the URL path segment. This PR adds a test case & migrates to `URLComponents(url: self, resolvingAgainstBaseURL: false)?.percentEncodedPath` which covers that use case.

I also updated the test server to switch on the `percentEncodedPath` so it's easier
to understand the desired behaviour. I've tested this locally on 10.15.4 but not on Ubuntu.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -464,7 +464,7 @@ extension URL {
         if self.path.isEmpty {
             return "/"
         }
-        return self.path.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? self.path
+        return URLComponents(url: self, resolvingAgainstBaseURL: false)?.percentEncodedPath ?? self.path
     }
 
     var pathHasTrailingSlash: Bool {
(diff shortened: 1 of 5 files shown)


Lukasa:

Lukasa:
Flaky test: #175.

Lukasa:

Lukasa:

artemredkin:

iainsmith:
@Lukasa, I just rechecked this locally (disabling the flakey test) and I noticed that I'm hitting a few legitimate failures on `testRequestURITrailingSlash` that I need to fix.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -464,33 +464,11 @@ extension URL {
...
-            }
-
-            return url[pathEndIndex] == "/"
-        }
+        return URLComponents(url: self, resolvingAgainstBaseURL: false)?.percentEncodedPath ?? self.path
     }
 
     var uri: String {
         var uri = self.percentEncodedPath
-        if self.pathHasTrailingSlash, uri != "/" {
```
iainsmith, line 491: So it seems this case is handled correctly through the URLComponenets.percentEncodedPath approach. This behaviour looks well tested in `testRequestURITrailingSlash` which should now be passing (at least on iOS and macOS).

iainsmith:
I pushed another commit that removes some redundant code. Would one of you mind re-triggering CI?

Also happy to squash the commits if that is preferred here.

artemredkin:

PR #202, opened 2020-04-17
Add support for FileRegion in HTTPClient.Body

adam-fowler:
Given you can initialise an `IOData` with a `FileRegion` it seems sensible to provide this functionality to users of AsyncHTTPClient.

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -1733,7 +1733,7 @@ class HTTPClientTests: XCTestCase {
             XCTAssertNoThrow(try httpClient.syncShutdown())
         }
         // create random block of text
-        let uint8Array = Array<Int8>(repeating: 0, count: 16384).map { _ in Int8.random(in: 32...127)}
+        let uint8Array = [Int8](repeating: 0, count: 16384).map { _ in Int8.random(in: 32...127)}
         let asciiString = String(utf8String: uint8Array)!
         let asciiData = Data(asciiString.utf8)
 
@@ -1755,7 +1755,6 @@ class HTTPClientTests: XCTestCase {
             let bytes = response.body.flatMap { $0.getData(at: 0, length: $0.readableBytes) }
             let responseData = try JSONDecoder().decode(RequestInfo.self, from: bytes!)
             XCTAssertEqual(String(data: asciiData[4190..<12745], encoding: .utf8)!, responseData.data)
-            
         } catch {
             XCTFail("\(error)")
         }
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -1724,4 +1724,40 @@ class HTTPClientTests: XCTestCase {
         }
         XCTAssertNoThrow(try promise.futureResult.wait())
     }
+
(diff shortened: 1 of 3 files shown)


Lukasa:

adam-fowler:
Ah I never tested this with https.

artemredkin:

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -1701,7 +1701,7 @@ class HTTPClientTests: XCTestCase {
             XCTAssertNoThrow(try httpClient.syncShutdown())
         }
         // create random block of text
-        let uint8Array = Array<Int8>(repeating: 0, count: 16384).map { _ in Int8.random(in: 32...127)}
+        let uint8Array = [Int8](repeating: 0, count: 16384).map { _ in Int8.random(in: 32...127)}
         let asciiString = String(utf8String: uint8Array)!
```
weissi, line 1705: isn't something like this nicer to read?

```
String(decoding: Array(repeating: UInt8(0), count: 16).map { _ in (UInt8(ascii: "A")...UInt8(ascii: "z")).randomElement()! }, as: UTF8.self)
```

`String(decoding:as:)` also doesn't return optional and is the fastest initialiser because it's in the stdlib and not Foundation (ie. doesn't go through NSString).

adam-fowler:
@Lukasa another issue to deal with before this. FileRegion doesn't work with NIOTS.

Lukasa:
Yup, sounds right. I'm pretty meh on the power of sendfile at the best of times due to the increasing move to bring networking (and especially the crypto associated with networking) into userspace. In userspace, _someone_ has to read the file, and it forces a lot of complexity into the pipeline to require that the channels or TLS handlers do it.

In particular, the "easy" way to do this for NIOTS (and indeed for NIOSSL) is to just memory-map the file. But this is a bad idea. In particular, it still does blocking I/O but instead of doing it in a syscall it does it on memory access, which is really not ideal, and does a stunning job of blocking the event loop.

So instead what we have to do is to write an abstraction over a write queue that knows to hold up all byte buffer writes while we're doing chunked reads from a file. It really brings in a lot of overhead. It also forces the handlers/channels to make decisions about how many threads to use for this blocking I/O, and those decisions are almost always going to be suboptimal because the user often won't even realise they had to be made.

That said, `IOData` exists today and it is seeing use, so we probably don't have a choice.

adam-fowler:
@Lukasa yeah it seems this might be something that is better supported at the application level. You are never going to to able to know the most optimal way to support this within different usages of AsyncHTTPClient/NIO. As you say you'll have to kick off new resources to get it to work, which might not be the most optimal way to do this. Maybe we should close this PR. By providing it at the AHC level you are implying to the user that `FileRegion` is in effect "free" when in actual fact it comes with a number of hidden costs.

Lukasa:
I'm open to closing this for now. I think there's a useful feature at a higher level that allows users to say "I want to upload this FileRegion" and where AHC arranges to read in the file in chunks. That would probably be a nice interface for those who want to avoid managing this complexity themselves.

adam-fowler:
It would be a useful feature. I've implemented something similar for S3 multipart upload, but I ended up with the following as a parameter of the function.
```swift
    public enum ThreadPoolProvider {
        case createNew
        case shared(NIOThreadPool)
    }
```

PR #207, opened 2020-05-12
make cookie parsing case insensitive

### LINKED ISSUE #205: Case-sensitive Set-Cookie parsing
The current parsing is case-sensitive:

https://github.com/swift-server/async-http-client/blob/master/Sources/AsyncHTTPClient/HTTPClient%2BHTTPCookie.swift#L91

Need to make it case-insensitive as per https://tools.ietf.org/html/rfc6265#section-5.2.2

artemredkin:
Make cookie parsing case-insensitive, closes #205

Motivation:
As per https://tools.ietf.org/html/rfc6265#section-5.2.2, cookie names should be case-insensitive, and currently they are not.

Modifications:
1. always split cookie component into name and optional value, lowercase the name and then match by this name using pre-defined lowercased constant
2. modify one of the tests to send differently-cased cookie components

Result:
Cookie parsing is case-insensitive.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
@@ -69,37 +69,24 @@ extension HTTPClient {
             self.secure = false
 
             for component in components[1...] {
-                if component.starts(with: "Path"), let value = parseComponentValue(component) {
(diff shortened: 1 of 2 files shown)


artemredkin:

vlm:
This code introduces a denial of service.

https://github.com/swift-server/async-http-client/issues/209

PR #208, opened 2020-05-12
allow empty value cookies

### LINKED ISSUE #206: Empty cookie values are not parsed
The empty cookie value makes cookie unparsable, which precludes deleting values:

https://github.com/swift-server/async-http-client/blob/master/Sources/AsyncHTTPClient/HTTPClient%2BHTTPCookie.swift#L53

E.g. the `Set-Cookie: cookieValue=; Path=/` won't be recognized as a valid `HTTPClient.Cookie` structure.

This violates https://tools.ietf.org/html/rfc6265#section-4.1.1 which requires empty cookies to be parsable.

One fix would be just to add a `omittingEmptySubsequences: true` to the split() call.

artemredkin:
Allow empty-value cookies, closes #206 

Motivation:
RFC allows empty-value cookies, so we should handle them as well

Modifications:
1. Fix split in cookie parsing
2. Add a test

Result:
We can now allow empty value cookies

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
@@ -50,7 +50,7 @@ extension HTTPClient {
                 return nil
             }
 
-            let nameAndValue = components[0].split(separator: "=", maxSplits: 1).map {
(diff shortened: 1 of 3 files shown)


artemredkin:

artemredkin:

artemredkin:

PR #210, opened 2020-05-13
fix malformed cookie parsing

### LINKED ISSUE #209: [Bug] Untimely client termination on malformed Set-Cookie values
The `Set-Cookie:` header might carry any combination of characters. In case a consecutive sequence of semicolons ";;" is encountered, the client will crash because of this code:

https://github.com/swift-server/async-http-client/pull/207/files#diff-9f1723c83114eacc62a23489d124343eR117

Client termination when encountering such headers is not desired and might result in a denial of service.

artemredkin:
Fixes malformed cookie parsing, closes #209.

Motivation:
We do not check that cookie component name could be empty, assuming that split array is not empty, that could lead to crash.

Modifications:
1. return empty pairs if split array is empty
2. add a test

Result:
Cookie parsing is more robust.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
@@ -61,7 +61,7 @@ extension HTTPClient {
             self.name = nameAndValue[0]
             self.value = nameAndValue[1]
 
-            guard !name.isEmpty else {
+            guard !self.name.isEmpty else {
                 return nil
             }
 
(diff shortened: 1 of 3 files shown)


artemredkin:

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -61,6 +61,10 @@ extension HTTPClient {
             self.name = nameAndValue[0]
             self.value = nameAndValue[1]
 
+            guard !self.name.isEmpty else {
+                return nil
+            }
```
Lukasa, line 66: This new guard does not seem to be met by an associated test: mind adding one?

artemredkin: we do have them:
```
XCTAssertNil(HTTPClient.Cookie(header: "=;", defaultDomain: "exampe.org"))
```
should hit it

Lukasa: :+1:

PR #214, opened 2020-05-18
don't flush on every body part

### LINKED ISSUE #203: AsyncHTTPClient calls flush after writing every part of the HTTP message
Repeatedly calling `flush` doesn't seem very optimal.

- `HTTPClient.execute` calls `writeAndFlush` and because the body write isn't queued until the head has been written, the flush occurs immediately after the head has been written.
- `TaskHandler.writeBody` calls `writeAndFlush`.

Ideally a simple HTTP request with one `ByteBuffer` shouldn't need to call flush until after the end part.

artemredkin:
Right now we call `flush` on every body part write, this can be suboptimal, closes #203 

Motivation:
Library users can write data in small chunk, in this case we are not buffering enough data, better solution would be to buffer at the NIO level in this case.

Modifications:
Body part write is not not flushing, just `write`

Result:
Body part writes are not flushing anymore

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -767,9 +767,10 @@ extension TaskHandler: ChannelDuplexHandler {
 
         return body.stream(HTTPClient.Body.StreamWriter { part in
             context.eventLoop.assertInEventLoop()
-            return context.writeAndFlush(self.wrapOutboundOut(.body(part))).map {
+            context.write(self.wrapOutboundOut(.body(part))).whenSuccess {
                 self.callOutToDelegateFireAndForget(value: part, self.delegate.didSendRequestPart)
             }
+            return context.eventLoop.makeSucceededFuture(())
         })
     }
 


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -767,9 +767,10 @@ extension TaskHandler: ChannelDuplexHandler {
 
         return body.stream(HTTPClient.Body.StreamWriter { part in
             context.eventLoop.assertInEventLoop()
-            return context.writeAndFlush(self.wrapOutboundOut(.body(part))).map {
+            context.write(self.wrapOutboundOut(.body(part))).whenSuccess {
```
weissi, line 770: hmm, this would now _never_ flush any body which means we'll load the whole body into memory before sending.

I think the right thing to do is:

- if you know that whole body straight away (without streaming), then we should do
  `write(.head)`, `.write(.body)`, `.write(end)`, `flush`
- if we actually want to stream the body, then
    `writeAndFlush(.head)`, `writeAndFlush(.body)`, `writeAndFlush(.end)`

weissi: We also need a test actually that shows that this PR doesn't work. We should have a test which checks that body chunks do arrive at the other end before the next bit is sent out.

weissi: wrote #219  to test this. This PR should go in after #219 to prove that it works.

ktoso:

PR #215, opened 2020-05-18
draft for streaming el fixes

### LINKED ISSUE #198: programming model: EventLoopPreference.delegate(on:) not respected for .stream
I'm not 100% sure if this counts as a bug but I think it should because it makes the programming model really hard.

If I configure

```
                let desiredEL = context.eventLoop // or some other EL I like
                let req = try HTTPClient.Request(url: url, method: req.method, headers: req.headers, body: .stream { _ in desiredEL.preconditionInEventLoop() })
                let task = self.httpClient.execute(request: req,
                                                   delegate: self,
                                                   eventLoop: .delegate(on: desiredEL))
```

this would fail as the closure passed to `.stream` isn't executed on `desiredEL`.

### LINKED ISSUE #200: StreamWriter expects to be called on some internal EventLoop
The `HTTPClient.Body.StreamWriter` that you get when using the default `.stream(length: contentLength, { thisStreamWriterHere in })` expects to be called on the internal `EventLoop` that belongs to the underlying `Channel`. The user however doesn't get access to that `Channel` at all so it's impossible to use this API :(.

This needs a fix really soon.

artemredkin:
Fixes programming model for upload streaming

Motivation:
Right now EL on which we execute streaming call back is not explicitly defined. Also, we do not check on which EL part write is executed, meaning we can fail if `.write` is not called on correct EL.

Modifications:
1. Call to `body.stream` is now executed on tasks EL
2. Calls to `writeAndFlush` are always called on channel EL
3. Added tests for those to cases
4. Removed a hack in test introduced in #219 
5. Fixed #219 shutdown to not call to shutdown EL

Result:
1. closes #198
2. closes #200

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -766,7 +766,7 @@ extension TaskHandler: ChannelDuplexHandler {
         }
 
         func doIt() -> EventLoopFuture<Void> {
-            body.stream(HTTPClient.Body.StreamWriter { part in
+            return body.stream(HTTPClient.Body.StreamWriter { part in
                 self.task.eventLoop.assertInEventLoop()
                 func doIt() -> EventLoopFuture<Void> {
                     return context.writeAndFlush(self.wrapOutboundOut(.body(part))).map {
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -765,12 +765,32 @@ extension TaskHandler: ChannelDuplexHandler {
             return context.eventLoop.makeSucceededFuture(())
         }
 
-        return body.stream(HTTPClient.Body.StreamWriter { part in
-            context.eventLoop.assertInEventLoop()
-            return context.writeAndFlush(self.wrapOutboundOut(.body(part))).map {
-                self.callOutToDelegateFireAndForget(value: part, self.delegate.didSendRequestPart)
+        func doIt() -> EventLoopFuture<Void> {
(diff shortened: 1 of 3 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -765,12 +765,32 @@ extension TaskHandler: ChannelDuplexHandler {
             return context.eventLoop.makeSucceededFuture(())
         }
 
-        return body.stream(HTTPClient.Body.StreamWriter { part in
-            context.eventLoop.assertInEventLoop()
-            return context.writeAndFlush(self.wrapOutboundOut(.body(part))).map {
-                self.callOutToDelegateFireAndForget(value: part, self.delegate.didSendRequestPart)
+        func doIt() -> EventLoopFuture<Void> {
+            return body.stream(HTTPClient.Body.StreamWriter { part in
+                let promise = self.task.eventLoop.makePromise(of: Void.self)
```
weissi, line 770: this whole block should just be

```swift
return channel.writeAndFlush(HTTPClientRequestPart.body(part)).map {
    self.callOutToDelegateFireAndForget(value: part, self.delegate.didSendRequetsPart)
}
```

artemredkin: we only have access to `context` here, so in order to get to `channel` we have to call `context.channel` that is precondition on an being on channel EL, am I missing something?

artemredkin: actually, even if I get the channel, I cannot write `.part` to it since first handler in the channel expects `Request`

PR #216, opened 2020-05-19
make public API tests not use @testable again

weissi:
Motivation:

AHC's tests were split in HTTPClientTests (which only use the public
API) and HTTPClientInternalTests (which can use `internal` API by using
`@testable`). At some point, the public API tests had `@testable` added
to the AHC import which breaks this idea.

Modification:

Restore the intent by removing `@testable` and moving the 2 tests that
needed it over.

Result:

Cleaner test suite.


Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@testable import AsyncHTTPClient
+/* NOT @testable */ import AsyncHTTPClient // Tests that need @testable go into HTTPClientInternalTests.swift
```
ktoso, line 15: nice strategy to only have a specific file for those, +1

PR #225, opened 2020-05-21
cpool: don't reuse connection if we sent close

weissi:
Motivation:

Previously, we'd only use the server's connection header to determine if
we should close the connection or not. That's wrong because if we set
`connection: close` ourselves, we must not reuse again.

Modification:

Set `TaskHandler.closing = false` if we send a close header.

Result:

More HTTP correctness.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -604,7 +604,13 @@ internal class TaskHandler<Delegate: HTTPClientResponseDelegate>: RemovableChann
     var state: State = .idle
     var pendingRead = false
     var mayRead = true
-    var closing = false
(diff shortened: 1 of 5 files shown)


weissi:
hit #226 , retrying

weissi:

weissi:
swiftpm bug. Not filing because it's the 5.0 one, if we see this on 5.2 we should file

```
[437/448] Compiling Swift Module 'NIO' (66 sources)
<unknown>:0: error: missing submodule 'CNIOLinux'
/code/.build/checkouts/swift-nio/Sources/NIO/GetaddrinfoResolver.swift:24:8: error: no such module 'CNIOLinux'
import CNIOLinux
       ^
```

weissi:

PR #228, opened 2020-05-27
Better support for UNIX Domain sockets

dimitribouniol:
These changes upgrade the unix domain socket handling to address the following issues:
- Using socket paths currently requires obtuse URL objects with a specially constructed baseURL. These are not easily serializable as strings.
- It is currently not possible to connect to a server over a socket path that uses TLS.

To maintain backward compatibility, this PR introduces two new URL schemes: `http+unix://`, and `https+unix://`. These url schemes, while not universal, are used in a standard way by other services ([some](https://github.com/httpie/httpie-unixsocket) - [examples](https://github.com/msabramo/requests-unixsocket)), and work by %-encoding the socket path as the url hostname. These are easy to construct using standard swift types:
```swift
let socketURL = URL(string: "https+unix://\(path.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)!)/echo-uri")
```
This produces a url that looks something like this: `https+unix://%2Ftmp%2FfilePath.socket/echo-uri`, which is fully supported by `URL`, and can be serialized and deserialized to and from a String easily.

When the `https+unix://` URL scheme is used, TLS is turned on. It is still up to the user to perform certificate validation under this scheme, but at the very least an encrypted connection over a socket is now possible if the server is configured to only provide such a configuration.

I'd also like to think I did a good job keeping things neat and tidy, and inline with the style of the repo, but I'm open to any and all feedback! (This is my first time contributing to any large open source project, so please be kind 😅)

### Where to go from here

Some future improvements that can be made:
- More tests…
- Some documentation on Request outlining the different possibilities.
- It would be nice to have a factory on URL to build a unix socket URL from a socket path and server path.
- We need some constants for "http", "https", "unix", "http+unix", and "https+unix" - it's a bit unwieldy at the moment.
- Perhaps deprecating the `unix:` scheme for the next major version? It's use is a bit unwieldy from a user's point of view (especially as many frameworks like Vapor have client constructors that take in strings as the URLs), and I doubt its currently widely used.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -99,20 +99,23 @@ extension HTTPClient {
     public struct Request {
         /// Represent kind of Request
         enum Kind {
+            enum UnixScheme {
+                case baseURL
+            }
+
             /// Remote host request.
             case host
             /// UNIX Domain Socket HTTP request.
-            case unixSocket
+            case unixSocket(_ scheme: UnixScheme)
 
             private static var hostSchemes = ["http", "https"]
             private static var unixSchemes = ["unix"]
 
             init(forScheme scheme: String) throws {
-                if Kind.host.supports(scheme: scheme) {
-                    self = .host
-                } else if Kind.unixSocket.supports(scheme: scheme) {
-                    self = .unixSocket
-                } else {
+                switch scheme {
+                case "http", "https": self = .host
+                case "unix": self = .unixSocket(.baseURL)
+                default:
                     throw HTTPClientError.unsupportedScheme(scheme)
                 }
             }
@@ -126,6 +129,24 @@ extension HTTPClient {
                     return host
                 case .unixSocket:
                     return ""
+                }
+            }
+
+            func socketPathFromURL(_ url: URL) throws -> String {
+                switch self {
+                case .unixSocket(.baseURL):
+                    return url.baseURL?.path ?? url.path
+                case .host:
+                    return ""
+                }
+            }
+
+            func uriFromURL(_ url: URL) -> String {
+                switch self {
+                case .host:
+                    return url.uri
+                case .unixSocket(.baseURL):
+                    return url.baseURL != nil ? url.uri : "/"
                 }
             }
 
@@ -147,6 +168,10 @@ extension HTTPClient {
         public let scheme: String
         /// Remote host, resolved from `URL`.
         public let host: String
+        /// Socket path, resolved from `URL`.
+        let socketPath: String
+        /// URI composed of the path and query, resolved from `URL`.
+        let uri: String
         /// Request custom HTTP Headers, defaults to no headers.
         public var headers: HTTPHeaders
         /// Request body, defaults to no body.
@@ -199,6 +224,8 @@ extension HTTPClient {
 
             self.kind = try Kind(forScheme: scheme)
             self.host = try self.kind.hostFromURL(url)
+            self.socketPath = try self.kind.socketPathFromURL(url)
+            self.uri = self.kind.uriFromURL(url)
 
             self.redirectState = nil
             self.url = url
@@ -712,19 +739,9 @@ extension TaskHandler: ChannelDuplexHandler {
         self.state = .idle
         let request = self.unwrapOutboundIn(data)
 
-        let uri: String
-        switch (self.kind, request.url.baseURL) {
-        case (.host, _):
-            uri = request.url.uri
-        case (.unixSocket, .none):
-            uri = "/" // we don't have a real path, the path we have is the path of the UNIX Domain Socket.
-        case (.unixSocket, .some(_)):
-            uri = request.url.uri
-        }
-
         var head = HTTPRequestHead(version: HTTPVersion(major: 1, minor: 1),
                                    method: request.method,
-                                   uri: uri)
+                                   uri: request.uri)
         var headers = request.headers
 
         if !request.headers.contains(name: "Host") {
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -101,6 +101,8 @@ extension HTTPClient {
         enum Kind {
             enum UnixScheme {
                 case baseURL
+                case http_unix
+                case https_unix
             }
 
             /// Remote host request.
@@ -109,12 +111,14 @@ extension HTTPClient {
             case unixSocket(_ scheme: UnixScheme)
 
             private static var hostSchemes = ["http", "https"]
-            private static var unixSchemes = ["unix"]
+            private static var unixSchemes = ["unix", "http+unix", "https+unix"]
 
             init(forScheme scheme: String) throws {
                 switch scheme {
                 case "http", "https": self = .host
                 case "unix": self = .unixSocket(.baseURL)
+                case "http+unix": self = .unixSocket(.http_unix)
+                case "https+unix": self = .unixSocket(.https_unix)
                 default:
                     throw HTTPClientError.unsupportedScheme(scheme)
                 }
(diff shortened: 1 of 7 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -147,6 +189,10 @@ extension HTTPClient {
         public let scheme: String
         /// Remote host, resolved from `URL`.
         public let host: String
+        /// Socket path, resolved from `URL`.
+        public let socketPath: String
+        /// URI, resolved from `URL`.
+        public let uri: String
```
Lukasa, line 195: Is it necessary for either of these to be `public`? In particular, `uri` is _extremely_ not well distinguished from `url`, and if we're keeping both in the API we really need to explain how they differ.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -129,11 +140,42 @@ extension HTTPClient {
...
+                }
+            }
+
             func supports(scheme: String) -> Bool {
                 switch self {
                 case .host:
                     return Kind.hostSchemes.contains(scheme)
-                case .unixSocket:
+                case .unixSocket(scheme: _):
                     return Kind.unixSchemes.contains(scheme)
```
Lukasa, line 179: This `supports` spelling is getting unwieldy here, not least because we're no longer using half of the `switch` statement. Should this just be a constructor of `Kind` from `String`?

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -234,7 +236,7 @@ internal final class HTTPBin {
 
         self.serverChannel = try! ServerBootstrap(group: self.group)
             .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
-            .childChannelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: 1)
+            .childChannelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: usingTCP ? 1 : 0)
```
Lukasa, line 239: Rather than do this, we should probably just only set this conditionally on whether we're using TCP. The kernel is within its rights to refuse _any_ manipulation of `TCP_NODELAY` on a unix socket, even turning it off.

weissi: We can delete this line without replacement. NIO in the required version _defaults_ to `TCP_NODELAY`.

weissi:
Very cool!

dimitribouniol:
Thank you @Lukasa and @weissi for your feedback! Should I amend my previous commits and replace the ones here, or would you prefer I add additional commits that address the remaining changes?

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
dimitribouniol: > We can delete this line without replacement. NIO in the required version _defaults_ to `TCP_NODELAY`.
I'll go ahead and remove this line, undo the rest of the changes here, and mention this in the commit.

Lukasa:
It's entirely up to you: we'll squash when we merge, so we'll end up with only one commit anyway.

Sources/AsyncHTTPClient/HTTPHandler.swift
dimitribouniol: Not at all — continuing on that thought, the host will be empty for socket-based Requests, even though that's where its encoded… Should we merge the implementations to make the socket path available as the `host`?

Sources/AsyncHTTPClient/HTTPHandler.swift
dimitribouniol: I'm not too sure — we are already constructing a kind from a String in https://github.com/swift-server/async-http-client/pull/228/files#diff-67e2f60fc709da32af89dfbfc0c774aaR116-R129, so that method can certainly be fixed up to no longer rely on `supports()`.

My bigger question is that `supports()` seems like it is being used by some redirect logic here: https://github.com/swift-server/async-http-client/blob/070c1e5f374c76d5f55c2b4b60cbb7aa04471301/Sources/AsyncHTTPClient/HTTPHandler.swift#L962

image
It seems like the older version was just making sure that the new scheme matched http/https - while the one currently in the repo will only accept http redirects from http requests, and socket redirects from socket requests. @krzyzanowskim Do you remember more about why you set it up this way?

I propose that if the request is an http request, we keep it as such, but if it is a socket request, we allow it to redirect to any supported scheme, but that is based on my limited understanding of what's going on here 😅

Lukasa: > I propose that if the request is an http request, we keep it as such, but if it is a socket request, we allow it to redirect to any supported scheme, but that is based on my limited understanding of what's going on here 😅

Yeah, that seems a reasonable design choice.

dimitribouniol: Would you support refactoring this method to something like this, since it is now only used for redirects?
image

Call site (which was checking against the old scheme for some reason that escapes me):
image


This would also simplify the initializer tremendously:
image

dimitribouniol:
I addressed most of the changes - all that's left are the following related but out of scope issues:

### Should we combine `host` and `socketPath`?

This should probably be a different discussion/PR due to where `Request.host` is being used, but from what I can determine, this would probably be the right thing to do in the long run.

From what I could tell:
- The socket path would end up in the Host header for the request (currently it is `""`)
- TLS would use check the socket path on certs as the common name/hostname (currently it is checking for `""` as opposed to `nil`, which it's doing for IPs)
    - Should IPs also be part of the cert validation? Doesn't make sense for them to be omitted honestly...
- Responses would propagate the socket path as the host as well (currently, it's also `""`)
- We would need better test coverage for these scenarios.

I'd be happy to submit the above as a separate PR.

### It's likely redirects for sockets never behaved as expected

The redirect logic checks the current request rather than the new URL to see if it can redirect to it - this likely never worked as designed due to the fact that it would always pass, since we have a request that is already valid, and we are checking It for validity...

Again, I'd be happy to refactor this and submit the fix as a separate PR.

Lukasa:

Lukasa:
As for the follow-on issues you identified, I'm more than happy for you to address them in follow-up PRs.

Lukasa:
It seems there are some minor formatting issues that need to be cleaned up.

weissi:
@dimitribouniol apologies, there seem to be formatting issues. If you just run `scripts/sanity.sh` on your machine, it should automatically invoke [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) and fix those.

Sources/AsyncHTTPClient/HTTPHandler.swift
krzyzanowskim: > @krzyzanowskim Do you remember more about why you set it up this way?

Sorry, I don't remember details at the moment. I'm sorry not being helpful here.

dimitribouniol:
I've just addressed the formatting issues — I'm creating separate tracking issues for the two points I brought up above for further discussion.

Sources/AsyncHTTPClient/HTTPHandler.swift
dimitribouniol: @krzyzanowskim Not a problem! I think the original code may have also had the same issue, so ultimately we just need more tests that capture the underlying problem 😅

weissi:

dimitribouniol:
Seems like on linux, the client will throw a `NIOSSLExtraError.invalidSNIHostname` if the SNI is equal to `""` (length must be between 1 and 255), so I replace an empty host with nil before proceeding (just like is done for IP addresses).

@weissi Could you re-trigger the tests please? (unless I'm able to do it using the same command you used?)

dimitribouniol:

dimitribouniol:
@weissi Please hold off on re-testing - I got Docker running locally, and it's still failing 😅

dimitribouniol:
@weissi All tests now pass on my local machine! Please re-test 🙂

weissi:

PR #235, opened 2020-06-04
Convenience methods for socket paths

dimitribouniol:
With the intent of making socket paths a bit easier to use, I added some convenience methods and initializers for creating Requests and URLs directly by specifying the socket path, and the resource path. I also added some additional tests for socket path based use that I missed the first time around.


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -224,37 +256,142 @@ public class HTTPClient {
...
+    ///     - deadline: The time when the request must have been completed by.
+    public func delete(secureSocketPath: String, url: String, deadline: NIODeadline? = nil) -> EventLoopFuture<Response> {
+        return self.execute(secureSocketPath: secureSocketPath, url: url, method: .DELETE, deadline: deadline)
+    }
+
+    /// Execute arbitrary HTTP request using specified URL.
+    ///
+    /// - parameters:
+    ///     - url: Request url.
+    ///     - method: Request rethod.
```
Lukasa, line 349:
```suggestion
    ///     - method: Request method.
```

dimitribouniol:
I’d also be happy rolling back the variations on the get/push/patch/put/delete methods, but keeping the execute variations, since they are a bit more general.

Alternatively, if there were a way of specifying a default baseURL that I’m unaware of for the whole client, that would alleviate the need for much of this, but it would probably require more internal re-wiring.

Sources/AsyncHTTPClient/HTTPClient.swift
dimitribouniol: Nw that I think about this a bit more, I'm wondering if it should be `func execute(method: HTTPMethod, url: String, body: Body? = nil, deadline: NIODeadline? = nil)` (swapping method and url)

dimitribouniol:
Rebased these commits on top of the logging changes.

@Lukasa I removed the get/push/patch/put/delete methods in https://github.com/swift-server/async-http-client/pull/235/commits/f314a6f2a84aa0de3e0a88ffe117c5eb9210a480 that I added in favor of a pair of execute() methods:
```swift
    public func execute(socketPath: String, url: String, method: HTTPMethod, body: Body? = nil, deadline: NIODeadline? = nil, logger: Logger? = nil) -> EventLoopFuture<Response>
    public func execute(secureSocketPath: String, url: String, method: HTTPMethod, body: Body? = nil, deadline: NIODeadline? = nil, logger: Logger? = nil) -> EventLoopFuture<Response>
```

In https://github.com/swift-server/async-http-client/pull/235/commits/cf4ff26a287d523d39d9f96c329de186fb9d5f1c, I actually rename the trio of URL execute methods to:
```swift
    public func execute(_ method: HTTPMethod = .GET, url: String, body: Body? = nil, deadline: NIODeadline? = nil, logger: Logger? = nil) -> EventLoopFuture<Response>
    public func execute(_ method: HTTPMethod = .GET, socketPath: String, urlPath: String, body: Body? = nil, deadline: NIODeadline? = nil, logger: Logger? = nil) -> EventLoopFuture<Response>
    public func execute(_ method: HTTPMethod = .GET, secureSocketPath: String, urlPath: String, body: Body? = nil, deadline: NIODeadline? = nil, logger: Logger? = nil) -> EventLoopFuture<Response>
```

This allows them to be a bit more expressive when used directly:
```swift
httpClient.execute(socketPath: "/tmp/myServer.socket", urlPath: "/path/to/resource").whenComplete (...) // Implicitly GET
httpClient.execute(.POST, secureSocketPath: "/tmp/myServer.socket", urlPath: "/path/to/resource", body: .string("hello")).whenComplete (...)
```

That said, we can drop the last commit if you refer the previous naming and argument ordering.

@weissi Since I had changes in the same place, I noticed what I think are some leftover issues from logging that I fixed in https://github.com/swift-server/async-http-client/pull/235/commits/b65f9806bddcb05e7d51164ab6f995b671037e17, but I may be wrong.

Also, please let me know if I handled adding a logger to the other new methods correctly in the later commits, as I opted to use an optional just like with deadline. I wasn't sure if that's OK though, since you had it as a required parameter, but in a separate method for the other examples (which I understand was to maintain backwards compatibility?). I think it should be fine since these are all new methods to begin with?

Lukasa:

weissi:

dimitribouniol:
@weiss I think I forgot to push the version after running sanity.sh - please test it again. (also rebased while I was at it)

weiss:
I guess you meant to mention @weissi :smile:

dimitribouniol:
Oh no 😅

dimitribouniol:
@weissi So I rebased off the latest changes, but `testUncleanCloseThrows()` now fails, even on master: 
image

So I guess, don't bother kicking off the tests 😅

Should I file an issue for the crashing test?

artemredkin:
I'm looking into that crash right now, it seem to happen only on the mac, linux tests are working as expected. Could you confirm that the error being processed in `HTTP1ConnectionProvider.connect` (13-th frame in the stack trace) is `connection refused`?

artemredkin:

dimitribouniol:
Seems like `isOnClosedChannel`, unless you were asking something else. 
image

I originally thought it may be the versions of my dependencies, but I haven't tested with the latest of each: 
image

artemredkin:
Interesting, yes, it is exactly was I was asking about, thank you! I'll look into that

dimitribouniol:
I can confirm the same crash occurs with a crash checkout of each dependency, so not version specific at the very least 🙂: 
image

artemredkin:
filed: https://github.com/swift-server/async-http-client/issues/248

dimitribouniol:
@artemredkin Rebased onto of latest changes. Although all tests pass on linux (tested via docker), there are two new concerns:

`testTLSFailError()` now fails on macOS: (maybe related to changes from #237?)
image

`testUncleanShutdownActuallyShutsDown()` crashes now on macOS:
image

Should I create issues for these?

artemredkin:

weissi:

weissi:
@artemredkin this just failed with

```
Test Case 'HTTPClientTests.testContentLengthTooLongFails' started at 2020-06-16 19:10:55.784
/code/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:2323: error: HTTPClientTests.testContentLengthTooLongFails : XCTAssertThrowsError failed: did not throw error - 
/code/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:2338: error: HTTPClientTests.testContentLengthTooLongFails : failed - Body missing: Response(host: "localhost", status: NIOHTTP1.HTTPResponseStatus.badRequest, version: HTTP/1.1, headers: [("Connection", "close"), ("Content-Length", "0")], body: nil)
Test Case 'HTTPClientTests.testContentLengthTooLongFails' failed (0.046 seconds)
```

dimitribouniol:
Rebased to keep up to date with master

dimitribouniol:
@weissi @artemredkin I don't know why its passing now - for what it's worth, #259 is failing with the same exact error on the same test, so it may only happen sometimes.

weissi:
@dimitribouniol thanks, sounds like a flaky test, filed #265

dimitribouniol:
Thank you!

PR #237, opened 2020-06-12
The host header should also include the port

adam-fowler:
See https://tools.ietf.org/html/rfc7230#section-5.4

- If port is not 80 or 443 then add to host header.
- Fixed up tests

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -253,6 +253,11 @@ extension HTTPClient {
             return self.scheme == "https" || self.scheme == "https+unix"
         }
 
+        /// Specified port.
+        public var specifiedPort: Int? {
+            return self.url.port
+        }
+
         /// Resolved port.
         public var port: Int {
             return self.url.port ?? (self.useTLS ? 443 : 80)
@@ -772,7 +777,7 @@ extension TaskHandler: ChannelDuplexHandler {
         var headers = request.headers
 
         if !request.headers.contains(name: "Host") {
-            headers.add(name: "Host", value: request.host)
+            headers.add(name: "Host", value: "\(request.host)\(request.specifiedPort.map { ":\($0)" } ?? "")")
         }
 
         do {
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -774,7 +774,7 @@ extension TaskHandler: ChannelDuplexHandler {
         if !request.headers.contains(name: "host") {
             let port = request.port
             var host = request.host
-            if (port != 80 || request.scheme != "http") && (port != 443 || request.scheme != "https") {
+            if !(port == 80 && request.scheme == "http"), !(port == 443 && request.scheme == "https") {
                 host += ":\(port)"
             }
             headers.add(name: "host", value: host)
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -254,7 +254,7 @@ extension HTTPClient {
         }
 
         /// Specified port.
-        public var specifiedPort: Int? {
+        var specifiedPort: Int? {
             return self.url.port
         }
 
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -253,11 +253,6 @@ extension HTTPClient {
             return self.scheme == "https" || self.scheme == "https+unix"
         }
 
-        /// Specified port.
-        var specifiedPort: Int? {
-            return self.url.port
-        }
-
         /// Resolved port.
         public var port: Int {
             return self.url.port ?? (self.useTLS ? 443 : 80)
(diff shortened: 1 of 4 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -253,6 +253,11 @@ extension HTTPClient {
             return self.scheme == "https" || self.scheme == "https+unix"
         }
 
+        /// Specified port.
+        public var specifiedPort: Int? {
```
artemredkin, line 257: why not just `port`?

artemredkin, line 257: Do we need to expose this in public API?

adam-fowler: ah we don't need that I guess

adam-fowler:
@Lukasa as an alternative to what is there I could remove the specifiedPort and just output a port when it isn’t 80 or 443.

Lukasa:
I think that's the right choice.

adam-fowler:
Tests aren't working just now, seems like there were some changes on the main branch that broke a few tests.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -772,7 +772,9 @@ extension TaskHandler: ChannelDuplexHandler {
         var headers = request.headers
 
         if !request.headers.contains(name: "Host") {
-            headers.add(name: "Host", value: request.host)
+            let port = request.port
+            let host = (port != 80 && port != 443) ? "\(request.host):\(port)" : request.host
```
weissi, line 776: shouldn't we only exclude `80` if we're speaking (plain text) HTTP and only exclude `443` when we're speaking HTTPS?

Say I wanted to do `http://localhost:443` (note the missing `s` in `http`), then I should get `localhost:443` or am I missing something? CC @Lukasa ?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -772,7 +772,9 @@ extension TaskHandler: ChannelDuplexHandler {
...
+            headers.add(name: "Host", value: host)
```
weissi, line 777: can we do lower-cased header names? That simplifies HTTP/2 because the 2to1 handler should then be able to just pass it through (yes, `Host` becomes `:authority` but I think we should do lowercase only).

Sources/AsyncHTTPClient/HTTPHandler.swift
adam-fowler: From RFC7230 "If the port is equal to the default port for a scheme, the normal form is to omit the port subcomponent"
I guess 443 is not the normal port, for `http` so it should not be submitted. Do we want to support the other way. ie `https://locahost:80/`?

weissi: I think so. I think the only instances where we can omit the port numbers are

- `http://...:80`
- `https://...:443`

adam-fowler: done

Sources/AsyncHTTPClient/HTTPHandler.swift
adam-fowler: done

Sources/AsyncHTTPClient/HTTPHandler.swift
adam-fowler: swift format is wanting me to output it as 
```
if port != 80 || request.scheme != "http", port != 443 || request.scheme != "https" {
```
instead of 
```
if (port != 80 || request.scheme != "http") && (port != 443 || request.scheme != "https") {
```
Is that what you are expecting?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -771,8 +771,13 @@ extension TaskHandler: ChannelDuplexHandler {
                                    uri: request.uri)
         var headers = request.headers
 
-        if !request.headers.contains(name: "Host") {
-            headers.add(name: "Host", value: request.host)
+        if !request.headers.contains(name: "host") {
+            let port = request.port
+            var host = request.host
+            if (port != 80 || request.scheme != "http") && (port != 443 || request.scheme != "https") {
```
weissi, line 777: random nit: I personally would find this easier to read after de-morganing this once to

    if ! ((port == 80 && request.scheme == "http") || (port == 443 || request.scheme == "https")) {

but I guess that's personal preference. So please feel totally free to ignore.

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: I really don't like swift-format messing with those things but I guess the argument is consistency ...

weissi: @adam-fowler I think we're discussing on the wrong line here :)

weissi: I meant applying [De Morgan](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) to the condition so switching from

       (port != 80 || request.scheme != "http") && (port != 443 || request.scheme != "https")

to

    ! ((port == 80 && request.scheme == "http") || (port == 443 && request.scheme == "https"))

but it's personal preference

Sources/AsyncHTTPClient/HTTPHandler.swift
adam-fowler: I’m happy to change

weissi: @adam-fowler pick what _you_ find easier to read :)

adam-fowler: I actually prefer the middle ground of 
```
if !(port == 80 && request.scheme == "http") && !(port == 443 && request.scheme == "https") 
```

adam-fowler: although swift format still changes the && into a ,

weissi: :( silly swift format

PR #241, opened 2020-06-12
add a test to make sure that future is bound to delegate el

### LINKED ISSUE #199: programming model:`task.futureResult` not bound to desired EL if `.delegate(on:)` is specified
Again, I'm not 100% sure if this counts as a bug but I think it should. If I set my `EventLoopPreference` to `.delegate(on: desiredEL)`, then I probably want everything to happen on `desiredEL`.

I would argue that all the user-visible things that are important for the programming model should then happen on `desiredEL`. Mainly:

- all delegate methods (true today)
- `task.futureResult` (not true today)
- the closure that gets me the `StreamWriter` (not true today, #198)

artemredkin:
Adds a test to make sure that `Task` future is bound to delegate EL

Motivation:
Tasks futures have to be bound to delegate EL.

Modifications:
Adds a test case

Result:
Closes #199

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -2005,22 +2005,4 @@ class HTTPClientTests: XCTestCase {
 
         self.defaultClient = nil // so it doesn't get shut down again.
     }
-
-    func testTaskPromiseBoundToEL() throws {
-        let elg = getDefaultEventLoopGroup(numberOfThreads: 2)
-        let el = elg.next()
-
-        let client = HTTPClient(eventLoopGroupProvider: .shared(elg))
-
-        defer {
-            XCTAssertNoThrow(try client.syncShutdown())
-            XCTAssertNoThrow(try elg.syncShutdownGracefully())
-        }
-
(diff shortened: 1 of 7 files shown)


Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2005,4 +2005,22 @@ class HTTPClientTests: XCTestCase {
...
+        let client = HTTPClient(eventLoopGroupProvider: .shared(elg))
+
+        defer {
+            XCTAssertNoThrow(try client.syncShutdown())
+            XCTAssertNoThrow(try elg.syncShutdownGracefully())
+        }
+
+        let request = try HTTPClient.Request(url: self.defaultHTTPBinURLPrefix + "/get")
+        let delegate = ResponseAccumulator(request: request)
+        let task = client.execute(request: request, delegate: delegate, eventLoop: .delegate(on: el))
```
weissi, line 2022: shouldn't we use `.testOnly(delegateOn: el, channelOn: otherEL)` here to _force_ a foreign channel?

artemredkin: ah, you are right, done, thanks!

PR #245, opened 2020-06-13
notify delegate about connect errors

### LINKED ISSUE #242: HTTPClientResponseDelegate is not notified of some errors (connection related?).
Tested against commit 8add6b8f0b8f045a3e43616b0f4ffe8b234e3137 on Ubuntu 16.04 with Swift version 5.2.4 (swift-5.2.4-RELEASE)

I've been debugging an issue where an HTTPClientResponseDelegate does not get notified of some errors via the didReceiveError(task:...) method.

My expectation was that HTTPClientResponseDelegate would see any error throughout the entire lifecycle of the request, not just some subset of them.

I've pasted a test case below that demonstrates the issue with remoteConnectionClosed errors but I've seen it with connectTimeout too.

I can appreciate that the error was returned in Task.futureResult but it took quite a bit of effort to work that out (no matter how obvious it is in retrospect). However if the current behaviour is intended then I would still push that the documentation for didReceiveError is updated to emphasize it only sees some subset of errors and not "any network-related error" as it states now.

Thank you!

```
    func testDelegateConnectionFailError() {
        class TestDelegate: HTTPClientResponseDelegate {
            typealias Response = Void

            var error : Error?

            func didFinishRequest(task: HTTPClient.Task<Response>) throws {}

            func didReceiveError(task: HTTPClient.Task<Response>, _ error: Error) -> EventLoopFuture<Void> {
                self.error = error

                return task.eventLoop.makeSucceededFuture(())
            }
        }

        let localHTTPBin = HTTPBin(refusesConnections: true)
        let localClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup))
        defer {
            XCTAssertNoThrow(try localClient.syncShutdown())
            XCTAssertNoThrow(try localHTTPBin.shutdown())
        }

        let delegate = TestDelegate()

        do {
            let request = try HTTPClient.Request(url: "http://localhost:\(localHTTPBin.port)/get")
            try localClient.execute(request: request, delegate: delegate).wait()
            XCTFail("Shouldn't succeed")
        } catch {
            print("error: \(error)")
            print("delegate.error: \(String(describing: delegate.error))")
            XCTAssert(delegate.error != nil)
        }
    }

error: HTTPClientError.remoteConnectionClosed
delegate.error: nil
/data/async-http-client/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:1569: error: HTTPClientTests.foo : XCTAssertTrue failed -
```

artemredkin:
Notify delegates about connect errors

Motivation:
When channel is established, all errors from that channel are piped into `didReceiveError` method of the `HTTPClientReponseDelegate`, but connection errors are only delivered to task's promise. That behaviour might be confusing, since there could be cases where task and delegate observe different errors.

Modifications:
Send error to delegate in addition to task's promise

Result:
Closes #242


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -556,7 +556,10 @@ public class HTTPClient {
             }
         }.always { _ in
             setupComplete.succeed(())
-        }.cascadeFailure(to: task.promise)
+        }.whenFailure { error in
```
Lukasa, line 559: Are we confident that this event loop is an acceptable one to use to call into the delegate? We aren't paying any attention to the preference here. Seems like we need a `.hop(to: taskEL)` here.

weissi: Seconded. This also needs a test case. Also, I think there's a special method to call out to delegates that takes care of the hopping so we don't get it wrong. Essentially, all naked calls to the delegate should really go away and use the methods that are meant for that.

artemredkin: Test case for correct EL?

weissi: Make a test case that would fail _without_ the necessary hopping.

artemredkin: done

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2008,6 +2008,40 @@ class HTTPClientTests: XCTestCase {
...
+        let httpBin = HTTPBin()
+        let httpClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup))
+        defer {
+            XCTAssertNoThrow(try httpClient.syncShutdown())
+        }
+
+        let delegate = TestDelegate()
+        let request = try HTTPClient.Request(url: "http://localhost:\(httpBin.port)/get")
+        do {
+            try httpBin.shutdown()
```
weissi, line 2030: this should be in an `XCTAssertNoThrow`

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2008,6 +2008,40 @@ class HTTPClientTests: XCTestCase {
...
+            _ = try httpClient.execute(request: request, delegate: delegate).wait()
```
weissi, line 2031: please use

```
XCTThrowsError(try httpClient.execute(...).wait()) { error in
   XCTAssert(...)
}
```

we literally had security issues before because of this pattern.

artemredkin: done, thank you for catching this!

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -562,7 +562,7 @@ class HTTP1ConnectionProvider {
                     error = HTTPClient.NWErrorHandler.translateError(error)
                 }
             #endif
-            return self.eventLoop.makeFailedFuture(error)
+            return eventLoop.makeFailedFuture(error)
```
weissi, line 565: think this is the wrong way around, no?

artemredkin: No, I think this is correct, all channel related stuff should be on channel EL, not on providers default EL

weissi: @artemredkin sorry, I meant from `self.` to missing `self`, `eventLoop` isn't a local variable, is it?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -561,12 +569,6 @@ public class HTTPClient {
             }
 
             return future.flatMap {
-                let taskHandler = TaskHandler(task: task,
```
weissi, line 564: was this move from inside the closure to outside the closure deliberate?

artemredkin: yes, we need task handler to access `calloutTo` methods

Sources/AsyncHTTPClient/ConnectionPool.swift
artemredkin: it is, yes, this is a channel EL, I can rename it to make it more obvious

weissi: @artemredkin yes, I think that'd be a good idea tbh

PR #246, opened 2020-06-13
add callin tests

### LINKED ISSUE #222: Add tests for delegate callins
We need tests that verify that we tolerate backpressure future completions on any thread/EL

artemredkin:
Add tests that execute delegate backpressure call-ins on a different EL

Motivation:
Programming model: all call-outs and call-ins have to be on the task EL, we need to add a safeguard so it would not be accidentally broken.

Modifications:
Adds a test that calls-in on a different EL

Result:
Closes #222

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -2016,11 +2016,11 @@ class HTTPClientTests: XCTestCase {
             }
 
             func didReceiveHead(task: HTTPClient.Task<Void>, _: HTTPResponseHead) -> EventLoopFuture<Void> {
-                self.eventLoop.makeSucceededFuture(())
+                return self.eventLoop.makeSucceededFuture(())
             }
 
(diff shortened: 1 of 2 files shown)


Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2005,4 +2005,50 @@ class HTTPClientTests: XCTestCase {
...
+        }
+
+        let delegate = TestDelegate(eventLoop: second)
+        let request = try HTTPClient.Request(url: "http://localhost:\(httpServer.serverPort)/")
+        let future = httpClient.execute(request: request, delegate: delegate)
+
+        XCTAssertNoThrow(try httpServer.readInbound()) // .head
+        XCTAssertNoThrow(try httpServer.readInbound()) // .end
+
+        XCTAssertNoThrow(try httpServer.writeOutbound(.head(.init(version: .init(major: 1, minor: 1), status: .ok))))
```
weissi, line 2048: Shouldn’t we add a callin to the stream writer from another EL here?

artemredkin: thats already covered by `testUploadStreamingCallinToleratedFromOtsideEL`

PR #247, opened 2020-06-13
assume chunked on a stream with no length

### LINKED ISSUE #218: If not given a content-length, .stream should assume transfer-encoding:chunked
[THIS SHOULD BE FIXED AFTER #217]

```
        let request = try! HTTPClient.Request(url: "http://localhost:\(httpBin.port)/get",
                                              method: .POST,
                                              body: .stream { streamWriter in
```

this `Request` will actually error (sort of, see #217) but it should just assume `transfer-encoding: chunked`, the user clearly hasn't given a content length.

If the user manually set a `content-length` in `headers:`, then we should obvs use that.

artemredkin:
Assume chunked encoding if body stream length is nil.

Motivation:
Streams length parameter is optional to allow cases were stream length is not known in advance, but we do not support this in request validation. This PR aims to address that.

Modifications:
Modifies request validation to default to chunked encoding if body length is zero or to passed in `content-length` header
Adds a test

Result:
Closes #218


weissi:
@artemredkin i’m not quite sure about the description/title of this PR. It shouldn’t do anything special for 0 lenghts. But yes, chunked encoding should always be used unless the user tells us a content length. If the user says length 0 though we should insert a content-length:0 header

artemredkin:
Ah, its totally not about zero length, yes, only if length is not specified, fixed!

Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift
```diff
@@ -109,6 +109,7 @@ extension HTTPClientTests {
             ("testNothingIsLoggedAtInfoOrHigher", testNothingIsLoggedAtInfoOrHigher),
             ("testAllMethodsLog", testAllMethodsLog),
             ("testClosingIdleConnectionsInPoolLogsInTheBackground", testClosingIdleConnectionsInPoolLogsInTheBackground),
+            ("testUploadStreamingNoLength", testUploadStreamingNoLength),
```
weissi, line 112: I think we should have a few more test cases here. Essentially we have quite a big matrix of possibilities:

| HTTP Method kind   |      User sets      |  Body | Expectation |
|----------|:-------------:|------:|------:|
| .GET, .HEAD, .DELETE, .CONNECT, .TRACE |  _nothing_ | `nil` | Neither CL nor chunked |
| _other_ |  _nothing_ | `nil` | chunked |
| .GET, .HEAD, .DELETE, .CONNECT, .TRACE |  `content-length`   |   `nil` | CL=0 |
| _other_ |  `content-length`   |  `nil` |  CL=0 |
| .GET, .HEAD, .DELETE, .CONNECT, .TRACE |  `transfer-encoding: chunked`   |   `nil` | chunked |
| _other_ |  `transfer-encoding: chunked`   |   `nil` | chunked |
| .GET, .HEAD, .DELETE, .CONNECT, .TRACE |  CL & chunked (illegal)  |   `nil` | throws error |
| _other_ |  CL & chunked (illegal)  |   `nil` | throws error |
| .GET, .HEAD, .DELETE, .CONNECT, .TRACE |  _nothing_ | not `nil` | Neither CL nor chunked |
| _other_ |  _nothing_ | not  `nil` | chunked |
| .GET, .HEAD, .DELETE, .CONNECT, .TRACE |  `content-length`   |   not  `nil` | CL |
| _other_ |  `content-length`   |  not  `nil` |  CL |
| .GET, .HEAD, .DELETE, .CONNECT, .TRACE |  `transfer-encoding: chunked`   |   not  `nil` | chunked |
| _other_ |  `transfer-encoding: chunked`   |   not  `nil` | chunked |
| .GET, .HEAD, .DELETE, .CONNECT, .TRACE |  CL & chunked (illegal)  |   not  `nil` | throws error |
| _other_ |  CL & chunked (illegal)  |   not  `nil` | throws error |

I think we need to test them systematically or else we'll get more bugs here.

weissi: @artemredkin should we do this in this ticket or create an issue from it? I think it's important because the code looks tricky

artemredkin: I was planning on doing it in this issue

weissi: Amazing, thank you so much!!

artemredkin: hmm, I'm not sure some combinations are correct, for example:

|Method|Header|Body|Expectation|
|-------|--------|-----|-----------|
|other | nothing | nil | chunked|

but there is this bit in https://tools.ietf.org/html/rfc7230#section-3.3.2:
```A user agent SHOULD send a Content-Length in a request message when
   no Transfer-Encoding is sent and the request method defines a meaning
   for an enclosed payload body.  For example, a Content-Length header
   field is normally sent in a POST request even when the value is 0
   (indicating an empty payload body).
```
shouldn't this be `CL=0`?

artemredkin: Another example:

|Method|Header|Body|Expectation|
|-------|--------|-----|-----------|
|.GET, .HEAD, .DELETE, .CONNECT, .TRACE | content-length | nil | CL=0|

but https://tools.ietf.org/html/rfc7230#section-3.3.2:
```
   A user agent SHOULD NOT send a Content-Length header field when the request
   message does not contain a payload body and the method semantics do not
   anticipate such a body.
```

artemredkin: Also, from https://tools.ietf.org/html/rfc7230#section-4.3.8:
```
   A client MUST NOT send a message body in a TRACE request.
```
So we must exclude `.TRACE` from the second half of the table

artemredkin: And, if users sets `chunked` but with `nil` body, we _know_ there will be zero bytes, so we should default to Neither CL nor chunked for `.GET` and to `CL=0` for other

artemredkin: And last point is when user sets nothing but provides a body, we again can do two things:
if body length known - use CL, otherwise - chunked

weissi: @artemredkin sorry, I made a number of C&P errors when building this table :). I agree with all the changes.

weissi: We should only expect `chunked` if the user has told us _that_ they have a body but they didn't tell us the length.

We should set CL=0 if the user set `body: nil` UNLESS it's `.GET, .HEAD, .DELETE, .CONNECT, .TRACE`

artemredkin: @weissi done!

Issue #252, opened 2020-06-13
uploading more bytes into a body after having said the body is fully uploaded doesn't fail


weissi:
Here a test that shows the issue (currently gets stuck because of #250 the server just never replies)

```swift
    func testBodyUploadAfterEndFails() {
        let url = self.defaultHTTPBinURLPrefix + "/post"

        func uploader(_ streamWriter: HTTPClient.Body.StreamWriter) -> EventLoopFuture<Void> {
            let done = streamWriter.write(.byteBuffer(ByteBuffer(string: "X")))
            done.recover { error -> Void in
                XCTFail("unexpected error \(error)")
            }.whenSuccess {
                // This is executed when we have already sent the end of the request.
                done.eventLoop.execute {
                    streamWriter.write(.byteBuffer(ByteBuffer(string: "BAD BAD BAD"))).whenComplete { result in
                        switch result {
                        case .success:
                            XCTFail("we succeeded writing bytes after the end!?")
                        case .failure(let error):
                            XCTFail("\(error)")
                        }

                    }
                }
            }
            return done
        }

        XCTAssertThrowsError(
            try self.defaultClient.execute(request:
                                            Request(url: url,
                                                    body: .stream(length: 1, uploader))).wait()) { error in
            // TODO: check that the request fails with the appropriate error (or decide that that's okay)
        }

        // Quickly try another request and check that it works. If we by accident wrote some extra bytes into the
        // stream (and reuse the connection) that could cause problems.
        XCTAssertNoThrow(try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "/get").wait())
    }
```

I can probably be convinced that this is programmer error and we'd just `preconditionFailure`. I don't know, we definitely need to do something (and test that behaviour).

PR #255, opened 2020-06-15
check body length

### LINKED ISSUE #251: we don't check if the user supplies the correct number of bytes when specifying content-length
I don't have time right now to fix this but these too tests show the issue:

```swift
    func testContentLengthTooLongFails() {
        let url = self.defaultHTTPBinURLPrefix + "/post"
        XCTAssertThrowsError(
            try self.defaultClient.execute(request:
                                            Request(url: url,
                                                    body: .stream(length: 10) { streamWriter in
                                                        streamWriter.write(.byteBuffer(ByteBuffer(string: "1")))
                                                    })).wait()) { error in
            // TODO: content-length wrong error.
        }
        // Quickly try another request and check that it works.
        XCTAssertNoThrow(try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "/get").wait())
    }

    // currently gets stuck because of #250 the server just never replies
    func testContentLengthTooShortFails() {
        let url = self.defaultHTTPBinURLPrefix + "/post"
        let tooLong = "XBAD BAD BAD NOT HTTP/1.1\r\n\r\n"
        XCTAssertThrowsError(
            try self.defaultClient.execute(request:
                                            Request(url: url,
                                                    body: .stream(length: 1) { streamWriter in
                                                        streamWriter.write(.byteBuffer(ByteBuffer(string: tooLong)))
                                                    })).wait()) { error in
            // TODO: content-length wrong error.
        }
        // Quickly try another request and check that it works. If we by accident wrote some extra bytes into the
        // stream (and reuse the connection) that could cause problems.
        XCTAssertNoThrow(try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "/get").wait())
    }
```

This probably needs to be a blocker because with a connection pool, this is security relevant. We could end up sending private bytes for one request into the next request by accident.

artemredkin:
Adds checking for supplied body length.

Motivation:
We need to guard against incorrect body length since we can re-use connection, this is a potential security issue.

Modifications:
 - Adds a check in `HTTPTaskHandler` that if validated headers contain `Content-Length`, user sends exactly that amount of body data.
 - Two tests

Result:
Closes #251

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -641,7 +641,7 @@ internal class TaskHandler<Delegate: HTTPClientResponseDelegate>: RemovableChann
         case head
         case redirected(HTTPResponseHead, URL)
         case body
-        case end
+        case endOrError
     }
 
     let task: HTTPClient.Task<Delegate.Response>
@@ -782,7 +782,7 @@ extension TaskHandler: ChannelDuplexHandler {
         } catch {
             promise?.fail(error)
             self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
-            self.state = .end
+            self.state = .endOrError
             return
         }
 
@@ -796,20 +796,24 @@ extension TaskHandler: ChannelDuplexHandler {
         assert(head.version == HTTPVersion(major: 1, minor: 1),
                "Sending a request in HTTP version \(head.version) which is unsupported by the above `if`")
 
-        self.expectedBodyLength = head.headers[canonicalForm: "content-length"].first.flatMap { Int($0) }
+
+        let contentLengths = head.headers[canonicalForm: "content-length"]
+        assert(contentLengths.count <= 1)
+
+        self.expectedBodyLength = contentLengths.first.flatMap { Int($0) }
 
         context.write(wrapOutboundOut(.head(head))).map {
             self.callOutToDelegateFireAndForget(value: head, self.delegate.didSendRequestHead)
         }.flatMap {
             self.writeBody(request: request, context: context)
         }.flatMap {
+            context.eventLoop.assertInEventLoop()
             if let expectedBodyLength = self.expectedBodyLength, expectedBodyLength != self.actualBodyLength {
-                self.state = .end
+                self.state = .endOrError
                 let error = HTTPClientError.bodyLengthMismatch
                 self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
                 return context.eventLoop.makeFailedFuture(error)
             }
-            context.eventLoop.assertInEventLoop()
             return context.writeAndFlush(self.wrapOutboundOut(.end(nil)))
         }.map {
             context.eventLoop.assertInEventLoop()
@@ -818,10 +822,10 @@ extension TaskHandler: ChannelDuplexHandler {
         }.flatMapErrorThrowing { error in
             context.eventLoop.assertInEventLoop()
             switch self.state {
-            case .end:
+            case .endOrError:
                 break
             default:
-                self.state = .end
+                self.state = .endOrError
                 self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
             }
             throw error
@@ -839,14 +843,15 @@ extension TaskHandler: ChannelDuplexHandler {
                 // All writes have to be switched to the channel EL if channel and task ELs differ
                 if context.eventLoop.inEventLoop {
                     context.writeAndFlush(self.wrapOutboundOut(.body(part)), promise: promise)
+                    self.actualBodyLength += part.readableBytes
                 } else {
                     context.eventLoop.execute {
                         context.writeAndFlush(self.wrapOutboundOut(.body(part)), promise: promise)
+                        self.actualBodyLength += part.readableBytes
                     }
                 }
 
                 return promise.futureResult.map {
-                    self.actualBodyLength += part.readableBytes
                     self.callOutToDelegateFireAndForget(value: part, self.delegate.didSendRequestPart)
                 }
             })
@@ -904,12 +909,12 @@ extension TaskHandler: ChannelDuplexHandler {
         case .end:
             switch self.state {
             case .redirected(let head, let redirectURL):
-                self.state = .end
+                self.state = .endOrError
                 self.task.releaseAssociatedConnection(delegateType: Delegate.self, closing: self.closing).whenSuccess {
                     self.redirectHandler?.redirect(status: head.status, to: redirectURL, promise: self.task.promise)
                 }
             default:
-                self.state = .end
+                self.state = .endOrError
                 self.callOutToDelegate(promise: self.task.promise, self.delegate.didFinishRequest)
             }
         }
@@ -924,14 +929,14 @@ extension TaskHandler: ChannelDuplexHandler {
                 context.read()
             }
         case .failure(let error):
-            self.state = .end
+            self.state = .endOrError
             self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
         }
     }
 
     func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
         if (event as? IdleStateHandler.IdleStateEvent) == .read {
-            self.state = .end
+            self.state = .endOrError
             let error = HTTPClientError.readTimeout
             self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
         } else {
@@ -941,7 +946,7 @@ extension TaskHandler: ChannelDuplexHandler {
 
     func triggerUserOutboundEvent(context: ChannelHandlerContext, event: Any, promise: EventLoopPromise<Void>?) {
         if (event as? TaskCancelEvent) != nil {
-            self.state = .end
+            self.state = .endOrError
             let error = HTTPClientError.cancelled
             self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
             promise?.succeed(())
@@ -952,10 +957,10 @@ extension TaskHandler: ChannelDuplexHandler {
 
     func channelInactive(context: ChannelHandlerContext) {
         switch self.state {
-        case .end:
+        case .endOrError:
             break
         case .body, .head, .idle, .redirected, .sent:
-            self.state = .end
+            self.state = .endOrError
             let error = HTTPClientError.remoteConnectionClosed
             self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
         }
@@ -966,7 +971,7 @@ extension TaskHandler: ChannelDuplexHandler {
         switch error {
         case NIOSSLError.uncleanShutdown:
             switch self.state {
-            case .end:
+            case .endOrError:
                 /// Some HTTP Servers can 'forget' to respond with CloseNotify when client is closing connection,
                 /// this could lead to incomplete SSL shutdown. But since request is already processed, we can ignore this error.
                 break
@@ -975,11 +980,11 @@ extension TaskHandler: ChannelDuplexHandler {
                 /// We can also ignore this error like `.end`.
                 break
             default:
-                self.state = .end
+                self.state = .endOrError
                 self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
             }
         default:
-            self.state = .end
+            self.state = .endOrError
             self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
         }
     }
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -651,6 +651,8 @@ internal class TaskHandler<Delegate: HTTPClientResponseDelegate>: RemovableChann
     let logger: Logger // We are okay to store the logger here because a TaskHandler is just for one request.
 
     var state: State = .idle
+    var expectedBodyLength: Int? = nil
+    var actualBodyLength: Int = 0
     var pendingRead = false
     var mayRead = true
     var closing = false {
@@ -794,11 +796,19 @@ extension TaskHandler: ChannelDuplexHandler {
         assert(head.version == HTTPVersion(major: 1, minor: 1),
                "Sending a request in HTTP version \(head.version) which is unsupported by the above `if`")
 
+        self.expectedBodyLength = head.headers[canonicalForm: "content-length"].first.flatMap { Int($0) }
+
         context.write(wrapOutboundOut(.head(head))).map {
             self.callOutToDelegateFireAndForget(value: head, self.delegate.didSendRequestHead)
         }.flatMap {
             self.writeBody(request: request, context: context)
         }.flatMap {
+            if let expectedBodyLength = self.expectedBodyLength, expectedBodyLength != self.actualBodyLength {
+                self.state = .end
+                let error = HTTPClientError.bodyLengthMismatch
+                self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
+                return context.eventLoop.makeFailedFuture(error)
+            }
             context.eventLoop.assertInEventLoop()
             return context.writeAndFlush(self.wrapOutboundOut(.end(nil)))
         }.map {
@@ -836,6 +846,7 @@ extension TaskHandler: ChannelDuplexHandler {
                 }
 
                 return promise.futureResult.map {
+                    self.actualBodyLength += part.readableBytes
                     self.callOutToDelegateFireAndForget(value: part, self.delegate.didSendRequestPart)
                 }
             })
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -651,7 +651,7 @@ internal class TaskHandler<Delegate: HTTPClientResponseDelegate>: RemovableChann
     let logger: Logger // We are okay to store the logger here because a TaskHandler is just for one request.
 
     var state: State = .idle
-    var expectedBodyLength: Int? = nil
+    var expectedBodyLength: Int?
     var actualBodyLength: Int = 0
     var pendingRead = false
     var mayRead = true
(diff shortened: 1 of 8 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -794,11 +796,19 @@ extension TaskHandler: ChannelDuplexHandler {
         assert(head.version == HTTPVersion(major: 1, minor: 1),
                "Sending a request in HTTP version \(head.version) which is unsupported by the above `if`")
 
+        self.expectedBodyLength = head.headers[canonicalForm: "content-length"].first.flatMap { Int($0) }
```
weissi, line 799: this needs to error if there's more than one CL header

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -794,11 +796,19 @@ extension TaskHandler: ChannelDuplexHandler {
...
+
         context.write(wrapOutboundOut(.head(head))).map {
             self.callOutToDelegateFireAndForget(value: head, self.delegate.didSendRequestHead)
         }.flatMap {
             self.writeBody(request: request, context: context)
         }.flatMap {
+            if let expectedBodyLength = self.expectedBodyLength, expectedBodyLength != self.actualBodyLength {
```
weissi, line 806: At least two things missing here:

- a `hop` to the correct EventLoop
- an assertion that we are on the correct EL
- [if not already present: a test that invokes this from the wrong EL]

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -794,11 +796,19 @@ extension TaskHandler: ChannelDuplexHandler {
...
+                self.state = .end
```
weissi, line 807: `.end` doesn't sound like a good state for an error, right? We need an `.error` state here I think or else we'll get into trouble later.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -836,6 +846,7 @@ extension TaskHandler: ChannelDuplexHandler {
                 }
 
                 return promise.futureResult.map {
+                    self.actualBodyLength += part.readableBytes
```
weissi, line 849: - assertion about the correct EL missing
- test that would fail this assertion right now probably missing.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2051,4 +2051,35 @@ class HTTPClientTests: XCTestCase {
...
+        XCTAssertThrowsError(
+            try self.defaultClient.execute(request:
+                                            Request(url: url,
+                                                    body: .stream(length: 10) { streamWriter in
+                                                        streamWriter.write(.byteBuffer(ByteBuffer(string: "1")))
+                                                    })).wait()) { error in
+            XCTAssertEqual(error as! HTTPClientError, HTTPClientError.bodyLengthMismatch)
+        }
+        // Quickly try another request and check that it works.
+        XCTAssertNoThrow(try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "/get").wait())
```
weissi, line 2066: Sorry, I know I wrote this test but we should check the correct `connectionNumber` and `requestNumber` from 

```
internal struct RequestInfo: Codable {
    var data: String
    var requestNumber: Int
    var connectionNumber: Int
}
```

just to be sure we get a fresh connection.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2051,4 +2051,35 @@ class HTTPClientTests: XCTestCase {
...
+            try self.defaultClient.execute(request:
+                                            Request(url: url,
+                                                    body: .stream(length: 1) { streamWriter in
+                                                        streamWriter.write(.byteBuffer(ByteBuffer(string: tooLong)))
+                                                    })).wait()) { error in
+            XCTAssertEqual(error as! HTTPClientError, HTTPClientError.bodyLengthMismatch)
+        }
+        // Quickly try another request and check that it works. If we by accident wrote some extra bytes into the
+        // stream (and reuse the connection) that could cause problems.
+        XCTAssertNoThrow(try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "/get").wait())
```
weissi, line 2083: similar here, we should validate request/connection number.

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: shouldn't `validateHeaders` do that?

artemredkin: I can add a return from `validate` with body length, as an alternative

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: ah, assertion is there, I just pasted the code above, will fix, also, we use special `callout` helper here for all EL correctness: `failTaskAndNotifyDelegate`

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: I'm happy if we have a test that proves that we error if the user sends two `content-length`s. In this code, I think we should just add an

```
assert(head.headers[canonicalForm: "content-length"].count <= 1)
```

to be sure.

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: it might be a good idea, but not for this PR, since it will require to change other places and all switch statements. I don't think that we can get into trouble, since we don't re-use the handler currently, and `.end` is a terminal state anyway, wdyt?

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: @artemredkin the callout looks great but you're accessing a mutable property on `self` which can only be correct if we're on the right EL.

I kinda doubt that we're on the right EL because `context.write` returns a future from the `Channel`'s EL and `self.writeBody` returns a user-controlled future. Unsure how we can be on `taskEL` for sure.

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: @artemredkin is there precedent for using `.end` for error? If yes, I'd rename to `endOrError`

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: it's even better, we _remove_ all content length headers and insert them ourselves :) except one case, I'll add a test for that (and an assert).

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: ahhh, you are totally correct, damn those event loops :P

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: yep, there is (since we had no pool before so handler was single-use anyway). `endOrError` sounds great!

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: @artemredkin very nice! However if the user sets two CL headers, shouldn't we error? Maybe it's fine because we're (now) checking the body so the user should get an error anyway if we pick the "wrong" one :P. Just mean it could be confusing if the user does something like this

```
var headers = HTTPHeaders(["content-length": 17])
let request = ...(headers)
[...] // lots of code

func setBody(_ body: String, request: inout Request) {
    request.headers.add("content-length": "\(body.utf8.count)")
    request.body = ... .stream { writer in write.write(... body ...) }
}

setBody("hello", &request)
self.myHTTPClient.execute(request)
```

and now you'll get a `bodySizeMismatchError` but the code (in the `setBody` function) looks correct. So I just felt we _could_ be nicer to the user and diagnose the actual issue (which is that there are two CL headers present) instead of a follow up one (because we picked the first body header).

Said that I'm happy if we do sanitisation + checking of the actually sent body size.

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: Hehe, we should probably start having a separation layer between NIO & AHC. AHC (rightly) chooses that it wants to cross ELs for performance but it's tricky to always get it right.

If we had a layer (a struct wrapping `Channel`) which would automatically `hop` all futures returned from the `Channel` operations, this should be much simpler. Then all the hopping it in one place.

weissi: [not in this PR ofc]

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: That will happen only if the stream has no length is not specified. In all other cases we actually ignore the user-provided headers, you think we should fail on those?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
artemredkin: hmm, do we even need this part? we already have a test that checks that, no?

artemredkin: ah, I think I know what you mean!

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: @artemredkin wait, why are we looking for the first CL header then if this code is only run if the user specified _no_ CL headers?

artemredkin: this code is run always, if the content-length was not specified, then `expectedLength` will be `nil` and final check will not be executed

weissi: alright, let's just add the assertion&test (if not present for all the possible cases) in case somebody modifies the sanitisation code.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
artemredkin: done!

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
artemredkin: done!

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: moved the increment to where it is certainly on correct EL

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done, plus modified one of the test to callin from outside the EL

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done!

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -829,9 +843,11 @@ extension TaskHandler: ChannelDuplexHandler {
                 // All writes have to be switched to the channel EL if channel and task ELs differ
                 if context.eventLoop.inEventLoop {
                     context.writeAndFlush(self.wrapOutboundOut(.body(part)), promise: promise)
+                    self.actualBodyLength += part.readableBytes
                 } else {
                     context.eventLoop.execute {
                         context.writeAndFlush(self.wrapOutboundOut(.body(part)), promise: promise)
+                        self.actualBodyLength += part.readableBytes
```
weissi, line 850: these two lines need to be flipped. `writeAndFlush` runs essentially arbitrary code (from the pipeline but also from the `promise`. So you need to do the state modification before that, or else you can get reordering.

Same a few lines above.

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -357,8 +361,8 @@ internal struct HTTPResponseBuilder {
     }
 }
 
-let globalRequestCounter = NIOAtomic<Int>.makeAtomic(value: 0)
-let globalConnectionCounter = NIOAtomic<Int>.makeAtomic(value: 0)
+//let globalRequestCounter = NIOAtomic<Int>.makeAtomic(value: 0)
```
weissi: comments

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2051,4 +2053,45 @@ class HTTPClientTests: XCTestCase {
...
+                        body: .stream(length: 1) { streamWriter in
+                            streamWriter.write(.byteBuffer(ByteBuffer(string: tooLong)))
+                                                    })).wait()) { error in
+            XCTAssertEqual(error as! HTTPClientError, HTTPClientError.bodyLengthMismatch)
+        }
+        // Quickly try another request and check that it works. If we by accident wrote some extra bytes into the
+        // stream (and reuse the connection) that could cause problems.
+        var response = try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
+        let info = try response.body!.readJSONDecodable(RequestInfo.self, length: response.body!.readableBytes)
+        XCTAssertEqual(info!.connectionNumber, 1)
```
weissi, line 2094: this cannot be right, this will only be correct if you run this tests on its own. To get the right connection number, execute a random, uninteresting request beforehand and then add 1 to it.

artemredkin: the way it was implemented before was actually not good, imho. If you use a global counter (as it was implemented before), then you cannot run test that do anything concurrently. I re-wrote that to have per-server/per-connection counters

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
artemredkin: oops, fixed, thanks!

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: @artemredkin that's true, but connection counter is still global, no? We could attach the counters to the HTTPBin maybe?

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: fixed, thanks!

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
artemredkin: no, connection counter is not global anymore, every new instance of HTTPBin has its own connection counter and every new connection (ie handler) has its own request counter

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -794,12 +796,23 @@ extension TaskHandler: ChannelDuplexHandler {
...
+
+        self.expectedBodyLength = contentLengths.first.flatMap { Int($0) }
+
         context.write(wrapOutboundOut(.head(head))).map {
             self.callOutToDelegateFireAndForget(value: head, self.delegate.didSendRequestHead)
         }.flatMap {
             self.writeBody(request: request, context: context)
         }.flatMap {
             context.eventLoop.assertInEventLoop()
+            if let expectedBodyLength = self.expectedBodyLength, expectedBodyLength != self.actualBodyLength {
```
weissi, line 810: @artemredkin is the checking here really good enough? What if a user does this:

1. `content-length: 1`
2. write body `xGET / HTTP/1.1\r\n\r\n`
3. fulfil the future returned from `self.writeBody`

We only notice that something's wrong in step 3 _after_ we have already written too much (in step 2). In this case, this would actually execute another request which is basically request smuggling. So if somebody builds a library/proxy with AHC that assumes (good assumption I think) that `.execute`ing 1 request only makes 1 request, they'll now have a request smuggling vulnerability.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: Ok, sounds good!

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: it is not good enough, yes, but that whats #252 is for :) unless you want to duplicate it to this issue? probably should fix here though

weissi: ah okay! Yes, I'm happy to fixing it in two!

weissi: @artemredkin ah sorry, #252  was meant for this:

1 content-length: 1
2. write body `x`
3. fulfil the future returned from self.writeBody
[...]
4. write body `GET / HTTP/1.1\r\n`

So #252 was meant for writes _after_ the end was indicated.

weissi: But I'm still fine fixing it in 2 parts, I'll make #252 a release blocker though because if only the #252 fix fixes too many bytes being sent, we need that.

artemredkin: ah, I see, let me then fix this here

artemredkin: sounds good

artemredkin:

artemredkin:

artemredkin:

artemredkin:

PR #257, opened 2020-06-15
fix test faling with NIOTS

### LINKED ISSUE #256: tests fail on macOS
```
Test Case '-[AsyncHTTPClientTests.HTTPClientInternalTests testUncleanCloseThrows]' started.
Fatal error: ServerBootstrap is only compatible with MultiThreadedEventLoopGroup and SelectableEventLoop. You tried constructing one with NIOTransportServices.NIOTSEventLoop which is incompatible.: file /Users/johannes/devel/swift-nio/Sources/NIO/Bootstrap.swift, line 87
Exited with signal code 4
```

artemredkin:
Fixes test crashing on precondition.

Motivation:
When tests are executed on macOS client ELG is actually a `NIOTS` ELG, so it's incompatible with `NIOHTTP1TestServer`

Modifications:
Use separate ELG in `testUncleanCloseThrows`

Result:
Closes #256

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
@@ -797,7 +797,7 @@ class HTTPClientInternalTests: XCTestCase {
     }
 
     func testUncleanCloseThrows() {
-        let server = NIOHTTP1TestServer(group: self.clientGroup)
+        let server = NIOHTTP1TestServer(group: MultiThreadedEventLoopGroup(numberOfThreads: 1))
         defer {
             XCTAssertNoThrow(try server.stop())
         }
--- a/Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
@@ -797,9 +797,11 @@ class HTTPClientInternalTests: XCTestCase {
     }
 
     func testUncleanCloseThrows() {
-        let server = NIOHTTP1TestServer(group: MultiThreadedEventLoopGroup(numberOfThreads: 1))
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        let server = NIOHTTP1TestServer(group: group)
         defer {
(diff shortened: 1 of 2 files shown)


Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
```diff
@@ -797,7 +797,7 @@ class HTTPClientInternalTests: XCTestCase {
     }
 
     func testUncleanCloseThrows() {
-        let server = NIOHTTP1TestServer(group: self.clientGroup)
+        let server = NIOHTTP1TestServer(group: MultiThreadedEventLoopGroup(numberOfThreads: 1))
```
weissi, line 800: who shuts this MTELG down?

weissi: I think we have `self.serverGroup` which could be used

artemredkin: oooops, no one, good catch, thank you! no, we don't have that in internal tests class

artemredkin: and its only monday... 😔

weissi: > oooops, no one, good catch, thank you! no, we don't have that in internal tests class

We should. I was probably lazy and changed only all of the non-internal tests

artemredkin: so far we won't be using it anywhere else in that class (it usually just creates `HTTPBin`), still worth it?

weissi: @artemredkin I think we should have a `defaultHTTPBin` and `defaultClient` and `client/serverGroup` all set up in `setUp`/`tearDown`. That makes tests more concise and therefore better

weissi: but we don't need to do it right in this PR, up to you

PR #259, opened 2020-06-15
Fixed an issue where redirects to socket path-based servers from any server was always allowed

### LINKED ISSUE #230: It's likely redirects for sockets and regular hosts never behaved as expected
While refactoring the unix domain socket implementation in #228, I noticed that redirects likely haven't been checking URL schemes correctly for some time now, or I'm not completely following the logic and intended use of `RedirectHandler`.

My _assumption_ is that `RedirectHandler.redirectTarget(status:, headers:)` validates the incoming header, and returns a URL to redirect to if it passes all the necessary checks.

If I'm right, that means that the redirect logic checks the current request rather than the new URL to see if it can redirect to it - this likely never worked as designed due to the fact that it would always pass, since we have a request that is already valid, and we are checking it once again for validity...

It's also likely that this check never worked, but redirects to foreign URLs were still possible, since creating the Request later in the process would eventually fail, but there are no checks to ensure the Request hasn't been redirected from a host-based server to a socket based one, so something is still fishy in that regard.

If my assumption isn't misguided in some way, I suggest we refactor the check for the scheme as follows:
image

The new `supportsRedirects(to:)` method (refactored, since it is only being used in this scenario now):
image

Additional tests should clarify if this is working as designed or not.

dimitribouniol:
Motivation:

An arbitrary HTTP(S) server should not be able to trigger redirects, and thus activity, to a local socket-path based server, though the opposite may be a valid scenario. Currently, requests in either direction are allowed since the checks don't actually check the destination scheme.

Modifications:

- Added a method to HTTPBin to redirect to the specified target.
- Added failing tests that perform redirects from a regular server to a socket-based server and vice versa.
- Refactored `hostSchemes`/`unixSchemes` to `hostRestrictedSchemes`/`allSupportedSchemes`, which better describes what they do.
- Refactored `Request.supports()` to `Request.supportsRedirects(to:)` since it is only used by Redirects now.
- Check the destination URL's scheme rather than the current URL's scheme when validating a redirect.

Result:

Closes #230

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -107,8 +107,8 @@ extension HTTPClient {
             /// UNIX Domain Socket HTTP request.
             case unixSocket(_ scheme: UnixScheme)
 
-            private static var hostSchemes = ["http", "https"]
-            private static var unixSchemes = ["unix", "http+unix", "https+unix"]
+            private static var hostRestrictedSchemes: Set = ["http", "https"]
+            private static var allSupportedSchemes: Set = ["http", "https", "unix", "http+unix", "https+unix"]
 
             init(forScheme scheme: String) throws {
                 switch scheme {
@@ -158,12 +158,14 @@ extension HTTPClient {
                 }
             }
 
-            func supports(scheme: String) -> Bool {
+            func supportsRedirects(to scheme: String?) -> Bool {
+                guard let scheme = scheme?.lowercased() else { return false }
+
                 switch self {
                 case .host:
-                    return Kind.hostSchemes.contains(scheme)
+                    return Kind.hostRestrictedSchemes.contains(scheme)
                 case .unixSocket:
-                    return Kind.unixSchemes.contains(scheme)
+                    return Kind.allSupportedSchemes.contains(scheme)
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -162,12 +162,14 @@ extension HTTPClient {
                 }
             }
 
-            func supports(scheme: String) -> Bool {
+            func supportsRedirects(to scheme: String?) -> Bool {
```
Lukasa, line 165: Why has this become optional?

dimitribouniol: Mostly to simplify the call site. I can change that to

```swift
guard let scheme = url.scheme, self.request.kind.supportsRedirects(to: scheme) {...
```

if you’d prefer.

Lukasa: Nah, this works too, just wanted to be sure I understood it.

dimitribouniol:
Thanks! Happy to contribute where I can 🙂

dimitribouniol:
Rebased to keep up to date with master.

dimitribouniol:
Rebased to keep up to date with master.

This is failing with:
```
Test Case 'HTTPClientTests.testRacePoolIdleConnectionsAndGet' started at 2020-06-20 17:20:52.345
```
... which I don't believe is related to these changes (first time I've run across it)

artemredkin:
Yep, this is test failure is related to #260

artemredkin:

artemredkin:

artemredkin:
Thank you, @dimitribouniol !

dimitribouniol:
Thank _you_! 😁

PR #261, opened 2020-06-16
Updated swift-nio to 2.18.0

dimitribouniol:
Not sure if this is worth opening a PR for, since any new checkouts will never see this issue, and it's only necessary due to tests, so please feel free to close this if it's not necessary.

Motivation:

Due to using new ByteBuffer initializers from https://github.com/apple/swift-nio/releases/tag/2.18.0 in some tests (specifically, `ByteBuffer(string: ...)`), the project will no longer compile if an older version of swift-nio is checked out.

Modifications:

Updated swift-nio from 2.16.0 to 2.18.0.

Result:

SPM should now grab the latest version of swift-nio if this version of async-http-client is used.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ let package = Package(
         .library(name: "AsyncHTTPClient", targets: ["AsyncHTTPClient"]),
     ],
     dependencies: [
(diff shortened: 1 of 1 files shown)


Lukasa:

dimitribouniol:
Happy as always to be solving the hard problems 😅

Issue #265, opened 2020-06-18
flaky test HTTPClientTests.testContentLengthTooLongFails


weissi:
```
Test Case 'HTTPClientTests.testContentLengthTooLongFails' started at 2020-06-16 19:10:55.784
/code/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:2323: error: HTTPClientTests.testContentLengthTooLongFails : XCTAssertThrowsError failed: did not throw error - 
/code/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:2338: error: HTTPClientTests.testContentLengthTooLongFails : failed - Body missing: Response(host: "localhost", status: NIOHTTP1.HTTPResponseStatus.badRequest, version: HTTP/1.1, headers: [("Connection", "close"), ("Content-Length", "0")], body: nil)
Test Case 'HTTPClientTests.testContentLengthTooLongFails' failed (0.046 seconds)
```

PR #267, opened 2020-06-19
fix missing connect timeout and make tests safer

artemredkin:
Fix missing connect timeout and make some tests faster and safer.

Motivation:
1. During pool implementation/refactoring we lost connection timeout, this PR fixes that
2. `testConnectionFailError` is slow, 10 seconds, due to not having a timeout set
3. `testConnectionFailError` and `testTLSVersionError` use dangerous `do-try-XCFail-catch` pattern

Modifications:
1. Set connection timeout if specified
2. Set connect timeout to 150 ms to make `testTLSVersionError` faster
3. Use `XCTAssertThrowsError` in `testConnectionFailError` and `testTLSVersionError`
4. Add general connection timeout test, since `testConnectionFailError` is NIOTS-specific.

Result:
Fixes an issue where connection timeout configuration was not respected, hopefully makes tests more robust

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientNIOTSTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientNIOTSTests.swift
@@ -89,7 +89,6 @@ class HTTPClientNIOTSTests: XCTestCase {
             switch error {
             case ChannelError.connectTimeout(let timeout):
                 XCTAssertLessThanOrEqual(timeout, .milliseconds(150))
-                break
             default:
                 XCTFail("Unexpected error: \(error)")
             }
@@ -109,9 +108,9 @@ class HTTPClientNIOTSTests: XCTestCase {
                 XCTAssertNoThrow(try httpBin.shutdown())
             }
 
-        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/get").wait()) { error in
-            XCTAssertEqual((error as? HTTPClient.NWTLSError)?.status, errSSLHandshakeFail)
-        }
+            XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(httpBin.port)/get").wait()) { error in
+                XCTAssertEqual((error as? HTTPClient.NWTLSError)?.status, errSSLHandshakeFail)
+            }
         #endif
     }
 }
--- a/Tests/AsyncHTTPClientTests/HTTPClientNIOTSTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientNIOTSTests.swift
@@ -74,19 +74,25 @@ class HTTPClientNIOTSTests: XCTestCase {
     func testConnectionFailError() {
         guard isTestingNIOTS() else { return }
         let httpBin = HTTPBin(ssl: true)
-        let httpClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup))
+        let httpClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup),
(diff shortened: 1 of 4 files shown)


Tests/AsyncHTTPClientTests/HTTPClientNIOTSTests.swift
```diff
@@ -74,19 +74,24 @@ class HTTPClientNIOTSTests: XCTestCase {
...
 
-        do {
-            _ = try httpClient.get(url: "https://localhost:\(port)/get").wait()
-            XCTFail("This should have failed")
-        } catch ChannelError.connectTimeout {
-        } catch {
-            XCTFail("Error should have been ChannelError.connectTimeout not \(type(of: error))")
+        XCTAssertThrowsError(try httpClient.get(url: "https://localhost:\(port)/get").wait()) { error in
+            switch error {
+            case ChannelError.connectTimeout(let timeout):
```
artemredkin, line 90: @weissi do you know if there is a safer pattern for this case, where enum has an associated value?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -422,6 +422,25 @@ class HTTPClientTests: XCTestCase {
...
+    func testConnectTimeout() throws {
+        let httpClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup),
+                                    configuration: .init(timeout: .init(connect: .milliseconds(100), read: .milliseconds(150))))
+
+        defer {
+            XCTAssertNoThrow(try httpClient.syncShutdown())
+        }
+
+        // This must throw as 198.51.100.254 is reserved for documentation only
+        XCTAssertThrowsError(try httpClient.get(url: "http://198.51.100.254:65535/get").wait()) { error in
```
artemredkin, line 434: @weissi I looked this up here:
https://github.com/apple/swift-nio/blob/6046aaa6291fb075c83b5c22bad41b9780b2d977/Tests/NIOTests/ChannelTests.swift#L1060
is this ok? or is there a better way to get a connect timeout?

artemredkin:
failed due to #265

artemredkin:

Tests/AsyncHTTPClientTests/HTTPClientNIOTSTests.swift
Lukasa: What don't you like about this pattern?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
Lukasa: This should be just fine. Misconfigured networks can make this test fail, but without a complete user space IP and TCP implementation we don't have a better option.

Tests/AsyncHTTPClientTests/HTTPClientNIOTSTests.swift
artemredkin: This is potentially dangerous, if you miss the `XCTFail` call in the default clause, you might miss when the test breaks. This is why I'm replacing `do-try-XCTFail-catch` pattern with `XCTAssertThrowsError`

Lukasa: I see what you mean. No, there isn't really a better pattern here sadly.

weissi: I usually do this

```
XCTAssertEqul(.connectTimeout(.milliSeconds(150)), error as? ChannelError)
```

as it's much safer

PR #270, opened 2020-06-23
fail if user tries writing bytes after request is sent

### LINKED ISSUE #252: uploading more bytes into a body after having said the body is fully uploaded doesn't fail

artemredkin:
Fail if user tries writing body parts after request is sent.

Motivation:
This is could be a security issue.

Modifications:
1. Check state when writing body parts and fail if its not idle
2. Added a test

Result:
Closes #252


Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
```diff
@@ -424,11 +424,10 @@ class HTTPClientInternalTests: XCTestCase {
         let randoEL = group.next()
 
         let httpClient = HTTPClient(eventLoopGroupProvider: .shared(group))
-        let promise: EventLoopPromise<Channel> = httpClient.eventLoopGroup.next().makePromise()
-        let httpBin = HTTPBin(channelPromise: promise)
+        let server = NIOHTTP1TestServer(group: MultiThreadedEventLoopGroup(numberOfThreads: 1))
```
weissi, line 427: aren't we leaking the `group` here? It needs to be stopped.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -901,6 +899,19 @@ extension TaskHandler: ChannelDuplexHandler {
...
     }
 
+    private func writeBodyPart(context: ChannelHandlerContext, part: IOData, promise: EventLoopPromise<Void>) {
+        switch self.state {
+        case .idle:
+            self.actualBodyLength += part.readableBytes
+            context.writeAndFlush(self.wrapOutboundOut(.body(part)), promise: promise)
+        default:
+            let error = HTTPClientError.writeAfterRequestSent
+            promise.fail(error)
```
weissi, line 909: This can't be quite right (and I think we should have a test for this). We're calling out to the user here _before_ we change our state. So we may get user code "sandwiched in" here.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -901,6 +899,19 @@ extension TaskHandler: ChannelDuplexHandler {
         }
     }
 
+    private func writeBodyPart(context: ChannelHandlerContext, part: IOData, promise: EventLoopPromise<Void>) {
+        switch self.state {
+        case .idle:
+            self.actualBodyLength += part.readableBytes
```
weissi, line 905: I _think_ (would need a test here too) that we're policing the body length too late here, no? We seem to just add the readableBytes but I can't see that we check that it's actually less than or equal the number of body bytes allowed.

What happens if we do say `content-length: 1` and then send `XABCDEFG`, aren't we then still sending `ABCDEFG`? I think we are and that'd be a security vulnerability because we could smuggle a request (say we did `XGET /something HTTP/1.1\r\n\r\n`

artemredkin: thats a great point, yeah, we only check in the end, let me see how I can handle it here

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
artemredkin: oops, we do

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: ok, I added an early fail here

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: I've re-ordered things, not sure yet how to test it though, I'll have to get access to a handlers internal state... Could you elaborate on "sandwiched" code a bit?

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
artemredkin: fixed

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: @artemredkin this _looks_ good but given that it's security relevant, could we add a test specifically for that case (one that'd fail without the new condition)

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: @artemredkin sure. If you do `promise.fail(...)`, then the promise's `whenFailure` and other code will run _inline_ here. So we detected an error (ie. we _should_ be in state `.endOrError`) but we call out _before_ we entered the `.endOrError` state. So if the user for example uses another method that does a state check, we potentially have an issue because their code runs before the `self.state = .endOrError` so they won't see it. That could lead us to make wrong decisions.

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: we already have it, by accident, we have two tests for content-length/body-length checks:
 - testContentLengthTooLongFails
 - testContentLengthTooShortFails
Last one will fail at the end of the request, before we send `.end`, but the first one will now be failed by this new check

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: ah, that is so obvious now :) I'll think about how to test this

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: Potentially we can split the error message, just to be sure

weissi: @artemredkin I think we need a new one that tests that we do _not_ send the bytes.

weissi: Because as soon as the bytes are on the wire, we created a security vulnerability. Even if we detect later that something's wrong, it'll be too late.

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: @Lukasa could you pick up from Johannes here? I've added a test to verify that if delegate is called here, state is changed. I haven't figured out how to attach anything to that promise though... It seems that I need to have access to `channel.write(HTTPClient.request, promise)`, and I don't think anything in the code can attach to it

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: @Lukasa same here, I've added a test to check that we do not send more bytes then body length, although I'm not 100% sure its foolproof

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
```diff
@@ -1025,4 +1031,50 @@ class HTTPClientInternalTests: XCTestCase {
...
+        delegate.handler = handler
+        try channel.pipeline.addHandler(handler).wait()
+
+        var request = try Request(url: "http://localhost:8080/post")
+        request.body = .stream(length: 1) { writer in
+            writer.write(.byteBuffer(ByteBuffer(string: "1234")))
+        }
+
+        XCTAssertThrowsError(try channel.writeOutbound(request))
+        XCTAssertTrue(delegate.triggered)
```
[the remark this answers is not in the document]
artemredkin: great idea, done!

Lukasa:
Sadly I don't have admin on this repo, so we'll need @tomerd to dismiss @weissi's out-of-date review.

tomerd:
@Lukasa go ahead

tomerd:
cc @ktoso

PR #273, opened 2020-06-23
Add TestBackpressure test

adam-fowler:
Add response delegate backpressure test. Creates delegate that returns a futureResult from a `scheduleTask` that will finish after one second has passed. The `processingBodyPart` flag is set at the beginning of `didReceiveBodyPart` and cleared when the scheduled task runs. Add check to see if this flag is set when entering `didReceiveBodyPart`. If so then the backpressure wasn't applied and the previous `didReceiveBodyPart` task is still running. 

Also added count which increments at entry to `didReceiveBodyPart` and decrements on scheduled task being run.

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -2581,13 +2581,13 @@ class HTTPClientTests: XCTestCase {
         }
 
         let backpressureResponseDelegate = BackpressureResponseDelegate()
-        guard let request = try? HTTPClient.Request(url: self.defaultHTTPBinURLPrefix + "zeros/100000") else {
+        guard let request = try? HTTPClient.Request(url: self.defaultHTTPBinURLPrefix + "zeros/150000") else {
             XCTFail("Failed to init Request")
             return
         }
         XCTAssertNoThrow(try client.execute(request: request, delegate: backpressureResponseDelegate).wait())
         XCTAssertEqual(backpressureResponseDelegate.didntWait, false)
-        XCTAssertGreaterThan(backpressureResponseDelegate.totalCount, 1)
+        XCTAssertGreaterThan(backpressureResponseDelegate.totalCount, 2)
         XCTAssertEqual(backpressureResponseDelegate.count, 0)
     }
 }
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -2534,10 +2534,11 @@ class HTTPClientTests: XCTestCase {
         XCTAssertEqual(info.requestNumber, 1)
     }
 
-    func testBackpressure() {
+    func testDownloadBackpressure() {
         class BackpressureResponseDelegate: HTTPClientResponseDelegate {
             typealias Response = Void
             var count = 0
+            var totalCount = 0
             var processingBodyPart = false
             var didntWait = false
             var lock = Lock()
@@ -2557,6 +2558,7 @@ class HTTPClientTests: XCTestCase {
                     }
                     processingBodyPart = true
                     count += 1
+                    totalCount += 1
                 }
                 // wait one second before returning a successful future
                 return task.eventLoop.scheduleTask(in: .milliseconds(200)) {
@@ -2571,21 +2573,21 @@ class HTTPClientTests: XCTestCase {
             func didFinishRequest(task: HTTPClient.Task<Response>) throws {}
         }
 
-        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 5)
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 3)
         let client = HTTPClient(eventLoopGroupProvider: .shared(elg))
         defer {
             XCTAssertNoThrow(try client.syncShutdown())
             XCTAssertNoThrow(try elg.syncShutdownGracefully())
         }
 
-        let data = Data(count: 65273)
         let backpressureResponseDelegate = BackpressureResponseDelegate()
-        guard let request = try? HTTPClient.Request(url: self.defaultHTTPBinURLPrefix + "get", body: .data(data)) else {
+        guard let request = try? HTTPClient.Request(url: self.defaultHTTPBinURLPrefix + "zeros/100000") else {
             XCTFail("Failed to init Request")
             return
         }
         XCTAssertNoThrow(try client.execute(request: request, delegate: backpressureResponseDelegate).wait())
         XCTAssertEqual(backpressureResponseDelegate.didntWait, false)
+        XCTAssertGreaterThan(backpressureResponseDelegate.totalCount, 1)
         XCTAssertEqual(backpressureResponseDelegate.count, 0)
     }
 }
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
         XCTAssertEqual(info.connectionNumber, 1)
         XCTAssertEqual(info.requestNumber, 1)
     }
+
+    func testBackpressue() {
+        class BackpressureResponseDelegate: HTTPClientResponseDelegate {
+            typealias Response = Void
+            var count = 0
+            var processingBodyPart = false
+            var didntWait = false
+            var lock = Lock()
+
+            init() { }
+
+            func didReceiveHead(task: HTTPClient.Task<Response>, _ head: HTTPResponseHead) -> EventLoopFuture<Void> {
+                return task.eventLoop.makeSucceededFuture(())
+            }
+
+            func didReceiveBodyPart(task: HTTPClient.Task<Response>, _ part: ByteBuffer) -> EventLoopFuture<Void> {
+                lock.withLock {
+                    // if processingBodyPart is true then previous body part is still being processed
+                    // XCTAssertEqual doesn't work here so store result to test later
+                    if processingBodyPart == true {
+                        didntWait = true
+                    }
+                    processingBodyPart = true
+                    count += 1
+                }
+                // wait one second before returning a successful future
+                return task.eventLoop.scheduleTask(in: .milliseconds(1000) ) {
+                    self.lock.withLock {
+                        self.processingBodyPart = false
+                        self.count -= 1
+                    }
+                }.futureResult
+            }
+
+            func didReceiveError(task: HTTPClient.Task<Response>, _ error: Error) { }
+            func didFinishRequest(task: HTTPClient.Task<Response>) throws { }
+        }
+
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 5)
+        let client = HTTPClient(eventLoopGroupProvider: .shared(elg))
+        defer {
+            XCTAssertNoThrow(try client.syncShutdown())
+            XCTAssertNoThrow(try elg.syncShutdownGracefully())
+        }
+
+        let data = Data(count: 65273)
+        let backpressureResponseDelegate = BackpressureResponseDelegate()
+        guard let request = try? HTTPClient.Request(url: self.defaultHTTPBinURLPrefix + "get", body: .data(data)) else {
+            XCTFail("Failed to init Request")
+            return
+        }
+        XCTAssertNoThrow(try client.execute(request: request, delegate: backpressureResponseDelegate).wait())
+        XCTAssertEqual(backpressureResponseDelegate.didntWait, false)
+        XCTAssertEqual(backpressureResponseDelegate.count, 0)
+    }
 }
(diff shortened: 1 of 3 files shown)


Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
         XCTAssertEqual(info.connectionNumber, 1)
         XCTAssertEqual(info.requestNumber, 1)
     }
+
+    func testBackpressue() {
```
Lukasa, line 2537: Nit:

```suggestion
    func testBackpressure() {
```

adam-fowler:
> Hey @adam-fowler, I'm about to do a full review but I wanted to let you know early that you'll need to run the test generation script to get the CI working.

sorry Cory forgot about that. Sorted now

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
...
+                    processingBodyPart = true
+                    count += 1
+                }
+                // wait one second before returning a successful future
+                return task.eventLoop.scheduleTask(in: .milliseconds(1000) ) {
+                    self.lock.withLock {
+                        self.processingBodyPart = false
+                        self.count -= 1
+                    }
+                }.futureResult
```
Lukasa, line 2567: We could bump the dependency version of NIO to 2.17.0 and use `flatScheduleTask` here instead.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
...
+                    // if processingBodyPart is true then previous body part is still being processed
+                    // XCTAssertEqual doesn't work here so store result to test later
+                    if processingBodyPart == true {
+                        didntWait = true
+                    }
+                    processingBodyPart = true
+                    count += 1
+                }
+                // wait one second before returning a successful future
+                return task.eventLoop.scheduleTask(in: .milliseconds(1000) ) {
```
Lukasa: Waiting a second does force this test to take quite a long time. Is there any reason we think this test wouldn't work equally well with a delay of, say, 200ms instead of the full second?

Lukasa:
> > Hey @adam-fowler, I'm about to do a full review but I wanted to let you know early that you'll need to run the test generation script to get the CI working.
> 
> sorry Cory forgot about that. Sorted now

No need to apologise! Easily done. :grin:

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: What does `flatScheduleTask` gives us here. All we want to do is set a variable.

Lukasa: Oh yup, my bad, misread what was going on here.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: oh dear

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: Reduced to 200ms

Lukasa:
Nice, looks exactly right. Test failures seem to back up #274 as well, so this is a good leaping off point for providing a fix.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
...
+                        self.count -= 1
+                    }
+                }.futureResult
+            }
+
+            func didReceiveError(task: HTTPClient.Task<Response>, _ error: Error) {}
+            func didFinishRequest(task: HTTPClient.Task<Response>) throws {}
+        }
+
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 5)
```
weissi, line 2574: can't we use the `self.defaultHTTPClient`?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
         XCTAssertEqual(info.connectionNumber, 1)
         XCTAssertEqual(info.requestNumber, 1)
     }
+
+    func testBackpressure() {
```
weissi, line 2537: I think we should maybe name it `testDownloadBackpressure`? That'd match the [existing test](https://github.com/swift-server/async-http-client/blob/aac4357b655231e2d31a9e24fc98756f95c73fd5/Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift#L232).

Asl

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: Then we don't need a group or a client

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
...
+            init() {}
+
+            func didReceiveHead(task: HTTPClient.Task<Response>, _ head: HTTPResponseHead) -> EventLoopFuture<Void> {
+                return task.eventLoop.makeSucceededFuture(())
+            }
+
+            func didReceiveBodyPart(task: HTTPClient.Task<Response>, _ part: ByteBuffer) -> EventLoopFuture<Void> {
+                self.lock.withLock {
+                    // if processingBodyPart is true then previous body part is still being processed
+                    // XCTAssertEqual doesn't work here so store result to test later
```
weissi, line 2554: out of interest: What about `XCTAssertEqual` doesn't work here?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
...
+            var didntWait = false
+            var lock = Lock()
+
+            init() {}
+
+            func didReceiveHead(task: HTTPClient.Task<Response>, _ head: HTTPResponseHead) -> EventLoopFuture<Void> {
+                return task.eventLoop.makeSucceededFuture(())
+            }
+
+            func didReceiveBodyPart(task: HTTPClient.Task<Response>, _ part: ByteBuffer) -> EventLoopFuture<Void> {
```
weissi, line 2551: how are we making sure that `didReceiveBodyPart` is actually invoked multiple times? Because if it's only invoked once, then we don't know if download streaming works or not I think?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2533,4 +2533,59 @@ class HTTPClientTests: XCTestCase {
...
+        }
+
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 5)
+        let client = HTTPClient(eventLoopGroupProvider: .shared(elg))
+        defer {
+            XCTAssertNoThrow(try client.syncShutdown())
+            XCTAssertNoThrow(try elg.syncShutdownGracefully())
+        }
+
+        let data = Data(count: 65273)
```
weissi, line 2581: out of interest: Why are we uploading data here? Aren't we testing download streaming?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: XCTAssertEqual doesn’t return an error while assert does not totally sure why

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: We aren’t at the moment. I can add a test to see if it did. Is there a way we can force it?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: Easiest way to get it to download a sizeable chunk of data was upload data to HTTPBin. HTTPBin just returns it back.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: Ok

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: I found a client that uses an elg with more than one thread is more likely to fail

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: If I replace this with an HTTPBin entry
```
            case "/chunks":
                let data = Data(count: 8000)
                var buf = context.channel.allocator.buffer(capacity: data.count)
                buf.writeBytes(data)
                context.write(wrapOutboundOut(.head(HTTPResponseHead(version: HTTPVersion(major: 1, minor: 1), status: .ok))), promise: nil)
                context.writeAndFlush(wrapOutboundOut(.body(.byteBuffer(buf))), promise: nil)
                context.writeAndFlush(wrapOutboundOut(.body(.byteBuffer(buf))), promise: nil)
                context.writeAndFlush(wrapOutboundOut(.body(.byteBuffer(buf))), promise: nil)
                context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
                return
```
would that work also to ensure that the number of body parts is greater than 1?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: I think we should get #177 in for that tbh. But your call.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: @adam-fowler yes, it continues the execution but it will still fail the test run once it's complete. So I think you can just use `XCTAssert...`

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: @adam-fowler From the NIO level: Yes, you could set `ChannelOptions.maxMessagesPerRead` to `1` and set the allocator to `FixedSizeRecvAllocator(1)` which means that NIO will read everything byte to byte. That'll cause AHC to call out multiple times.

The issue here is that through AHC's API, you can't easily mess with the underlying `Channel`... You could make it an internal test and force it that way.

The other (probably easier) option is:
- Download say 1 MB of data. NIO's default allocator will never send you more than 64k. Given that AHC doesn't change it, you can't get 1 MB in one big chunk.
- Add `XCTAssertGreaterOrEqual(numberOfCalls, 2)` or so.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: I see, that makes sense, thank you. I would probably create a route that returns a fixed amount of bytes.

Maybe something like `/send-zeroes/1000000` would send 1 million zeroes. Then you can just use

```
context.channel.allocator.buffer(repeating: 0, count: count)
```

And btw, the faster and shorter way of writing

```swift
                let data = Data(count: 8000)
                var buf = context.channel.allocator.buffer(capacity: data.count)
                buf.writeBytes(data)
```

is

```swift
let buf = context.channel.allocator.buffer(repeating: 0, count: 8000)
```

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: Until #177 is merged I leave this using the local client

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: In this situation it doesn't work.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
adam-fowler: Set it 100000, because if the test is successful it will take quite a while if you set the download size too high.

weissi: @adam-fowler Ok, 150,000 should be enough, barring AHC changes, the max is 64k.

ktoso:

adam-fowler:
> Just as a heads up, the main development branch has been changed to `main`, following the [Swift policy](https://forums.swift.org/t/moving-default-branch-to-main/38515/2) on this.
> 
> This PR has been re-targeted to main and should just work. However when performing rebases etc please keep this in mind -- you may want to fetch the main branch and rebase onto the `main` branch from now on since master is not up-to-date anymore and is going to be deleted shortly.

All for this, but the default branch is still `master` here and recent PRs have been merged into `master` and not `main`

adam-fowler:
@Lukasa This looks like you fixed this and added tests in #352. Shall I close

Lukasa:
Sure thing.

PR #275, opened 2020-06-24
Add FileDownloadDelegate for simple file downloads

MaxDesiatov:
A simple test is included, let me know if you'd like it to be more sophisticated.

### DIFF
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -50,13 +50,15 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
         reportHeaders: ((HTTPHeaders) -> Void)? = nil,
         reportProgress: ((Progress) -> Void)? = nil
     ) throws {
-        self.handle = try self.io.openFile(
-            path: path,
-            mode: .write,
-            flags: .allowFileCreation()
-        )
         pool.start()
         self.io = NonBlockingFileIO(threadPool: pool)
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        self.handle = self.io.openFile(
+            path: path,
+            mode: .write,
+            flags: .allowFileCreation(),
+            eventLoop: eventLoopGroup.next()
+        )
 
         self.reportHeaders = reportHeaders
         self.reportProgress = reportProgress
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -15,7 +15,10 @@
 import NIO
 import NIOHTTP1
 
+/// Handles a streaming download to a given file path, allowing headers and progress to be reported.
 public final class FileDownloadDelegate: HTTPClientResponseDelegate {
+    /// The response type for this delegate: the total count of bytes as reported by the response
+    /// "Content-Length" header (if available) and the count of bytes downloaded.
     public typealias Response = (totalBytes: Int?, receivedBytes: Int)
 
     private var totalBytes: Int?
@@ -28,6 +31,13 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
 
     private var writeFuture: EventLoopFuture<Void>?
 
+    /// Initializes a new file download delegate.
+    /// - parameters:
+    ///     - path: Path to a file you'd like to write the download to.
+    ///     - pool: A thread pool to use for asynchronous file I/O.
+    ///     - reportHeaders: A closure called when the response headers are available.
+    ///     - reportProgress: A closure called when a body chunk has been downloaded, with
+    ///       the total byte count and download byte count passed to it as arguments.
     public init(
         path: String,
         pool: NIOThreadPool = NIOThreadPool(numberOfThreads: 1),
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -116,7 +116,7 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
     }
 
     public func didFinishRequest(task: HTTPClient.Task<Response>) throws -> Response {
-        if let writeFuture = self.fileHandleFuture {
+        if let writeFuture = self.writeFuture {
             writeFuture.whenComplete { _ in
                 self.fileHandleFuture?.whenSuccess(self.close(fileHandle:))
                 self.writeFuture = nil
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -19,15 +19,19 @@ import NIOHTTP1
 public final class FileDownloadDelegate: HTTPClientResponseDelegate {
     /// The response type for this delegate: the total count of bytes as reported by the response
     /// "Content-Length" header (if available) and the count of bytes downloaded.
-    public typealias Response = (totalBytes: Int?, receivedBytes: Int)
+    public struct Progress {
+        public var totalBytes: Int?
+        public var receivedBytes: Int
+    }
 
-    private var totalBytes: Int?
-    private var receivedBytes = 0
+    private var progress = Progress(totalBytes: nil, receivedBytes: 0)
 
-    private let handle: NIOFileHandle
+    public typealias Response = Progress
+
+    private let handle: EventLoopFuture<NIOFileHandle>
     private let io: NonBlockingFileIO
     private let reportHeaders: ((HTTPHeaders) -> Void)?
-    private let reportProgress: ((_ totalBytes: Int?, _ receivedBytes: Int) -> Void)?
+    private let reportProgress: ((Progress) -> Void)?
 
     private var writeFuture: EventLoopFuture<Void>?
 
@@ -44,9 +48,13 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
         path: String,
         pool: NIOThreadPool = NIOThreadPool(numberOfThreads: 1),
         reportHeaders: ((HTTPHeaders) -> Void)? = nil,
-        reportProgress: ((_ totalBytes: Int?, _ receivedBytes: Int) -> Void)? = nil
+        reportProgress: ((Progress) -> Void)? = nil
     ) throws {
-        self.handle = try NIOFileHandle(path: path, mode: .write, flags: .allowFileCreation())
+        self.handle = try self.io.openFile(
+            path: path,
+            mode: .write,
+            flags: .allowFileCreation()
+        )
         pool.start()
         self.io = NonBlockingFileIO(threadPool: pool)
 
@@ -62,7 +70,7 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
 
         if let totalBytesString = head.headers.first(name: "Content-Length"),
             let totalBytes = Int(totalBytesString) {
-            self.totalBytes = totalBytes
+            self.progress.totalBytes = totalBytes
         }
 
         return task.eventLoop.makeSucceededFuture(())
@@ -72,19 +80,26 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
         task: HTTPClient.Task<Response>,
         _ buffer: ByteBuffer
     ) -> EventLoopFuture<Void> {
-        self.receivedBytes += buffer.readableBytes
-        self.reportProgress?(self.totalBytes, self.receivedBytes)
+        self.progress.receivedBytes += buffer.readableBytes
+        self.reportProgress?(self.progress)
 
-        let writeFuture = self.io.write(fileHandle: self.handle, buffer: buffer, eventLoop: task.eventLoop)
+        let writeFuture = self.handle.flatMap {
+            self.io.write(fileHandle: $0, buffer: buffer, eventLoop: task.eventLoop)
+        }
+
         self.writeFuture = writeFuture
         return writeFuture
     }
 
     public func didFinishRequest(task: HTTPClient.Task<Response>) throws -> Response {
-        self.writeFuture?.whenComplete { _ in
-            try! self.handle.close()
-            self.writeFuture = nil
+        if let writeFuture = self.writeFuture {
+            writeFuture.whenComplete { _ in
+                self.handle.whenSuccess { try! $0.close() }
+                self.writeFuture = nil
+            }
+        } else {
+            self.handle.whenSuccess { try! $0.close() }
         }
-        return (self.totalBytes, self.receivedBytes)
+        return self.progress
     }
 }
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -91,7 +91,7 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
         let writeFuture: EventLoopFuture<Void>
         if let fileHandleFuture = self.fileHandleFuture {
             writeFuture = fileHandleFuture.flatMap {
-                return self.write(buffer: buffer, to: $0, eventLoop: task.eventLoop)
+                self.write(buffer: buffer, to: $0, eventLoop: task.eventLoop)
             }
         } else {
             let fileHandleFuture = self.io.openFile(
@@ -102,7 +102,7 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
             )
             self.fileHandleFuture = fileHandleFuture
             writeFuture = fileHandleFuture.flatMap {
-                return self.write(buffer: buffer, to: $0, eventLoop: task.eventLoop)
+                self.write(buffer: buffer, to: $0, eventLoop: task.eventLoop)
             }
         }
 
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -0,0 +1,88 @@
...
+    ///     - reportProgress: A closure called when a body chunk has been downloaded, with
+    ///       the total byte count and download byte count passed to it as arguments.
+    public init(
+        path: String,
+        pool: NIOThreadPool = NIOThreadPool(numberOfThreads: 1),
+        reportHeaders: ((HTTPHeaders) -> Void)? = nil,
+        reportProgress: ((_ totalBytes: Int?, _ receivedBytes: Int) -> Void)? = nil
+    ) throws {
+        self.handle = try NIOFileHandle(path: path, mode: .write, flags: .allowFileCreation())
+        pool.start()
```
Lukasa, line 48: This doesn't work at the moment, but it should. apple/swift-nio#1572.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -0,0 +1,88 @@
...
+    ///     - path: Path to a file you'd like to write the download to.
+    ///     - pool: A thread pool to use for asynchronous file I/O.
+    ///     - reportHeaders: A closure called when the response headers are available.
+    ///     - reportProgress: A closure called when a body chunk has been downloaded, with
+    ///       the total byte count and download byte count passed to it as arguments.
+    public init(
+        path: String,
+        pool: NIOThreadPool = NIOThreadPool(numberOfThreads: 1),
+        reportHeaders: ((HTTPHeaders) -> Void)? = nil,
+        reportProgress: ((_ totalBytes: Int?, _ receivedBytes: Int) -> Void)? = nil
```
Lukasa, line 45: Can we try to establish certainty about what threading context will be used to invoke these callbacks? It's good to write these things down in documentation.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -0,0 +1,88 @@
...
+        self.receivedBytes += buffer.readableBytes
+        self.reportProgress?(self.totalBytes, self.receivedBytes)
+
+        let writeFuture = self.io.write(fileHandle: self.handle, buffer: buffer, eventLoop: task.eventLoop)
+        self.writeFuture = writeFuture
+        return writeFuture
+    }
+
+    public func didFinishRequest(task: HTTPClient.Task<Response>) throws -> Response {
+        self.writeFuture?.whenComplete { [weak self] _ in
```
Lukasa, line 82: Capturing `self` weakly here is an anti-pattern: if the user drops the `FileDownloadDelegate` such that it is deinited, we will never close the handle and so we will leak it. We must capture self strongly here.

More generally, there is little risk when capturing self strongly in one-shot closures. The reason to use a weak capture is when we are passing self to something that we ourselves hold. Future callbacks, like dispatch work items, are transient: when they have been invoked they no longer exist and are freed. It's almost always safe to capture self strongly in those contexts.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -488,6 +488,29 @@ class HTTPClientTests: XCTestCase {
         XCTAssertEqual(10, count)
     }
 
+    func testFileDownload() throws {
+        var request = try Request(url: self.defaultHTTPBinURLPrefix + "events/10/1")
+        request.headers.add(name: "Accept", value: "text/event-stream")
+
+        let (totalBytes, receivedBytes) =
+            try TemporaryFileHelpers.withTemporaryUnixDomainSocketPathName { path -> (Int?, Int) in
```
Lukasa, line 496: This is a bit of a nit, but I'd like us to rename this function. I'm fine with having `withTemporaryUnixDomainSocketPathName` call into an underlying function, but I think it's confusing to claim we're going to use a file as a unix socket and then to use it for something else altogether.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
MaxDesiatov: I am aware of this reasoning, the problem is that the future itself references this closure, which captures `self`. There is no need for `[weak self]` if this is a one-shot closure, but since it's owned by `self` it creates a circular reference: `self` -> `writeFuture` -> closure -> `self`.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
MaxDesiatov: Would you like me to add a comment that clarifies that the pool shouldn't be started before being passed to the initializer, or should I just wait for that PR to be merged first?

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
MaxDesiatov: I don't think there's any documentation whatsoever on the threading context of `HTTPClientResponseDelegate` itself, can I even make any assumptions in absence of that?

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
Lukasa: Strong reference cycles are not a problem so long as they are deterministically broken. In this case, this future callback _must_ complete: it is a program error for it to fail to do so. Thus, this closure will be executed, and once it's executed the future will drop its reference to it and the cycle will break.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
Lukasa: I think waiting for that PR to be merged and bumping the minimum NIO version is fine. We're aiming to get a NIO release out this week so we shouldn't be waiting long.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
Lukasa: You can say that the callbacks will be invoked in the same threading context that the delegate itself is invoked, as controlled by `EventLoopPreference`. That is covered elsewhere in the docs.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
MaxDesiatov: Cool, makes sense. I didn't see this behavior documented anywhere, either for `EventLoopFuture` specifically, or in general in related NIO docs. I had enough experience with both 3rd-party and system libraries that don't follow this behavior and also don't document it in any way, so I'm used to staying on the safe side.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
MaxDesiatov: Would you mind pointing what this "elsewhere" is? Unfortunately, the Jazzy docs are only searchable for exact symbol names, none of my search queries for threading contexts led to anything.

Lukasa: Sure, it's documented on [these fields](https://github.com/swift-server/async-http-client/blob/63fa13ce706f16c9e374f454f59e1c5c26d4598f/Sources/AsyncHTTPClient/HTTPClient.swift#L764-L782).

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
Lukasa: apple/swift-nio#1575.

MaxDesiatov: That's great, thanks!

MaxDesiatov:
As a quick note, I'd like to get more opinions on whether the feature itself should be included before I address the technical feedback provided by Cory.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
weissi: I think we shouldn't start the thread pool here. If the user hands you a thread pool, they're responsible for starting & stopping it.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -0,0 +1,88 @@
...
+            try! self?.handle.close()
```
artemredkin, line 83: I don't think we should crash if the `close` fails

MaxDesiatov: I've made it `try!` based on this suggestion from @weissi in the Swift Server Slack (`#async-http-client` channel):

>and you want to `try!` the `close.` The only reason for the `close` to fail is that you have already closed it. If you have already closed it, then that's a serious programmer error that need fixing.

MaxDesiatov:
Hi @artemredkin, do you think that this feature makes sense overall in AsyncHTTPClient? Just wondering if it's worth updating the PR to address the feedback, otherwise I'd close it.

artemredkin:
@MaxDesiatov Hi! Sorry I missed your question. I do think this feature would be useful

ktoso:

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
MaxDesiatov: That would be great if the `pool` parameter didn't have a default value, the great thing about it though that we don't have to expose pools to our users if they don't want it. Given that apple/swift-nio#1572 was merged, would you mind if the pool is still started here?

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -0,0 +1,90 @@
...
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOHTTP1
+
+/// Handles a streaming download to a given file path, allowing headers and progress to be reported.
+public final class FileDownloadDelegate: HTTPClientResponseDelegate {
+    /// The response type for this delegate: the total count of bytes as reported by the response
+    /// "Content-Length" header (if available) and the count of bytes downloaded.
+    public typealias Response = (totalBytes: Int?, receivedBytes: Int)
```
Lukasa, line 22: Should we make this a struct? Making this a nominal type now makes life a bit easier down the road if we want protocol conformances or any other helpers.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -0,0 +1,90 @@
...
+    ///       the total byte count and download byte count passed to it as arguments. The callbacks
+    ///       will be invoked in the same threading context that the delegate itself is invoked,
+    ///       as controlled by `EventLoopPreference`.
+    public init(
+        path: String,
+        pool: NIOThreadPool = NIOThreadPool(numberOfThreads: 1),
+        reportHeaders: ((HTTPHeaders) -> Void)? = nil,
+        reportProgress: ((_ totalBytes: Int?, _ receivedBytes: Int) -> Void)? = nil
+    ) throws {
+        self.handle = try NIOFileHandle(path: path, mode: .write, flags: .allowFileCreation())
```
Lukasa, line 49: I know this is annoying, but `open` is technically a blocking operation. We should consider making the handle available by a future using `self.io.openFile(path:mode:flags:)`.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -814,9 +814,8 @@ extension TaskHandler: ChannelDuplexHandler {
         do {
             try headers.validate(method: request.method, body: request.body)
         } catch {
+            self.errorCaught(context: context, error: error)
```
Lukasa, line 817: What's the purpose of this change?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -1038,14 +1034,20 @@ extension TaskHandler: ChannelDuplexHandler {
...
+                // error was already handled
+                break
+            }
         }
     }
 
     func handlerAdded(context: ChannelHandlerContext) {
         guard context.channel.isActive else {
-            self.failTaskAndNotifyDelegate(error: HTTPClientError.remoteConnectionClosed, self.delegate.didReceiveError)
+            self.errorCaught(context: context, error: HTTPClientError.remoteConnectionClosed)
```
Lukasa, line 1050: There are a bunch of changes in this file that don't seem well motivated: what are they there for?

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
```diff
@@ -1080,4 +1080,87 @@ class HTTPClientInternalTests: XCTestCase {
...
+extension TaskHandler.State {
+    var isEndOrError: Bool {
+        switch self {
+        case .endOrError:
+            return true
+        default:
+            return false
+        }
+    }
 }
```
Lukasa, line 1166: Similarly these changes here.

Lukasa (changes requested):

Sources/AsyncHTTPClient/HTTPHandler.swift
MaxDesiatov: These were cause afer the base branch was changed from `master` to `main`, sorry about that, I'll clean it up.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
MaxDesiatov: Thanks, I've also created https://github.com/apple/swift-nio/pull/1630 to make this more clear in `NIOFaileHandle` doc comments.

MaxDesiatov: @Lukasa BTW, there's no such thing as `openFile(path:mode:flags:)`, there's only `openFile(path:mode:flags:eventLoop:)`, where would you like the event loop in the initializer to come from?

Sources/AsyncHTTPClient/HTTPHandler.swift
MaxDesiatov: I'm changing the base branch back from `main` to `master`, because `main` is outdated compared to `master` in this repo.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
Lukasa: It'll need to be the event loop that the delegate callbacks fire on.

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: @ktoso What happened with the migration to `main` here?

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
MaxDesiatov: The event loop for delegate callbacks is passed to every callback separately as an argument, do you think there's a way to get access to that event loop in the initializer? Whoever creates the delegate doesn't have access to that event loop too, it's internal and owned by the `HTTPClient` instance, to which the delegate doesn't have any direct access to (or even an indirect one, as far as I understand).

Lukasa: Yeah, it's possible that `open` may have to be delayed until the first operation on the file occurs.

MaxDesiatov: That's a great idea, I've implemented it and it's now ready for review.

Sources/AsyncHTTPClient/HTTPHandler.swift
ktoso: Something's wrong... I have it noted that we completed the work here but seems master still was the main thing, wat?

Please do not target `master` @MaxDesiatov, please keep targeting main.
I'll bring main up to date again and re-do the steps about branch protection.
CI is also ticked off to have been changed but I can't confirm that since no admin rights there.

ktoso: It seems we/I missed to change the default branch so a few PRs were made against master and merged. I've brought main up to date again, fixed that, re targeted this at main and _removed_ master.

Please keep using main from now-on, it should be exactly the same as master just was (it was behind a few commits indeed).

MaxDesiatov: Thanks, I've rebased it on top of `main`. @Lukasa now ready for review.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -0,0 +1,121 @@
...
+        self.writeFuture = writeFuture
+        return writeFuture
+    }
+
+    private func close(fileHandle: NIOFileHandle) {
+        try! fileHandle.close()
+        self.fileHandleFuture = nil
+    }
+
+    public func didFinishRequest(task: HTTPClient.Task<Response>) throws -> Response {
```
Lukasa, line 110: Can we add a test on the error path? I think we're missing cleanup in the `didReceiveError` case.

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -531,7 +552,8 @@ internal final class HttpBinHandler: ChannelInboundHandler {
                 context.writeAndFlush(wrapOutboundOut(.head(HTTPResponseHead(version: HTTPVersion(major: 1, minor: 1), status: .ok))), promise: nil)
                 return
             case "/events/10/1": // TODO: parse path
-                context.write(wrapOutboundOut(.head(HTTPResponseHead(version: HTTPVersion(major: 1, minor: 1), status: .ok))), promise: nil)
+                let headers = HTTPHeaders([("Content-Length", "50")])
```
Lukasa, line 555: This changes the behaviour of this endpoint (it previously automatically sent chunked encoding, now it doesn't). I don't _think_ any of the tests rely on that behaviour, but as it's fairly cheap to add new endpoints I think I'd rather we added a new one of these.

We can keep the common code by factoring this out into a method that conditionally adds content-length if we want.

Lukasa (commented):
Cool, I think this looks really good! I've left a few notes in the diff.

Lukasa (approved):
Very nice, I like it a lot!

Issue #229, opened 2020-05-28
Should we combine .host and .socketPath on HTTPClient.Request?


dimitribouniol:
While refactoring the unix domain socket implementation in #228, I noticed that we could probably combine the `host` and (new) `socketPath` properties on `HTTPClient.Request`.

From what I could tell:
- The socket path would end up in the Host header for the request (currently it is `""`)
- TLS would use the socket path on certs as the common name/hostname (currently it is checking for `""` as opposed to `nil`, which it's doing for IPs)
    - IPs are explicitly forbidden from the [SNI spec](https://www.ietf.org/rfc/rfc4366.html#page-10). It says:
> Literal IPv4 and IPv6 addresses are not permitted in "HostName".

- - The same spec does not forbid using socket paths however:
> Currently, the only server names supported are DNS hostnames; however, this does not imply any dependency of TLS on DNS, and other name types may be added in the future (by an RFC that updates this document).  TLS MAY treat provided server names as opaque data and pass the names and types to the application.

- - However, we should still check the common name in cert validation for matching hostnames
- Responses would propagate the socket path as their `host` as well (currently, it's also set to `""` for socket domain responses)
- We would need better test coverage for these scenarios.

Lukasa:
I don't think trying to validate TLS certificates with the socket path in them is likely to work well: so far as I know, nobody does that. The same applies to SNI, given that again, socket paths are essentially never used there. It's possible that the right thing to do for those two cases is to allow an override, such that users can say what Host they want to treat the request as being for, irregardless of what the URL actually says.

curl does this with `--resolve`, which works nicely.

Lukasa:
Actually, in general I wonder if that's not the correct solution to all of this. Essentially, to say that there are some subset of hostnames that are "fake", and resolve to some other underlying socket address. Then the client will essentially map the request to that other socket address, but for things like SNI, cert verification, and other purposes we'll behave as though the host was the one provided directly to us.

In fact, this might be something we can do with apple/swift-nio#1487.

dimitribouniol:
I agree that being able to pass in a custom hostname in any scenario is probably the better way to go about this. If one is not provided though, should it fall back to the socket path? Is it ok to make a request with a Host that is just `""`?

Lukasa:
Yeah, I don't think it's worth passing the socket path as `Host`. I don't think any server will ever expect it.

dimitribouniol:
If I were to add this to `Configuration`, I can do it in two ways:
[...]

PR #277, opened 2020-06-27
Redirect config per request

### LINKED ISSUE #196: redirect configuration is only available top-level on HTTPClient
Whether (and how many times) you want to redirect is usually a setting that you want to be able to set per request and not globally per HTTPClient. Unfortunately, from what I can tell right now, we only allow setting this setting globally which is not practical given the fact that many client need to choose how they handle redirects, yet a single `HTTPClient` should be widely shared across the application.

ncke:
Allows redirect handling to be configured per request.

Motivation:

Redirect configuration is set globally when configuring the `HTTPClient`. However, it would also be useful to configure redirects on a per request basis. This would mean that a single `HTTPClient` could be configured and shared widely throughout an application, while retaining the flexibility to tailor specific requests with custom redirect handling should the need arise.

Modifications:

This change adds an optional `redirectConfiguration` parameter to all request methods of `HTTPClient`. When the request is executed the `prevailingConfiguration` is determined using the `redirectConfiguration` argument, if set. Otherwise the global configuration is used.

Result:

Allows redirect handling to be configured on a per request basis, closes #196


Lukasa:

Lukasa:
Before we go down this road, we should take this opportunity to ask ourselves whether there is any other configuration that can sensibly be provided "per call". If there is, we may want to whack it all in at once.

Lukasa:
I think I've talked myself into saying that redirect is special here. Most of the others are either managed by the connection pool or cannot safely be changed without invalidating the pool, or are otherwise already handled. So I think I'm ok with the shape of this change. @artemredkin any preference for anything else here?

Lukasa:
Oh, one note though @ncke: we can't add optional parameters to these methods, it's a semver major change to do it. We need to add new methods with extra arguments instead.

ncke:
I'll add new methods rather than the arguments.

I was also wondering whether the `Task` might be a better place to store redirect state (e.g. count) rather than the `Request`. I'm concerned that replaying an executed (or executing) `Request` may copy the dirty redirect state rather than creating pristine state for the second execution, with unexpected results.

Lukasa:
`Request` is a value type: mutations made to that field by `HTTPClient` are not reflected in your local copy.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -221,13 +221,26 @@ public class HTTPClient {
...
-        return self.get(url: url, deadline: deadline, logger: HTTPClient.loggingDisabled)
+        return self.get(url: url, deadline: deadline, logger: HTTPClient.loggingDisabled, redirectConfiguration: nil)
+    }
+
+    /// Execute `GET` request using specified URL.
+    ///
+    /// - parameters:
+    ///     - url: Remote URL.
+    ///     - deadline: Point in time by which the request must complete.
+    ///     - redirectConfiguration: The configuration of redirect handling for this request.
```
Lukasa, line 241: Can we update these docstrings to explain what `nil` means? This applies to all new methods.

ncke: Sure. Can we also change the argument name from `redirectConfiguration` to `redirects`? I think there was an earlier suggestion that this is more pithy and I like the idea.

Lukasa: No objection from me.

dimitribouniol:
> I think I've talked myself into saying that redirect is special here. Most of the others are either managed by the connection pool or cannot safely be changed without invalidating the pool, or are otherwise already handled. So I think I'm ok with the shape of this change. @artemredkin any preference for anything else here?

Perhaps we can have a configuration type specific to requests (ie. `execute(..., configuration: RequestConfiguration? = nil)`)? One more option that would be configurable per request would be the ability to specify how Host is represented, as brought up here: https://github.com/swift-server/async-http-client/issues/229#issuecomment-639001947, and having a configuration type we can extend and add to could be useful here.

dimitribouniol:
> Oh, one note though @ncke: we can't add optional parameters to these methods, it's a semver major change to do it. We need to add new methods with extra arguments instead.

@Lukasa Most of these methods (ie. the ones that add logging and conveniences for socket paths) are new and haven't actually been released — would modifying those be ok?

ncke:
@dimitribouniol With regard to simplifying future extension I wondered about retaining the current `httpClient.get(...)` style for simple cases, but contriving to allow `httpClient.with(configuration: ...).get(...)` to allow the caller to chain modifications to the execution context before reaching the final verb?

Lukasa:
I definitely like the idea of having request configuration. I don't think I love the chaining idea though: I don't think it adds meaningful API clarity. Having a struct you can pass that defines a per-request configuration overload seems sensible though.

As for changing methods, anything that hasn't shipped in an existing release can be changed, yeah.

ktoso:

PR #279, opened 2020-06-29
Added default value for queue in HTTPClient.shutdown(queue:callback:)

adam-fowler:
Default it to .global()


artemredkin:
yep, thanks @adam-fowler !

PR #282, opened 2020-07-16
Use SwiftLogNoOpLogHandler from swift-log instead of local version

adam-fowler:


Lukasa:
The biggest issue right now is that the master branch seems not to build due to warnings as errors. We'll need to update it to fix this build problem, one sec.

ktoso:
Thanks, Adam :)

PR #284, opened 2020-07-16
add a separate configuration struct for pool

artemredkin:
Move pool idle time configuration to separate struct.

Motivation:
Right now there is only one configuration parameter for the pool, idle time, but we might have more in the future. I think it makes sense to wrap all pool-related setting in a separate struct.

Modifications:
Add PoolConfiguration struct and move idle time to that struct

Result:
All pool level settings will be in one namespace.


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -641,8 +641,8 @@ public class HTTPClient {
         public var redirectConfiguration: RedirectConfiguration
         /// Default client timeout, defaults to no timeouts.
         public var timeout: Timeout
-        /// Timeout of pooled connections
-        public var maximumAllowedIdleTimeInConnectionPool: Optional<TimeAmount>
+        /// Connection pool configuration.
+        public var poolConfiguration: PoolConfiguration
```
Lukasa, line 645: We cannot remove fields in semver minors. The old field must continue to exist, though it's acceptable for it to be computed from the new field.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -653,14 +653,14 @@ public class HTTPClient {
         public init(tlsConfiguration: TLSConfiguration? = nil,
                     redirectConfiguration: RedirectConfiguration? = nil,
                     timeout: Timeout = Timeout(),
-                    maximumAllowedIdleTimeInConnectionPool: TimeAmount,
+                    poolConfiguration: PoolConfiguration = PoolConfiguration(),
```
Lukasa, line 656: Similarly, we cannot add or remove arguments in methods. We need a new `init`, and to change the old one to forward to this one.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -702,15 +702,15 @@ public class HTTPClient {
         public init(certificateVerification: CertificateVerification,
                     redirectConfiguration: RedirectConfiguration? = nil,
                     timeout: Timeout = Timeout(),
-                    maximumAllowedIdleTimeInConnectionPool: TimeAmount = .seconds(60),
+                    poolConfiguration: TimeAmount = .seconds(60),
```
Lukasa, line 705: Same here.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -860,6 +860,16 @@ extension HTTPClient.Configuration {
         /// - warning: Cycle detection will keep all visited URLs in memory which means a malicious server could use this as a denial-of-service vector.
         public static func follow(max: Int, allowCycles: Bool) -> RedirectConfiguration { return .init(configuration: .follow(max: max, allowCycles: allowCycles)) }
     }
+
+    /// Connection pool configuration.
+    public struct PoolConfiguration {
```
Lukasa, line 865: Can we make this `Hashable`?

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: That field was never in the released version, only in master

Lukasa: Good catch, you're right.

Sources/AsyncHTTPClient/HTTPClient.swift
artemredkin: Out of curiosity, why do you think we might need it? Also, `TimeAmount` is not hashable, so this will require implementing `hash(into:)`, should I add it?

Lukasa: `TimeAmount` is `Hashable` as of 2.19.0.

Lukasa: As to why we need it, it's not so much that we need it as that, for trivial structures, missing `Equatable` and `Hashable` conformances tend to be minor annoyances. It's nicer just to have them.

artemredkin: done!

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -860,6 +860,16 @@ extension HTTPClient.Configuration {
         /// - warning: Cycle detection will keep all visited URLs in memory which means a malicious server could use this as a denial-of-service vector.
         public static func follow(max: Int, allowCycles: Bool) -> RedirectConfiguration { return .init(configuration: .follow(max: max, allowCycles: allowCycles)) }
     }
+
+    /// Connection pool configuration.
+    public struct PoolConfiguration: Hashable {
+        // Specifies amount of time connections are kept idle in the pool.
+        var idleTimeout: TimeAmount
```
Lukasa, line 867: Whoops, nearly missed this.

```suggestion
        public var idleTimeout: TimeAmount
```

artemredkin: 🙏

artemredkin: fixed!

PR #288, opened 2020-07-30
rename connection pool configuration

### LINKED ISSUE #286: Redundant naming in `poolConfiguration`
https://github.com/swift-server/async-http-client/blob/2e6a64abb31fbc2d51da3919b0d8b35a6c7dc0b5/Sources/AsyncHTTPClient/HTTPClient.swift#L645

The new `poolConfiguration` property sits inside a configuration object so often, users will type 

```swift
HTTPClient(eventLoopGroupProvider: ...,
    configuration: Configuration(poolConfiguration: ...)
```

the word "configuration" appears way too much here. We can't fix all instances anymore but the worst example is probably `poolConfiguration` which isn't released yet so we should fix it. The other sibling properties also don't have the word "configuration". We could for example name it `connectionPoolParameters` or maybe just `connectionPool`?

I also think we shouldn't use the word "pool" without "connection" in a public API, users will probably wonder what is being pooled as we may be pooling other things (not just connections).

[I marked it release blocker for now because this can only be done before 1.2.0, feel free to remove if you think this is too much]

artemredkin:
Rename pool configuration type and property to `ConnectionPool`

Motivation:
`HTTPConfiguration.poolConfiguration` is redundant and does not make clear what pool are we talking about.

Modifications:
Rename property and type to `ConnectionPool`

Result:
Closes #286

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -353,7 +353,7 @@ class HTTP1ConnectionProvider {
         case .park(let connection):
             logger.trace("parking connection",
                          metadata: ["ahc-connection": "\(connection)"])
-            connection.setIdleTimeout(timeout: self.configuration.poolConfiguration.idleTimeout,
+            connection.setIdleTimeout(timeout: self.configuration.connectionPool.idleTimeout,
                                       logger: self.backgroundActivityLogger)
         case .closeProvider:
             logger.debug("closing provider",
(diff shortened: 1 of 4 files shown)


artemredkin:
Error:
```
ERROR:  Error installing jazzy:
	cocoapods-downloader requires Ruby version >= 2.3.3.
```

artemredkin:

artemredkin:
Same error in 5.1, @tomerd could you help with that?

tomerd:
@artemredkin I think we can do something like https://github.com/apple/swift-nio-ssl/pull/234/

PR #289, opened 2020-08-04
Add tracing instrumentation

slashmo:
Very much WIP.

- Closes slashmo/gsoc-swift-tracing#46
- Closes slashmo/gsoc-swift-tracing#95


slashmo:
FYI, this is how a request currently looks like in my example project:

```swift
InstrumentationSystem.bootstrap(FakeTracer())

let client = HTTPClient(eventLoopGroupProvider: .createNew)
var context = BaggageContext()
// Extracted from a prior request
context.fakeTraceID = "123456789"

var request = try HTTPClient.Request(url: "https://httpbin.org/anything", method: .POST, body: .string(#"{"hello":"World"}"#))
request.headers.add(name: "Content-Type", value: "application/json")

let response = try client
    .execute(request: request, context: context)
    .wait()
```

In the background, `FakeTracer` was automatically used to create (and end) a `Span`, with the following attributes being set:

```swift
[
  "http.request_content_length": TracingInstrumentation.SpanAttribute.int(17),
  "http.status_code": TracingInstrumentation.SpanAttribute.int(200),
  "http.target": TracingInstrumentation.SpanAttribute.string("/anything"), 
  "http.host": TracingInstrumentation.SpanAttribute.string("httpbin.org"), 
  "http.status_text": TracingInstrumentation.SpanAttribute.string("OK"),
  "http.scheme": TracingInstrumentation.SpanAttribute.string("https"),
  "http.response_content_length": TracingInstrumentation.SpanAttribute.int(433),
  "http.method": TracingInstrumentation.SpanAttribute.string("POST")
]
```

It also added a trace header to the request containing the `fakeTraceID`.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -225,115 +230,67 @@ public class HTTPClient {
...
-    /// Execute `GET` request using specified URL.
-    ///
-    /// - parameters:
-    ///     - url: Remote URL.
-    ///     - deadline: Point in time by which the request must complete.
-    ///     - logger: The logger to use for this request.
-    public func get(url: String, deadline: NIODeadline? = nil, logger: Logger) -> EventLoopFuture<Response> {
-        return self.execute(.GET, url: url, deadline: deadline, logger: logger)
+    public func get(url: String, context: BaggageContext, deadline: NIODeadline? = nil) -> EventLoopFuture<Response> {
+        return self.execute(.GET, url: url, context: context, deadline: deadline)
```
pokryfka, line 236: for backward compatibility, the context should be optional or have default (static? "none") value?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -461,32 +382,29 @@ public class HTTPClient {
...
-                                                              delegate: Delegate,
-                                                              eventLoop eventLoopPreference: EventLoopPreference,
-                                                              deadline: NIODeadline? = nil,
-                                                              logger originalLogger: Logger?) -> Task<Delegate.Response> {
-        let logger = (originalLogger ?? HTTPClient.loggingDisabled).attachingRequestInformation(request, requestID: globalRequestID.add(1))
+        // TODO: net.peer.ip / Not required, but recommended
+
+        var request = request
+        InstrumentationSystem.instrument.inject(context, into: &request.headers, using: HTTPHeadersInjector())
+
```
pokryfka, line 405: should be the (new) span context, no?

```swift
InstrumentationSystem.instrument.inject(span.context, into: &request.headers, using: HTTPHeadersInjector())
```

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -461,32 +382,29 @@ public class HTTPClient {
...
-    ///     - delegate: Delegate to process response parts.
-    ///     - eventLoop: NIO Event Loop preference.
-    ///     - deadline: Point in time by which the request must complete.
-    public func execute<Delegate: HTTPClientResponseDelegate>(request: Request,
-                                                              delegate: Delegate,
-                                                              eventLoop eventLoopPreference: EventLoopPreference,
-                                                              deadline: NIODeadline? = nil,
-                                                              logger originalLogger: Logger?) -> Task<Delegate.Response> {
-        let logger = (originalLogger ?? HTTPClient.loggingDisabled).attachingRequestInformation(request, requestID: globalRequestID.add(1))
+        // TODO: net.peer.ip / Not required, but recommended
```
pokryfka, line 401: @ktoso @slashmo could this is be propagated in baggage shared by tracers (??)

pokryfka:
Awesome 👍

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -590,7 +509,15 @@ public class HTTPClient {
...
             }
-        }.always { _ in
+        }
+        .and(task.futureResult)
+        .always { result in
+            if case let .success((_, response)) = result, let httpResponse = response as? HTTPClient.Response {
+                span.attributes.http.statusCode = Int(httpResponse.status.code)
+                span.attributes.http.statusText = httpResponse.status.reasonPhrase
+                span.attributes.http.responseContentLength = httpResponse.body?.readableBytes ?? 0
+            }
```
pokryfka, line 519: It should record error if it was failure

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -590,7 +509,15 @@ public class HTTPClient {
                 connection.release(closing: true, logger: logger)
                 return channel.eventLoop.makeFailedFuture(error)
             }
-        }.always { _ in
+        }
+        .and(task.futureResult)
+        .always { result in
+            if case let .success((_, response)) = result, let httpResponse = response as? HTTPClient.Response {
+                span.attributes.http.statusCode = Int(httpResponse.status.code)
```
pokryfka, line 516: other that the http attributes it should set Span Status depending on the httpResponse.status.code,

OTel provides a table hot to map one to the other: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#status

pokryfka:
@ktoso 

do you think its possible and makes sense to create `feature/instrumentation` integration branch in `async-http-client`?

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: No, contexts should never be defaulted to nothing/empty - that leads to fragile APIs where dropping the traces becomes the default operation, rather than doing the right thing -- keep passing along contexts. We want our APIs to help people do the right thing. Note in a same way, Go's context values must _never_ be defaulted to empty, we're in the same boat.

It's another discussion _how and when_ this PR will be merged though; I do strongly believe contexts _must not_ be optional, as that leads to breaking traces accidentally (been there, done that). See discussions in https://forums.swift.org/t/server-distributed-tracing/37464/10 and the guidelines in https://github.com/slashmo/gsoc-swift-baggage-context#argument-namingpositioning

ktoso:
> do you think its possible and makes sense to create feature/instrumentation integration branch in async-http-client?

Maybe an `experimental/tracing` branch could be a good idea indeed... so people can more earily try things out as we work on this, and _before_ it actually fully lands hm (esp since it has API impact).

WDYT @Lukasa @artemredkin @tomerd ?

Sources/AsyncHTTPClient/HTTPClient.swift
pokryfka: Makes sense. This will, however, affect integration in a big way.

Please see my comment about an integration branch below.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -225,115 +230,67 @@ public class HTTPClient {
...
-    /// Execute `DELETE` request using specified URL.
-    ///
-    /// - parameters:
-    ///     - url: Remote URL.
-    ///     - deadline: The time when the request must have been completed by.
-    ///     - logger: The logger to use for this request.
-    public func delete(url: String, deadline: NIODeadline? = nil, logger: Logger) -> EventLoopFuture<Response> {
-        return self.execute(.DELETE, url: url, deadline: deadline, logger: logger)
+    public func delete(url: String, context: BaggageContext, deadline: NIODeadline? = nil) -> EventLoopFuture<Response> {
+        return self.execute(.DELETE, url: url, context: context, deadline: deadline)
```
ktoso, line 279: that's an interesting question here: so... `deadline` was defaulted in the public API, but on execute it's not...
Technically our rules say "context last, before defaulted" so... Technically this would mean `self.execute(.DELETE, url: url, deadline: deadline, context: context)` I wonder if it matters / if it is worth it to be strict about this hm...

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: Sure it will, it breaks API, we're aware.

Sources/AsyncHTTPClient/HTTPClient.swift
pokryfka: note that its typically propagated in `X-Forwarded-For` header

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -461,32 +382,29 @@ public class HTTPClient {
...
-                            logger: HTTPClient.loggingDisabled)
-    }
+        var span = InstrumentationSystem.tracingInstrument.startSpan(named: request.method.rawValue, context: context, ofKind: .client, at: nil)
+        span.attributes.http.method = request.method.rawValue
+        span.attributes.http.scheme = request.scheme
+        span.attributes.http.target = request.uri
+        span.attributes.http.host = request.host
+        if let requestContentLength = request.body?.length {
+            span.attributes.http.requestContentLength = requestContentLength
+        }
```
pokryfka, line 400: there is `http.url` attribute in [OTel example](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#http-client-server-example)

> 

http.url | "https://example.com:8080/webshop/articles/4?s=1"
-- | --

Lukasa:
> Maybe an experimental/tracing branch could be a good idea indeed... so people can more earily try things out as we work on this, and before it actually fully lands hm (esp since it has API impact).

I'm fine with that. By the way, right now this patch doesn't just have API impact, it's API breaking. I'd like to find a way to avoid that if we can.

slashmo:
> I'd like to find a way to avoid that if we can.
> @Lukasa 

We briefly chatted about the reasoning of having it API breaking here: slashmo/gsoc-swift-baggage-context#8

Lukasa:
I don't think that conversation does a good job of making the case.

ktoso:
The topic is a bit deeper than "this breaks API so it's bad", it is _what is the end goal we want to head towards_, and if we agree on this, we can talk more how we get there. This was touched upon in various places but as Tanner just opened up https://forums.swift.org/t/the-context-passing-problem/39162 which also relates to this, perhaps we use that forums thread to discuss this again, and then decide how to roll changes into the client here -- sounds good?

Long story short: Yes, context parameter replaces logger parameter, because of the insistence of the "pass a logger" yet context is the actual object that has tracing metadata so so it is the actual thing that needs passing around more than a logger even, but since we also want to pass loggers, we need e.g. a composite context protocol, like designed here: https://github.com/slashmo/gsoc-swift-baggage-context/pull/9 - so we can pass one parameter, and not logger + context all the time. I believe it was very much known up front with Johannes's work and merging the logger parameter that there would have to be an API breaking change once tracing/logging is "done", I talked with Johannes about this at length back then.

As mentioned above, let's first continue in the thread that Tanner kicked off, agree there and loop back here, sounds good?

ktoso:
I added a do-not-merge marker, perhaps that was partially the confusion here as well? This PR is not intended to be merged as is.

Sources/AsyncHTTPClient/HTTPClient.swift
pokryfka: @ktoso @slashmo please have a look and let me know if I understand the context propagation in the same way

I think you should inject `span.context` not `context`

See also https://github.com/pokryfka/aws-xray-sdk-swift#context-propagation

Sources/AsyncHTTPClient/HTTPClient.swift
slashmo: Updated 👍

Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift
```diff
@@ -64,7 +64,7 @@ extension HTTPClientTests {
             ("testNoResponseWithIgnoreErrorForSSLUncleanShutdown", testNoResponseWithIgnoreErrorForSSLUncleanShutdown),
             ("testWrongContentLengthForSSLUncleanShutdown", testWrongContentLengthForSSLUncleanShutdown),
             ("testWrongContentLengthWithIgnoreErrorForSSLUncleanShutdown", testWrongContentLengthWithIgnoreErrorForSSLUncleanShutdown),
-            ("testEventLoopArgument", testEventLoopArgument),
+//            ("testEventLoopArgument", testEventLoopArgument),
```
slashmo, line 67:
```suggestion
// TODO: Comment back in once failure was resolved
//            ("testEventLoopArgument", testEventLoopArgument),
```

Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift
```diff
@@ -88,7 +88,7 @@ extension HTTPClientTests {
             ("testUncleanShutdownCancelsTasks", testUncleanShutdownCancelsTasks),
             ("testDoubleShutdown", testDoubleShutdown),
             ("testTaskFailsWhenClientIsShutdown", testTaskFailsWhenClientIsShutdown),
-            ("testRaceNewRequestsVsShutdown", testRaceNewRequestsVsShutdown),
+//            ("testRaceNewRequestsVsShutdown", testRaceNewRequestsVsShutdown),
```
slashmo, line 91:
```suggestion
// TODO: Comment back in once failure was resolved
//            ("testRaceNewRequestsVsShutdown", testRaceNewRequestsVsShutdown),
```

Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift
```diff
@@ -64,7 +64,8 @@ extension HTTPClientTests {
             ("testNoResponseWithIgnoreErrorForSSLUncleanShutdown", testNoResponseWithIgnoreErrorForSSLUncleanShutdown),
             ("testWrongContentLengthForSSLUncleanShutdown", testWrongContentLengthForSSLUncleanShutdown),
             ("testWrongContentLengthWithIgnoreErrorForSSLUncleanShutdown", testWrongContentLengthWithIgnoreErrorForSSLUncleanShutdown),
-            ("testEventLoopArgument", testEventLoopArgument),
+// TODO: Comment back in once failure was resolved
+//            ("testEventLoopArgument", testEventLoopArgument),
```
slashmo, line 68: Needs investigation as it currently leads to failing precondition. Commented out, for now, to be able to run the full test-suite.

Tests/AsyncHTTPClientTests/HTTPClientTests+XCTest.swift
```diff
@@ -88,7 +89,8 @@ extension HTTPClientTests {
             ("testUncleanShutdownCancelsTasks", testUncleanShutdownCancelsTasks),
             ("testDoubleShutdown", testDoubleShutdown),
             ("testTaskFailsWhenClientIsShutdown", testTaskFailsWhenClientIsShutdown),
-            ("testRaceNewRequestsVsShutdown", testRaceNewRequestsVsShutdown),
+// TODO: Comment back in once failure was resolved
+//            ("testRaceNewRequestsVsShutdown", testRaceNewRequestsVsShutdown),
```
slashmo, line 93: Needs investigation as it currently leads to a failing assertion. Commented out, for now, to be able to run the full test-suite.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -147,3 +148,37 @@ extension Connection {
         }.recover { _ in }
     }
 }
+
+extension SpanStatus {
+    /// Map status code to canonical code according to OTel spec
+    ///
+    /// - SeeAlso: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#status
+    init(_ responseStatus: HTTPResponseStatus) {
+        switch responseStatus.code {
```
pokryfka, line 157: I think mapping of HTTP status code (UInt) to `SpanStatus` is very much reusable and as such should be provided in `TracingInstrumentation`, sth like

```swift
SpanStatus(code: UInt, message: String?)
``` 

otherwise each library making HTTP calls and not using AHC will need to map it on its own

ktoso: Hah, this reminded me I had an not submitted review here, yeah this seems like a good candidate to move up into OpenTelemetryInstrumentationSupport/OpenTelemetrySemanticConventions 👍

slashmo: Or not such a good candidate after all 😁👉 https://github.com/slashmo/gsoc-swift-tracing/issues/134

ktoso:

Sources/AsyncHTTPClient/HTTPClient.swift
slashmo: I just reread the Otel definition for HTTP client spans. Your correct, the combination of host/scheme/target is not their recommended way for client spans: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#http-client

> One of the following sets of attributes is required (in order of usual preference unless for a particular web client/framework it is known that some other set is preferable for some reason; all strings must be non-empty):

> - http.url
> - http.scheme, http.host, http.target
> - http.scheme, net.peer.name, net.peer.port, http.target
> - http.scheme, net.peer.ip, net.peer.port, http.target

Therefore I'll be replacing scheme/host/target with the full url. Thanks for pointing this out 👍

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2602,4 +2612,98 @@ class HTTPClientTests: XCTestCase {
...
+        ofKind kind: SpanKind,
+        at timestamp: Timestamp
+    ) -> Span {
+        let span = TestSpan(operationName: operationName,
+                            kind: kind,
+                            startTimestamp: timestamp,
+                            context: context.baggage)
+        recordedSpans.append(span)
+        return span
+    }
```
pokryfka, line 2653: the context should be updated, see https://github.com/slashmo/gsoc-swift-tracing/issues/126#issuecomment-674630622

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -461,32 +383,29 @@ public class HTTPClient {
...
+
+        // TODO: net.peer.ip / Not required, but recommended
+
+        var request = request
+        InstrumentationSystem.instrument.inject(span.context.baggage, into: &request.headers, using: HTTPHeadersInjector())
+
+        let logger = context.logger.attachingRequestInformation(request, requestID: globalRequestID.add(1))
```
pokryfka, line 407: @ktoso @slashmo 

this updates an instance of logger,
should it not set requestId in baggage (which then would be used to setup logger)?

pokryfka: (copied from https://github.com/swift-aws/aws-sdk-swift-core/pull/348#discussion_r478508462)

how about sth like:

```swift
    private struct RequestMetadata {
        var requestId: Int
        var service: String
        var operation: String
    }
    
    private enum RequestKey: BaggageContextKey {
        typealias Value = RequestMetadata
        var name: String { "AWSRequest" }
    }
 ```

then updating the baggage like:

```swift
context.baggage[RequestKey.self] = RequestMetadata(requestId: Self.globalRequestID.add(1),
    service: serviceConfig.service,
    operation: "signURL")
```

instead of updating the logger directly (current):

```swift
context.logger = context.logger.attachingRequestId(Self.globalRequestID.add(1), operation: "signURL", service: serviceConfig.service)
```

PR #291, opened 2020-08-06
implement new streamwriter api

### LINKED ISSUE #194: unclear docs: How to signal to StreamWriter that body is done?
`StreamWriter`'s complete docs are 

```
        /// Chunk provider.
        public struct StreamWriter {
            /// Write data to server.
            ///
            /// - parameters:
            ///     - data: `IOData` to write.
            public func write(_ data: IOData) -> EventLoopFuture<Void>
        }
```

I understand how I can provide chunks: by calling `write` but how do I tell the `StreamWriter` that I'm done sending request body bytes?

### LINKED ISSUE #264: StreamWriter API revamp: access to ByteBufferAllocator
Once we revamp the StreamWriter API (see also #194), we should have StreamWriter give the user access to a `ByteBufferAllocator`.

artemredkin:
Revamps `StreamWriter`.

Motivation:
There are two issues with current `StreamWriter` API:
1. no access to allocator
2. writer completion is super confusing, it will be considered finished when the returned `EventLoopFuture` is completed, which is not ideal.

Modifications:
1. Deprecate old `StreamWriter` API
2. Introduce new type `StreamWriter2` (really need help with the naming here) with the following API:
```swift
let body: HTTPClient.Body = .stream(length: 8) { writer in
    writer.write(writer.allocator.buffer(string: "1234").whenComplete { _ in
        writer.write(writer.allocator.buffer(string: "1234").whenComplete { _ in
            writer.end()
        }
    }
}
```

Result:
Closes #194 
Closes #264

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -32,16 +32,59 @@ extension HTTPClient {
             ///
             /// - parameters:
             ///     - closure: function that will be called to write actual bytes to the channel.
+            @available(*, deprecated, message: "StreamWriter is deprecated, please use StreamWriter2")
             public init(closure: @escaping (IOData) -> EventLoopFuture<Void>) {
                 self.closure = closure
+            }
+
+            // This is needed so we don't have build warnings in the client itself
+            init(internalClosure: @escaping (IOData) -> EventLoopFuture<Void>) {
+                self.closure = internalClosure
             }
 
             /// Write data to server.
             ///
             /// - parameters:
             ///     - data: `IOData` to write.
+            @available(*, deprecated, message: "")
             public func write(_ data: IOData) -> EventLoopFuture<Void> {
                 return self.closure(data)
+            }
+        }
+
+        public struct StreamWriter2 {
+            public let allocator: ByteBufferAllocator
+            let onChunk: (IOData) -> EventLoopFuture<Void>
+            let onComplete: EventLoopPromise<Void>
+
+            public init(allocator: ByteBufferAllocator, onChunk: @escaping (IOData) -> EventLoopFuture<Void>, onComplete: EventLoopPromise<Void>) {
+                self.allocator = allocator
+                self.onChunk = onChunk
+                self.onComplete = onComplete
+            }
+
+            public func write(_ buffer: ByteBuffer) -> EventLoopFuture<Void> {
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -67,17 +127,30 @@ extension HTTPClient {
...
             return Body(length: length, stream: stream)
         }
 
+        /// Create and stream body using `StreamWriter`.
+        ///
+        /// - parameters:
+        ///     - length: Body size. Request validation will be failed with `HTTPClientErrors.contentLengthMissing` if nil,
+        ///               unless `Transfer-Encoding: chunked` header is set.
+        ///     - stream: Body chunk provider.
+        public static func stream2(length: Int? = nil, _ stream: @escaping (StreamWriter2) -> Void) -> Body {
```
artemredkin, line 141: I we add method with the same name but different type we will break compilation, because some calls could become ambiguous 😢

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -32,32 +32,92 @@ extension HTTPClient {
...
+            @available(*, deprecated, message: "StreamWriter is deprecated, please use StreamWriter2")
             public func write(_ data: IOData) -> EventLoopFuture<Void> {
                 return self.closure(data)
             }
         }
 
+        public struct StreamWriter2 {
+            public let allocator: ByteBufferAllocator
+            let onChunk: (IOData) -> EventLoopFuture<Void>
+            let onComplete: EventLoopPromise<Void>
```
Lukasa, line 58: Is there a reason we're writing this as a callback-taking type rather than just holding onto the task object?

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: Given that there is discussion about possibly breaking API in the nearish future _anyway_, we may find we want to delay this until then to resolve this problem.

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: and "forward" all write requests to it's channel? that might be a good idea, will try that, thanks!

ktoso:

PR #295, opened 2020-08-20
Fixes double-release of a connection.

### LINKED ISSUE #294: Task Handler handles events even after channel.close
This could lead to double-release errors in connection pool. We need to guard on handler state.

artemredkin:
Motivation:
TaskHandler unconditionally releases it's connection on error,
this can lead to double release. This issue actually indicates
a more general issue where handler continues to handle errors
even after its state is `.endOrError`. We need to fix this by
ignoring all subsequent errors.

Modifications:
1. Check state before calling out delegate and pool
2. Replace all error callouts with call to `errorCaught`

Result:
Fixes #294


Lukasa (approved):
Looks like a solid fix to me.

PR #298, opened 2020-08-21
Fix state reverting

### LINKED ISSUE #297: TaskHandler state can be reset after error
`.endOrError` should be terminal state

artemredkin:
Prevent TaskHandler state change after `.endOrError`
    
Motivation:
Right now if task handler encounters an error, it changes state to `.endOrError`. We gate on that state to make sure that we do not process errors in the pipeline twice. Unfortunately, that state can be reset when we upload body or receive response parts.
    
Modifications:
Adds state validation before state is updated to a new value
Adds a test
    
Result:
Fixes #297

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -840,15 +840,22 @@ extension TaskHandler: ChannelDuplexHandler {
             self.writeBody(request: request, context: context)
         }.flatMap {
             context.eventLoop.assertInEventLoop()
+            if case .endOrError = self.state {
+                return context.eventLoop.makeSucceededFuture(())
+            }
+
             self.state = .bodySent
             if let expectedBodyLength = self.expectedBodyLength, expectedBodyLength != self.actualBodyLength {
                 let error = HTTPClientError.bodyLengthMismatch
-                self.errorCaught(context: context, error: error)
                 return context.eventLoop.makeFailedFuture(error)
             }
             return context.writeAndFlush(self.wrapOutboundOut(.end(nil)))
         }.map {
             context.eventLoop.assertInEventLoop()
+            if case .endOrError = self.state {
+                return
+            }
+
             self.state = .sent
             self.callOutToDelegateFireAndForget(self.delegate.didSendRequest)
         }.flatMapErrorThrowing { error in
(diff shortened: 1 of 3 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -924,23 +934,28 @@ extension TaskHandler: ChannelDuplexHandler {
         let response = self.unwrapInboundIn(data)
         switch response {
         case .head(let head):
-            if !head.isKeepAlive {
-                self.closing = true
-            }
+            switch self.state {
```
Lukasa, line 937: This may read more clearly as:

```swift
if case .endOrError = self.state {
    return
}
```

Rather than forcing the indentation of the rest of the body.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -839,18 +839,28 @@ extension TaskHandler: ChannelDuplexHandler {
...
-                self.errorCaught(context: context, error: error)
-                return context.eventLoop.makeFailedFuture(error)
+            switch self.state {
+            case .endOrError:
+                return context.eventLoop.makeSucceededFuture(())
+            default:
+                self.state = .bodySent
+                if let expectedBodyLength = self.expectedBodyLength, expectedBodyLength != self.actualBodyLength {
+                    let error = HTTPClientError.bodyLengthMismatch
+                    self.errorCaught(context: context, error: error)
```
Lukasa, line 850: We don't need `errorCaught` here: the future chain will call this code path anyway.

Lukasa (commented):
Broadly looks good, I've left a couple of notes in the diff.

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: ah! thank you, always forget about this pattern

Sources/AsyncHTTPClient/HTTPHandler.swift

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: Seems like the only place you didn't apply this pattern was this block? :wink:

artemredkin: 😅

Lukasa (approved):
Love it, looks great.

PR #303, opened 2020-09-08
Use assertInEventLoop over assert(el.inEventLoop)

fabianfett:
### Motivation

@weissi made me aware at multiple instances that `el.assertInEventLoop()` should be used over `assert(el.inEventLoop)`, since it ensures that the correct behavior is called for `NIOTSEventLoop`s. 

### Changes

Replace three instances of `assert(el.inEventLoop)` with `el.assertInEventLoop()`.


Lukasa (approved):

Lukasa:
Great catch!

PR #305, opened 2020-09-10
Fix FileDownloadDelegate example code in README.md

MaxDesiatov:
The progress API of `FileDownloadDelegate` changed at the last minute, but the example in `README.md` wasn't updated accordingly.


Lukasa:
Good catch, thanks!

Lukasa (approved):

Issue #234, opened 2020-05-29
HTTP1ConnectionsProvider.state needs to become `private`


weissi:
Right now, `HTTP1ConnectionsProvider.state` is internal and some tests modify the state without `HTTP1ConnectionsProvider` knowing about this.

That cannot be right because `HTTP1ConnectionsProvider` cannot clean up stuff it didn't know happened. Therefore we see extremely fragile and ugly cleanup code in some tests.

weissi:
when this is fixed, `HTTP1ConnectionProvider.closePromise` must become `private` too, there's no reason for the tests to fulfil  internal promises of the provider.

PR #310, opened 2020-09-28
Code clean up to to fix issue #234 - ConnectionsState exposes a setter into internal state

ya-pulser:

PR #311, opened 2020-09-28
Fixes issue #234 : ConnectionsState exposes a setter into internal state

ya-pulser:
This PR is built on top of preparation work submitted as PR #310 which contains code clean up, before changes in this PR will affect tests and implementation of ConnectionsState.

Work with the tests has shown that some state invariants could be achieved only by direct manipulation with the state, so this proposal removes tests and code related to such "invalid" states.


artemredkin:

Sources/AsyncHTTPClient/ConnectionsState.swift
```diff
@@ -151,37 +142,29 @@ extension HTTP1ConnectionProvider {
...
                         let (eventLoop, required) = self.resolvePreference(waiter.preference)
 
                         // If returned connection is on same EL or we do not require special EL - lease it
-                        if connection.channel.eventLoop === eventLoop || !required {
+                        if connection.eventLoop === eventLoop || !required {
                             return .lease(connection, waiter)
                         }
 
-                        // If there is an opened connection on the same loop, lease it and park returned
-                        if let found = self.availableConnections.firstIndex(where: { $0.channel.eventLoop === eventLoop }) {
```
artemredkin, line 169: the following two ifs were removed since they cannot be reached, we added and assert on line 156 instead

Sources/AsyncHTTPClient/ConnectionsState.swift
```diff
@@ -285,22 +259,14 @@ extension HTTP1ConnectionProvider {
...
-        mutating func processNextWaiter() -> Action {
+        mutating func processNextWaiter() -> Action<ConnectionType> {
             if let waiter = self.waiters.popFirst() {
-                let (eventLoop, required) = self.resolvePreference(waiter.preference)
+                // There should be no case where we have waiters and available connections. As soon as connection
+                // is released and can be available, it should immideately go to the next waiter.
+                assert(self.availableConnections.isEmpty)
 
-                // If specific EL is required, we have only two options - find open one or create a new one
-                if required, let found = self.availableConnections.firstIndex(where: { $0.channel.eventLoop === eventLoop }) {
```
artemredkin, line 293: the following block was remove since it cannot be reached, we replaced it with an assert on line 266

artemredkin (approved):

artemredkin:

artemredkin (approved):
🙏

Sources/AsyncHTTPClient/ConnectionsState.swift
```diff
@@ -158,30 +149,22 @@ extension HTTP1ConnectionProvider {
 
                 if connection.isActiveEstimation, !closing { // If connection is alive, we can offer it to a next waiter
                     if let waiter = self.waiters.popFirst() {
+                        // There should be no case where we have both capacity and a waiter here.
+                        // Waiter can only exists if there was no capacity at aquire. If some connection
+                        // is released when we have waiter it can only indicate that we should lease (if EL are the same),
+                        // or replace (if they are different). But we cannot increase connection count here.
+                        assert(!self.hasCapacity)
```
Lukasa, line 156: Does this code behave gracefully if this invariant is not true?

Sources/AsyncHTTPClient/ConnectionsState.swift
```diff
@@ -287,20 +261,12 @@ extension HTTP1ConnectionProvider {
 
         mutating func processNextWaiter() -> Action<ConnectionType> {
             if let waiter = self.waiters.popFirst() {
-                let (eventLoop, required) = self.resolvePreference(waiter.preference)
+                // There should be no case where we have waiters and available connections. As soon as connection
+                // is released and can be available, it should immideately go to the next waiter.
```
Lukasa, line 265: Is this true in the face of event loop preferences? Is it not possible to have a waiter that is waiting for a connection on the appropriate event loop?

artemredkin: Yes. We can only have waiters if there are no available connections, otherwise any new waiter will be leased a connection or a new connection will be created. If connection is released and it's active (meaning it could become "available"), it will be handled by a first branch in `release`. If the connection is not active, we will get here. But at this point the available invariant holds, released connection cannot become available, so the only course of action is to create a new one. Does that make sense?

artemredkin: There are three other call sites for `processNextWaiter`:
1. `connectFailed`. This is somewhat similar - that failed connection will not be returned to the available set, since its already closed, so again, the only course of action is to create a new connection.
2. `remoteClose` - same thing
3. `timeout` - same thing as well

Lukasa: We need to be extremely clear about our reasoning in this code. The comment reads: "As soon as connection is released and can be available, it should immideately go to the next waiter." I have natural questions about this: if the waiter has expressed a loop preference, it may be possible that a released connection is unsuitable for that waiter, despite being valid and active. What happens in this case?

It's totally possible that the _code_ is right, but the comment is at best incomplete.

artemredkin: if the released connection is not suitable for the next waiter, it will be `.replace`d. How about the following comment instead:
```
This method is called in 4 cases:
1. from `release` when connection is inactive and cannot be re-used
2. from `connectFailed` when we failed to establish a new connection
3. from `remoteClose` when connection was closed by the remote side and cannot be re-used
4. from `timeout` when connection was closed due to idle timeout and cannot be re-used.
In all cases connection will not be `available`. Given that the waiter can only be present if there are no available connections (otherwise it would be leased a connection), where can be no situation where we can lease some available connection and the only course of action is to create a new one.
```

Sources/AsyncHTTPClient/ConnectionsState.swift
artemredkin: Good question. Worst case scenario - we will replace the connection with a new one without checking for potentially available connections first. Though if this happens, it means that there is a bug in our code... What do you think we should do if for some reason we fail this?

Sources/AsyncHTTPClient/ConnectionsState.swift
ya-pulser: > where can be no situation where we can

ya-pulser:
```
This method is called in following cases:

1. from `release` when connection is inactive and cannot be re-used
2. from `connectFailed` when we failed to establish a new connection
3. from `remoteClose` when connection was closed by the remote side and cannot be re-used
4. from `timeout` when connection was closed due to idle timeout and cannot be re-used.

In all cases connection, which triggered the transition, will not be in `available` state.

Given that the waiter can only be present in the pool if there were no available connections 
(otherwise it would be leased a connection immediately), we do not see a situation when we can lease another
available connection, therefore the only course of action is to create a new connection for the waiter.
```

Complex comment highlights that logic is convoluted here, but I am not ready to propose an immediate improvement for the logic.

Lukasa: Happy for the complex comment to go in.

ya-pulser:
elaborated comment for https://github.com/swift-server/async-http-client/pull/311#discussion_r498146048 is pushed into the code

ya-pulser:
I did amend / force push which I am not sure is the best way if we have "squash" when we merge in this repository. Sorry.

artemredkin (approved):

Lukasa (approved):

PR #312, opened 2020-10-07
remove symbolicate-linux-fatal from Docker

tomerd:
motivation: we are not actually using symbolicate-linux-fatal in any meaningful way in CI and it's pinned to the master branch which has been removed

changes: remove symbolicate-linux-fatal fetching from Docker


ktoso (approved):

ktoso:
LGTM, I suspect no disagreement about removing this 👍

PR #315, opened 2020-11-11
add another expires date format

### LINKED ISSUE #314: A valid Expires= part of Set-Cookie header is not recognized
The https://tools.ietf.org/html/rfc2616#section-3.3.1 specifies the `Sunday, 06-Nov-94 08:49:37 GMT` as a valid date format.

The https://tools.ietf.org/html/rfc6265#section-5.1.1 reinforces that.

The `sane-cookie-date` non-terminal in https://tools.ietf.org/html/rfc6265#section-4.1.1 reinforces that.

However, the parser only recognizes the `Sun, 06 Nov 1994 08:49:37 GMT` variant of the Expires= token, effectively not restricting validity of the cookie header.

artemredkin:
Adds `rfc2616#section-3.3.1` date format parsing.

Motivation:
The https://tools.ietf.org/html/rfc2616#section-3.3.1 specifies the Sunday, 06-Nov-94 08:49:37 GMT as a valid date format. We need to support that.

Modifications:
Updated cookie parsing to support additional date format.

Result:
Closes #314


Lukasa (changes requested):
RFC 2616 was obsoleted a while back. The relevant citation is now [RFC 7231 § 7.1.1.1](https://tools.ietf.org/html/rfc7231#section-7.1.1.1), and that provides us a normative MUST:

>    A recipient that parses a timestamp value in an HTTP header field
>    MUST accept all three HTTP-date formats.

We are currently only accepting two: we should update this patch to tolerate all three.

artemredkin:
Good catch, fixed!

Lukasa (approved):
LGTM.

PR #316, opened 2020-11-11
Fixes propagation of errors during TLS handshake

### LINKED ISSUE #313: NoResult error on errors before TaskHandler is added
If something other that `handshakeFailed` happens during SSL handshake (for example `uncleanShutdown`), we will get `NoResult` error and real error will not be propagated

artemredkin:
Motivation:
Right now we only handle one type of SSL error: `.handshakeFailed`,
but in reality a multitude of errors can happen, for example, remote
party might just close connection that will in turn raise
`.uncleanShutdown` error, that will be dropped on the floor and
users will only get non-descriptive `NoResult` error.

Modifications:
Handle all types of SSL errors during handshake instead of just one.
Adds a test.

Result:
Closes #313

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -935,16 +935,9 @@ class TLSEventsHandler: ChannelInboundHandler, RemovableChannelHandler {
     }
 
     func errorCaught(context: ChannelHandlerContext, error: Error) {
-        if let sslError = error as? NIOSSLError {
-            switch sslError {
-            case .handshakeFailed:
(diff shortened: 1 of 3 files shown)


Lukasa (commented):
Why don't other errors want to do the same things here?

artemredkin:
Great question, in test we actually produce two errors, first is `connection reset by peer` and second is `uncleanShutdown`, could there be a case were we get `IOError` and no `SSLError`? I've updated the PR to handle all errors for now

Lukasa (approved):
Seems sensible to me. I think we should be pessimistic about errors and assume weird things can happen.

Issue #318, opened 2020-11-24
Access to ConnectionPool.count should be thread safe


ya-pulser:
ConnectionPool has a collection of `providers` registered with the pool and  access to the collection is guarded by ConnectionPool.lock.

Size of the collection is exposed through `count` accessor without guarding by the lock.
https://github.com/swift-server/async-http-client/blob/1.2.2/Sources/AsyncHTTPClient/ConnectionPool.swift#L112

That leads to warnings in thread safety analyser.

To reproduce run with thread safety analyzer:

```
docker run --rm -v $(pwd):/src -w /src swift:5.3.1 swift test -c release --sanitize=thread --enable-test-discovery
```

Analysis of output shows that this `count` accessor is main source of thread safety warnings.

ya-pulser:
PR #319 as a proposed solution

Lukasa:
I believe we’re done here: see https://github.com/swift-server/async-http-client/pull/319#discussion_r530515030 for more reasoning.

Lukasa:
@weissi has a good reference for a TSAN issue in 5.3 that manifests this way.

weissi:
@fabianfett I think we can close this?

fabianfett:
I'll add it to the http/2 milestone and close it as soon as the connection pool has landed.

fabianfett:
Fixed in #319 and removed from an active code path in #427.

PR #319, opened 2020-11-25
issue #318: Address warning reported by thread safety analyzer 

ya-pulser:
Motivation:

Thread safety analyzer reports warning about observable state
leaving lock guarded code blocks. This PR address the warnings.

Analysis of the warning does not prove that we have a real bug and all the
warnings are considered as potential bugs.

Analyzer used is:
`docker run --rm -v $(pwd):/src -w /src swift:5.3.1 swift test -c
release --sanitize=thread --enable-test-discovery`

Modifications:

* accessor to count of connections in connection pool is guarded by lock
* forced materialization of a potentially lazy collection into an array before returning
  the collection from lock guarded code block

Result:

Most of thread safety warnings are addressed without modification of observable
code behaviour.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -110,7 +110,9 @@ final class ConnectionPool {
     }
 
     var count: Int {
-        return self.providers.count
+        return self.lock.withLock {
+            return self.providers.count
+        }
     }
 
     /// Used by the `ConnectionPool` to index its `HTTP1ConnectionProvider`s


ya-pulser:
@Lukasa Could you please review and guide. I am not sure that we can have tests to accompany the change and I am not sure that I can fully defend that `Array(Values)` change.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -103,14 +103,19 @@ final class ConnectionPool {
 
     func close(on eventLoop: EventLoop) -> EventLoopFuture<Bool> {
         let providers = self.lock.withLock {
-            self.providers.values
+            // we materialize Values into an Array before returning from under a lock
+            // as thread safety analyzer considers safety of Values as undefined and dependant on implementation
+            // e.g. a lazy collection implementation could by thread unsafe to be returned from within a lock
+            Array(self.providers.values)
```
Lukasa, line 109: I do not believe this fix is necessary: I think TSAN is wrong.

Lukasa (commented):
Agree that testing this is very hard. The best test I can come up with is to spin off several dispatch queues and rapidly attempt to mutate the providers in one thread while reading from another. This will catch the bug deterministically under TSAN.

Sources/AsyncHTTPClient/ConnectionPool.swift
ya-pulser: I am lacking deep knowledge of Swift runtime / libraries here :(.

Is it possible that @ktoso could advise here?

I seen in person issues with returning a lazy collection out of lock guarded block in Scala many years ago and performing of "materialise" was a go-to tool then.

Lukasa: While the dictionary values view is "lazy", conceptually, it is nonetheless still a value type with value semantics. That means that changes to the dictionary to which it points must not manifest in changes to this value. _Acquiring_ the view must be done with a lock, but once it's copied out we're safe. Either TSAN is wrong, or Dictionary in Swift 5.3 is implemented incorrectly.

ya-pulser: PR is updated to contain only `.count` fix according to the discussion above.

Lukasa (approved):

PR #320, opened 2020-12-05
Trace HTTPClient request execution

slashmo:
Motivation:

## Context Propagation
In order to instrument distributed systems, metadata such as trace ids
must be propagated across network boundaries through HTTP headers.
As HTTPClient operates at one such boundary, it should take care of
injecting metadata into HTTP headers automatically using the configured
instrument.

## Built-in tracing
Furthermore, `HTTPClient` should create a `Span` for executed requests 
under the hood, so that users benefit from tracing effortlessly.

Modifications:

- [x] Inject instrumentation metadata into HTTP headers
- [ ] Add `HTTPClient` method overloads accepting `LoggingContext`
- [ ] Create `Span` for executed HTTP request

Result:

- New HTTPClient method overloads accepting LoggingContext
- Existing overloads accepting Logger construct a DefaultLoggingContext
- Existing methods that neither take Logger nor LoggingContext construct
  a DefaultLoggingContext


slashmo:
I chatted with @ktoso earlier to discuss the manual context propagation, and we agreed that we probably shouldn't deprecate the "old" API accepting a `Logger` for each request overload, as we don't want to push users too much into the direction of manual context passing because that's ideally not necessary once the mentioned language changes have been made: https://github.com/apple/swift-distributed-tracing#important-note-on-adoption

Package.swift
```diff
@@ -26,12 +26,13 @@ let package = Package(
...
         .package(url: "https://github.com/apple/swift-nio-transport-services.git", from: "1.5.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
+        .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "0.1.0"),
     ],
     targets: [
         .target(
             name: "AsyncHTTPClient",
             dependencies: ["NIO", "NIOHTTP1", "NIOSSL", "NIOConcurrencyHelpers", "NIOHTTPCompression",
-                           "NIOFoundationCompat", "NIOTransportServices", "Logging"]
+                           "NIOFoundationCompat", "NIOTransportServices", "Logging", "Instrumentation"]
```
ktoso, line 35: Can we right away go with `Tracing` and do the full thing in a single PR?

ktoso (commented):
So since technically we're 0.1 and something _may_ change... how do we want to tackle adoption here.

I was thinking to kick off a branch like `tracing` for now, so we can polish up there and once we're all confident merge into mainline? We could also tag those tracing releases, they'd follow normal releases e.g. 1.2.2-tracing. 

I don't really expect anything breaking in the core APIs but the open telemetry support which we may want to use here could still fluctuate a little bit until they're final hmmm...

Package.swift
slashmo: That's my intention. I've added a checklist to the PR including creating a Span. I first wanted to get the instrumentation part down and then continue with tracing, but all inside this PR.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -227,7 +229,21 @@ public class HTTPClient {
...
+        )
+    }
+
+    /// Execute `GET` request using specified URL.
+    ///
+    /// - parameters:
+    ///     - url: Remote URL.
+    ///     - deadline: Point in time by which the request must complete.
+    ///     - context: The logging context carrying a logger and instrumentation metadata.
+    public func get(url: String, deadline: NIODeadline? = nil, context: LoggingContext) -> EventLoopFuture<Response> {
```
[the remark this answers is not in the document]
slashmo: Yep, you're right. To be honest, I didn't think about that. Putting it in front of `deadline` sounds good to me.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -71,6 +73,7 @@ public class HTTPClient {
     private let stateLock = Lock()
 
     internal static let loggingDisabled = Logger(label: "AHC-do-not-log", factory: { _ in SwiftLogNoOpLogHandler() })
+    internal static let topLevelContextLoggingDisabled = DefaultLoggingContext(logger: loggingDisabled, baggage: .topLevel)
```
slashmo, line 76: I'm not very happy with the naming of this constant. Does anybody have any other suggestions? On a related note, I was also thinking about extending `BaggageContext` to add a factory method accepting an optional `Logger` and using this instead. That way in the `HTTPClient` method overloads taking a `Logger`, we could delegate to the matching overload taking `LoggingContext` like this:

```swift
self.get(url: url, context: .topLevel(logger: logger), deadline: deadline)
```

What do you think about this? Might be something to put in the `Baggage` repo too 🤔

Ping @ktoso

slashmo: > What do you think about this? Might be something to put in the Baggage repo too 🤔
> @slashmo

On the other hand, we also don't want to make it too easy to create empty `Baggage`s.

Lukasa:

Lukasa:
I'd like to punt this to a side-branch for iterative development if we can.

slashmo:
> I'd like to punt this to a side-branch for iterative development if we can.
> @Lukasa 

Sure, sounds like a good approach. I can change the target branch once it's created.

Lukasa:
I've opened up the `tracing-development` branch.

slashmo:
@ktoso The CI seems to fail because the Baggage repo cannot be cloned through the Git URL. Should we pin this Tracing to 0.1.1 here in order to get the fix? (apple/swift-distributed-tracing/pull/25)

ktoso:
No, we need to tag a 0.1.1, I'll do that in a moment.

ktoso:
0.1.1. tagged, please depend on that.

Thanks Cory for the development branch, sounds good 👍

ktoso:

ktoso:

ktoso:
Can drafts get CI validation? 🤔

Lukasa:
Yes, they can: I think the CI isn't targeting that branch at the moment.

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: > BaggageContext to add a factory method accepting an optional Logger and using this instead. 

There is no BaggageContext; You mean `LoggingContext`? Yeah I wish we could do that but we can't because a protocol, so this won't work:

```swift
 15> func take(_ p: P) {  }
 16> take(.make())
error: repl.swift:16:7: error: static member 'make' cannot be used on protocol metatype 'P.Protocol'
take(.make())
    ~~^~~~~~~
```

This API must be geared towards "pass-along whatever context you already have" which is what the LoggingContext is designed for. Passing in an "empty" or "just log" is annoying and that's for better or worse the price we pay. If someone wants just logging with no context the logger accepting overload exists today, so nothing we should try to improve about this.

--- 


> > Might be something to put in the Baggage repo too 🤔
> On the other hand, we also don't want to make it too easy to create empty Baggages.

Yeah it's on purpose that creating "empty not logging anything" is not built in. If you choose to provide that as library, that's on you.

slashmo: > If someone wants just logging with no context the logger accepting overload exists today, so nothing we should try to improve about this.
> @ktoso 

Ah, I haven't noticed the `DefaultLoggingContext.topLevel` method. With that in place I'd do:

```swift
internal static let loggingDisabledTopLevelContext = DefaultLoggingContext.topLevel(logger: loggingDisabled)
```

Or get rid of the constant and use it in place, becomes even shorter than the long constant name 😅

```swift
.topLevel(logger: loggingDisabled) // or
.topLevel(logger: logger)
```

slashmo: Ah, `.topLevel` won't work because it accepts `LoggingContext` as you mentioned. I'm using the `topLevel` method for the existing `Logger` overloads, and `loggingDisabledTopLevelContext` for the default methods.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -112,46 +126,76 @@ extension HTTPClient {
...
-                    self.poolManager.executeRequest(transaction)
-                }
-            },
-            onCancel: {
-                cancelHandler.cancel(reason: .taskCanceled)
+        return try await withSpan(request.head.method.rawValue, context: context, ofKind: .client) { span in
+            var request = request
+            request.head.headers.propagate(span.context)
+            span.updateAttributes { attributes in
+                attributes["http.request.method"] = request.head.method.rawValue
```
czechboy0, line 147: Do we also attach the path as a separate attribute, or is `url.full` the only place it's expected to show up?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -112,46 +126,76 @@ extension HTTPClient {
...
+                        span.setStatus(.init(code: .error))
+                        span.attributes["error.type"] = "\(CancellationError.self)"
+                        cancelHandler.cancel(reason: .taskCanceled)
+                    }
+                )
+                return response
+            } catch {
+                span.setStatus(.init(code: .error))
+                span.attributes["error.type"] = "\(type(of: error))"
+                throw error
```
czechboy0, line 196: Does this mean the span doesn't include the error description, and it's up to user code to attach it higher up the stack?

ktoso: Comes back to the entire "can we log errors" (not really since no idea what they contain) discussion we were having elsewhere recently. 😅 

The otel guidance is the same tbh: https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-spans but you'll notice it gets hand wavy with it with exception messages -- Java influenced wording there; 

> SHOULD set the span **status code** to Error
> 
> SHOULD set the [**error.type**](https://opentelemetry.io/docs/specs/semconv/attributes-registry/error/#error-type) attribute
> 
> SHOULD set the span status description **when it has additional information about the error which is not expected to contain sensitive details** and aligns with [Span Status Description](https://opentelemetry.io/docs/specs/otel/trace/api/#set-status) definition.
> 
> It’s NOT RECOMMENDED to duplicate status code or error.type in span status description.
> 
> **When the operation fails with an exception, the span status description SHOULD be set to the exception message.**

So that leaves us back at step 1 where we need to decide what is safe -- the "exception message" is not a thing in Swift after all, and "\(error)" may or may not be safe...

Therefore I think the default behavior of HTTP client can only be the error type probably?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -112,46 +126,76 @@ extension HTTPClient {
...
-                    )
-
-                    cancelHandler.registerTransaction(transaction)
-
-                    self.poolManager.executeRequest(transaction)
-                }
-            },
-            onCancel: {
-                cancelHandler.cancel(reason: .taskCanceled)
+        return try await withSpan(request.head.method.rawValue, context: context, ofKind: .client) { span in
```
ktoso, line 143: LGTM on using just the method as low cardinality name ([matches otel well](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name)) 👍

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
ktoso: Path may be nice tbh... it's not a "standardized" convention but I could see it be useful... Up to y'all if we want to duplicate; url.full seems to be the "expected one" to set otel wise...

czechboy0: `url.path` seems to be recommended for the server span only here: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-semantic-conventions? But here it's simply listed, without that server guidance: https://opentelemetry.io/docs/specs/semconv/attributes-registry/url/#url-path

I guess wouldn't hurt to add it, but I guess the client is expected not to have the full URL broken out into components, but the server does.

I'm happy to keep it as-is, with just `url.full`.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
czechboy0: Unsure, but I wouldn't want this PR to be blocked on that, I'm happy with the first iteration not including the message here.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
czechboy0, line 143: Should the `Tracer` be a property of the client, to allow for easier unit testing? And just default it to the current process-bootstrapped tracer in the initializer, unless it's overridden?

slashmo: We could do that, yeah. Elsewhere we just use `InstrumentationSystem.bootstrapInternal` in unit tests which also works fine but I'd be happy either way.

czechboy0: Unless there are downsides, I prefer the explicit dependency injection. Easy enough to debug and reason about.

slashmo: Sounds good to me as long as we don't expose this initializer publicly.

czechboy0: What's the downside of adding it to the public initializer?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
slashmo: Yeah, I think it's fine to just include `url.full` for now. That being said, something I haven't yet implemented in the PR is redaction as recommended (although `experimental`) in OTel SemConv:

https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client

> [4] `url.full`: For network calls, URL usually has scheme://host[:port][path][?query][#fragment] format, where the fragment is not transmitted over HTTP, but if it is known, it SHOULD be included nevertheless.

> `url.full` MUST NOT contain credentials passed via URL in form of https://username:password@www.example.com/. In such case username and password SHOULD be redacted and attribute’s value SHOULD be https://REDACTED:REDACTED@www.example.com/.
> Sensitive content provided in `url.full` SHOULD be scrubbed when instrumentations can identify it.

> Experimental Query string values for the following keys SHOULD be redacted by default and replaced by the value REDACTED:
> [AWSAccessKeyId](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth) / [Signature](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth) / [sig](https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token) / [X-Goog-Signature](https://cloud.google.com/storage/docs/access-control/signed-urls)
> This list is subject to change over time.

> When a query string value is redacted, the query string key SHOULD still be preserved, e.g. https://www.example.com/path?color=blue&sig=REDACTED.

What do you think about adding these? I'm a bit torn but I don't think we'd want to include these specific keys in `AsyncHTTPClient`. Perhaps we could skip adding them for now and only redact the basic auth parameters, and tackle the other redaction in [swift-otel/swift-otel-semantic-conventions](https://github.com/swift-otel/swift-otel-semantic-conventions) and eventually add a dependency on it to `AsyncHTTPClient`.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
slashmo: It goes against the way Swift Distributed Tracing usually works, where a single tracer is chosen for the system and used transparently everywhere. Having said that, Logging also allows passing specific `Logger`s around and still use `LoggingSystem`. I'm curious to hear what @ktoso thinks about this.

PR #321, opened 2020-12-14
Add NIOTransportServices TLS Configuration options

adam-fowler:
This PR comes from a discussion had on the swift-server with @Lukasa 

I implemented NIOTS TLS support for client certificates, trust roots and applicationProtocols in my mqtt-nio project https://github.com/adam-fowler/mqtt-nio/blob/main/Sources/MQTTNIO/TSTLSConfiguration.swift. I ended up implementing a separate structure from the nio-ssl `TLSConfiguration` as there aren't easy translations to NIOTS structures from `TLSConfiguration`, specifically converting client certificates to a `SecIdentity`. 

@Lukasa was not keen to have NIOTS build an abstraction over the `Network.framework` TLS implementation but he mentioned AHC could make use of something like this.

Changes include
- New structure `TSTLSConfiguration`
- `HTTPClient.Configuration` has a new `init` that takes a `TSTLSConfiguration`
- If no `TSTLSConfiguration` is provided then TLS setup falls back to what we had already
- Added test `HTTPClientNIOTSTests.testHTTPS`. Doesn't look like `HTTPBin` provides support for testing trust roots or client certificates. I have thoroughly tested this code in the MQTTNIO client but don't have any tests for AHC.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -628,6 +628,10 @@ public class HTTPClient {
     public struct Configuration {
         /// TLS configuration, defaults to `TLSConfiguration.forClient()`.
         public var tlsConfiguration: Optional<TLSConfiguration>
+        #if canImport(Network)
+            /// TLS configuration, defaults to `TLSConfiguration.forClient()`.
+            public var tsTlsConfiguration: Optional<TSTLSConfiguration>
+        #endif
         /// Enables following 3xx redirects automatically, defaults to `RedirectConfiguration()`.
         ///
         /// Following redirects are supported:
@@ -658,6 +662,9 @@ public class HTTPClient {
                     ignoreUncleanSSLShutdown: Bool = false,
                     decompression: Decompression = .disabled) {
             self.tlsConfiguration = tlsConfiguration
+            #if canImport(Network)
+                self.tsTlsConfiguration = nil
+            #endif
             self.redirectConfiguration = redirectConfiguration ?? RedirectConfiguration()
             self.timeout = timeout
             self.connectionPool = connectionPool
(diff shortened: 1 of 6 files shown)


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -628,6 +628,10 @@ public class HTTPClient {
     public struct Configuration {
         /// TLS configuration, defaults to `TLSConfiguration.forClient()`.
         public var tlsConfiguration: Optional<TLSConfiguration>
+        #if canImport(Network)
+            /// TLS configuration, defaults to `TLSConfiguration.forClient()`.
```
Lukasa, line 632: This comment looks wrong.

Lukasa (commented):
I'm curious as to why we need a different data structure. It seems like we could shim this under the existing `TLSConfiguration` structure from NIOSSL, couldn't we?

Sources/AsyncHTTPClient/HTTPClient.swift
adam-fowler: > I'm curious as to why we need a different data structure. It seems like we could shim this under the existing `TLSConfiguration` structure from NIOSSL, couldn't we?

We can shim most of it. `SecCertificate` can be created from the DER data of a `NIOSSLCertificate`. `SecIdentity` is the problem case. It requires a p12 or a private key already in the keychain and currently there is no way to generate this from the combination of `NIOSSLPrivateKey` and `NIOSSLCertificate` unless I missed something.

Lukasa: Ah yeah, SecIdentity is rough.

Perhaps this is the moment we encourage AHC to create its own representation of a TLSConfiguration that can abstract over some of these better.

adam-fowler: Can we expose BoringSSL pkcs#12 generation code in NIOSSL? If we have this, we can probably generate a SecIdentity from a TLSConfiguration.

Lukasa: We could, though I don't love doing it.

adam-fowler: If we want a common structure for both NIOSSL and NIOTransportServices TLS setup I can't see any other way around supplying a p12 generation.

Lukasa: Yes, I think I agree. I'd like to know what some other folks prefer here: @artemredkin @ktoso?

artemredkin: I think common structure would be preferable over implementing an additional abstraction over NIOSSL and NIOTransportServices. I wonder, can that structure live in NIOTS?

adam-fowler: @artemredkin I believe with p12 generation from a NIOSSL private key and certificate chain we can use `TLSConfiguration` to cover most situations.

Lukasa: >  I wonder, can that structure live in NIOTS?

I don't really want it to at this early stage.

adam-fowler:
Anymore thoughts on this? I can implement the non `SecIdentity` code for TLSConfiguration already. If the desire is to use `TLSConfiguration` for NIOTS as well.

Lukasa:
@adam-fowler What's your opinion on having AHC define its own TLS configuration structure that abstracts over this mess?

adam-fowler:
> @adam-fowler What's your opinion on having AHC define its own TLS configuration structure that abstracts over this mess?

I think that is something we could aspire to. But that'd be a breaking change and would have to wait for v2.0. I'm not sure what is the full list of parameters from `TLSConfiguration` that apply to AHC. Also how do we support features not available in NIOTS but available in NIOSSL?

In the immediate future we can provide additional support for trust root certificates, application protocols via the TLSConfiguration we already have and if you supply the pkcs#12 generation we can get SecIdentity equivalent of a certificate chain and private key working.

Lukasa:
Hmm, ok. Let's get going with everything but the identity stuff, and we can schedule the PKCS#12 work for SwiftNIO SSL.

adam-fowler:
Closing this as it has been usurped by #350

PR #328, opened 2021-01-19
Add defensive connection closure.

### LINKED ISSUE #324: Connections not being closed, even when server sets the "Connection: close" header
Hi there,

I'm trying to perform some requests to an embedded device in my local network (by IP address). I noticed that after a few requests, the device would stop responding (requests would time out). It looks like the device supports a limited number of concurrent connections / open sockets, so after going over the limit, requests time out. 

The embedded device sets the `Connection: close` header, but looking inside the Active Connections panel in Xcode, I notice that the connections are not being closed:

 [
Screen Shot 2020-12-23 at 21 22 21
](url)

The only workaround I found was to shutdown the `HTTPClient` used to make the request and create a new one for the next request.

I'm running macOS 11, and the issue happens on both release and debug builds. My code is very similar to the sample code on the README (I don't think I'm doing anything that would cause this issue).

Any ideas?

Thanks in advance!

---- 

Environment:
```
- AsyncHTTPClient commit hash: 947429beb4b3f7515bfe2f1d7bb721318cf30935
- Swift version: Apple Swift version 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28) (Version 12.3 (12C33))
- OS: macOS 11.1
```

Lukasa:
Motivation:

Currently when either we or the server send Connection: close, we
correctly do not return that connection to the pool. However, we rely on
the server actually performing the connection closure: we never call
close() ourselves. This is unnecessarily optimistic: a server may
absolutely fail to close this connection. To protect our own file
descriptors, we should make sure that any connection we do not return
the pool is closed.

Modifications:

If we think a connection is closing when we release it, we now call
close() on it defensively.

Result:

We no longer leak connections when the server fails to close them.

Fixes #324.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -352,6 +352,14 @@ class HTTP1ConnectionProvider {
             self.state.release(connection: connection, closing: closing)
         }
 
+        // We close defensively here: we may have failed to actually close on other codepaths,
+        // or we may be expecting the server to close. In either case, we want our FD back, so
+        // we close now to cover our backs. We don't care about the result: if the channel is
+        // _already_ closed, that's fine by us.
+        if closing {
+            connection.close(promise: nil)
+        }
+
         switch action {
         case .none:
             break
--- a/Sources/AsyncHTTPClient/Connection.swift
+++ b/Sources/AsyncHTTPClient/Connection.swift
@@ -72,6 +72,10 @@ extension Connection {
     func close() -> EventLoopFuture<Void> {
         return self.channel.close()
     }
+
+    func close(promise: EventLoopPromise<Void>?) {
+        return self.channel.close(promise: promise)
+    }
 }
 
 /// Methods of Connection which are used in ConnectionsState extracted as protocol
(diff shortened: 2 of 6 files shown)


artemredkin (approved):

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -352,6 +352,15 @@ class HTTP1ConnectionProvider {
             self.state.release(connection: connection, closing: closing)
         }
 
+        // We close defensively here: we may have failed to actually close on other codepaths,
+        // or we may be expecting the server to close. In either case, we want our FD back, so
+        // we close now to cover our backs. We don't care about the result: if the channel is
+        // _already_ closed, that's fine by us.
+        if closing {
+            logger.debug("closing connection")
```
weissi, line 360: I'm totally fine with this but the information here is redundant (because we log it above (`metadata: ["ahc-closing": "\(closing)"]`).

weissi (commented):
Looks good to me, thank you! Left one random comment :).

Sources/AsyncHTTPClient/ConnectionPool.swift
weissi: @Lukasa whilst we're here, do you recall if there is metadata attached to the logger still here? Ie. can this message be correlated with the request that made the close happen? I think that would be good.

Lukasa: Yes, it does: we never remove the metadata key from this logger. We create a new logger when a new request is minted.

PR #336, opened 2021-01-22
fix doc generation setup

tomerd:
motivation: fix doc generation - sourcekitten requires Swift 5.2

changes: expose the doc generation task on the 5.3 docker-compose file so we can trigger it there

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ Package.resolved
 DerivedData
 .DS_Store
 .swiftpm/
+.SourceKitten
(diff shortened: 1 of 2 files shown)


tomerd:
alternatively we can change the script to pin to the older version 

@Lukasa @weissi probably need to check nio as well

ktoso:
Uff yeah, I may have hit this before 🤔 
Will check in other repos

weissi (approved):
Thank you! LGTM

weissi:
> alternatively we can change the script to pin to the older version
> 
> @Lukasa @weissi probably need to check nio as well

@tomerd NIO seems okay, last release is 2.25.1 and the docs are for that very release.

tomerd:
> > alternatively we can change the script to pin to the older version
> 
> > 
> 
> > @Lukasa @weissi probably need to check nio as well
> 
> 
> 
> @tomerd NIO seems okay, last release is 2.25.1 and the docs are for that very release.

looking at the code i think its susceptible to the same issue. it likely using a cached version which is why have not surfaced yet

PR #339, opened 2021-02-09
Update Readme to account for Package.swift format

winsmith:
Adding the product dependency to the target by name only produces an error in Xcode 12.4. Instead, the product dependency should be given as a `.product`. 

This PR updates the README with the new format, so that new user's won't stumble over this.


Lukasa (approved):
Good catch, thanks.

Lukasa:

PR #344, opened 2021-03-03
Fixes bi-directional streaming

### LINKED ISSUE #327: bi-directional streaming broken
AsyncHTTPClient can no longer stream bi-directionally since commit f69b68ffa8e30d0690778f3f67f93505fb970c96 . When being bed the body, it throws a `HTTPClientError.writeAfterRequestSent`.

The problem is this code

```swift
    private func writeBodyPart(context: ChannelHandlerContext, part: IOData, promise: EventLoopPromise<Void>) {
        switch self.state {
        case .idle:
            if let limit = self.expectedBodyLength, self.actualBodyLength + part.readableBytes > limit {
                let error = HTTPClientError.bodyLengthMismatch
                self.state = .endOrError
                self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
                promise.fail(error)
                return
            }
            self.actualBodyLength += part.readableBytes
            context.writeAndFlush(self.wrapOutboundOut(.body(part)), promise: promise)
        default:
            let error = HTTPClientError.writeAfterRequestSent
            self.state = .endOrError
            self.failTaskAndNotifyDelegate(error: error, self.delegate.didReceiveError)
            promise.fail(error)
        }
    }
```

So we only accept bytes for the body in state `.idle` but whilst streaming bi-directionally we can be in state `.body` which makes AHC then wrongly reject the bytes.

Test case:

If you `swift run` the repository at https://github.com/weissi/HTTPBidiStreamingExamples it will work.

If you however edit its `Package.swift` and change

    .package(url: "https://github.com/swift-server/async-http-client.git", .revision("2e6a64abb31fbc2d51da3919b0d8b35a6c7dc0b5")),

to

        .package(url: "https://github.com/swift-server/async-http-client.git", from: "1.2.3")),

then it'll fail. How can I tell if it worked? If you grep for all the output lines that contain `Final result`, what you should see is

```
2021-01-18T21:15:38+0000 info bidi-streaming.driver : driver=client: urlSession, server: swiftNIO Final result: ✅ OK
2021-01-18T21:15:38+0000 info bidi-streaming.driver : driver=client: asyncHTTPClient, server: swiftNIO Final result: ✅ OK
2021-01-18T21:15:38+0000 info bidi-streaming.driver : driver=client: urlSession, server: vapor Final result: ✅ OK
2021-01-18T21:15:38+0000 info bidi-streaming.driver : driver=client: asyncHTTPClient, server: vapor Final result: ✅ OK
```

but if you select the latest AHC, you'll see

```
2021-01-18T21:17:10+0000 info bidi-streaming.driver : driver=client: urlSession, server: swiftNIO Final result: ✅ OK
2021-01-18T21:17:10+0000 error bidi-streaming.driver : driver=client: asyncHTTPClient, server: swiftNIO Final result: ❌ ERROR: HTTPClientError.writeAfterRequestSent
2021-01-18T21:17:10+0000 info bidi-streaming.driver : driver=client: urlSession, server: vapor Final result: ✅ OK
2021-01-18T21:17:10+0000 error bidi-streaming.driver : driver=client: asyncHTTPClient, server: vapor Final result: ❌ ERROR: HTTPClientError.writeAfterRequestSent
```

As you can see: URLSession works and AHC fails against both NIO and Vapor.

artemredkin:
Motivation:
When we stream request body, current implementation expects that body
will finish streaming _before_ we start to receive response body parts.
This is not correct, reponse body parts can start to arrive before we
finish sending the request.

Modifications:
 - Simplifies state machine, we only case about request being fully sent
   to prevent sending body parts after .end, but response state machine
   is mostly ignored and correct flow will be handled by NIOHTTP1
   pipeline
 - Adds HTTPEchoHandler, that replies to each response body part
 - Adds bi-directional streaming test

Result:
Closes #327


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -1025,8 +1019,7 @@ extension TaskHandler: ChannelDuplexHandler {
                 /// Some HTTP Servers can 'forget' to respond with CloseNotify when client is closing connection,
                 /// this could lead to incomplete SSL shutdown. But since request is already processed, we can ignore this error.
                 break
-            case .head where self.ignoreUncleanSSLShutdown,
-                 .body where self.ignoreUncleanSSLShutdown:
+            case .sending where self.ignoreUncleanSSLShutdown:
```
artemredkin, line 1022: test `testNoResponseWithIgnoreErrorForSSLUncleanShutdown` fails if I add `.sent`, I'm not sure why yet...

artemredkin: .sending means that we have not sent client `.end`

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -840,23 +839,20 @@ extension TaskHandler: ChannelDuplexHandler {
             self.writeBody(request: request, context: context)
         }.flatMap {
             context.eventLoop.assertInEventLoop()
+
             if case .endOrError = self.state {
```
weissi, line 845: could we replace this `if case` and the fallthrough by one full `switch self.state`?

It's unclear to me why we can move from almost any state to `.sent` and why all those transitions are legal.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -664,12 +664,10 @@ internal struct TaskCancelEvent {}
 
 internal class TaskHandler<Delegate: HTTPClientResponseDelegate>: RemovableChannelHandler {
     enum State {
-        case idle
-        case bodySent
+        case inactive
+        case sending
```
weissi: I think we could come up with slightly more prescriptive names? What are wen `sending` here? The request, the response, the head, the body, the trailers? Some of them, all of them?

Same for `sent`

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: For example a transition from `.sent` to `.sent` seems odd, right? Or from `.inactive` to `.sent` seems wrong too.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -902,7 +898,7 @@ extension TaskHandler: ChannelDuplexHandler {
 
     private func writeBodyPart(context: ChannelHandlerContext, part: IOData, promise: EventLoopPromise<Void>) {
         switch self.state {
-        case .idle:
+        case .sending:
```
weissi: this seems a little dangerous? Previously we could send in `.idle` and now we can't?

weissi: A full `switch` without the `default` would be better too here.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -1009,10 +1003,10 @@ extension TaskHandler: ChannelDuplexHandler {
 
     func channelInactive(context: ChannelHandlerContext) {
         switch self.state {
+        case .inactive, .sending, .sent, .redirected:
+            self.errorCaught(context: context, error: HTTPClientError.remoteConnectionClosed)
```
weissi, line 1020: hmm, `channelInactive` in `.inactive` (which should probably be named `idle` because the channel isn't "inactive") is really remote connection closed? Isn't that error "between requests" which is expected for a server to do?

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: What's the difference between `.sent` and `.endOrError`? In many cases (`content-length` present) after having sent the response body, we're done?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2708,15 +2708,12 @@ class HTTPClientTests: XCTestCase {
         let task = client.execute(request: request, delegate: TestHTTPDelegate())
 
         XCTAssertThrowsError(try task.wait()) { error in
-            #if os(Linux)
+            if isTestingNIOTS() {
+                XCTAssertEqual(error as? ChannelError, .connectTimeout(.milliseconds(100)))
+            } else {
+                //XCTAssertEqual((error as? IOError).map { $0.errnoCode }, ECONNRESET)
```
weissi: there's some commented code here

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2756,15 +2753,12 @@ class HTTPClientTests: XCTestCase {
         let task = client.execute(request: request, delegate: TestHTTPDelegate())
 
         XCTAssertThrowsError(try task.wait()) { error in
-            #if os(Linux)
+            if isTestingNIOTS() {
+                XCTAssertEqual(error as? ChannelError, .connectTimeout(.milliseconds(200)))
+            } else {
+                //XCTAssertEqual((error as? IOError).map { $0.errnoCode }, ECONNRESET)
```
weissi: commented code

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -664,12 +664,12 @@ internal struct TaskCancelEvent {}
 
 internal class TaskHandler<Delegate: HTTPClientResponseDelegate>: RemovableChannelHandler {
     enum State {
-        case idle
-        case bodySent
-        case sent
-        case head
+        case inactive
+        case sending_waiting
```
weissi, line 668: - we should use `lowerCamelCase` to match swift's usual style.
- I think we can do better than `sendingWaiting`, ie. what are we sending and what are we waiting for?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -664,12 +664,12 @@ internal struct TaskCancelEvent {}
 
 internal class TaskHandler<Delegate: HTTPClientResponseDelegate>: RemovableChannelHandler {
     enum State {
-        case idle
-        case bodySent
-        case sent
-        case head
+        case inactive
```
weissi, line 667: `idle` is probably better than `inactive` because the channel is actually `isActive = true`, right?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
artemredkin: already fixed

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
artemredkin: same here

weissi (commented):
thank you! This looks like a great start, left some comments

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: you are looking at slightly outdated code

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: `.idle` + `channelInactive` means channel was closed _before_ we sent request head, right now we do not re-use the handler between requests (we remove and add it between requests)

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: this is a slightly weird part, it tries to do the following: if `ignoreUncleanSSLShutdown` is set and we received response `.head`, then it's allowed to not fail

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done

Lukasa (approved):
Seems reasonable enough to me, nothing obviously problematic in the code.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -665,11 +668,11 @@ internal struct TaskCancelEvent {}
 internal class TaskHandler<Delegate: HTTPClientResponseDelegate>: RemovableChannelHandler {
     enum State {
         case idle
-        case bodySent
-        case sent
-        case head
+        case sendingBodyWaitingResponseHead
```
weissi, line 671: nice!

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -840,23 +844,33 @@ extension TaskHandler: ChannelDuplexHandler {
...
+            switch self.state {
+            case .idle:
+                // since this code path is called from `write` and write sets state to sendingBody
+                preconditionFailure("should not happen")
+            case .sendingBodyWaitingResponseHead:
+                self.state = .bodySentWaitingResponseHead
+            case .sendingBodyResponseHeadReceived:
+                self.state = .bodySentResponseHeadReceived
+            case .bodySentWaitingResponseHead, .bodySentResponseHeadReceived:
+                preconditionFailure("should not happen")
```
weissi, line 857: should we print `self.state` in that message?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -840,23 +844,33 @@ extension TaskHandler: ChannelDuplexHandler {
...
+            case .endOrError, .redirected:
```
weissi, line 858: mind adding a comment why that's okay in `.endOrError`, or `.redirected`? I don't quite get how we can be in `.end` _before_ we get the end of the body.

weissi: (similar for `.redirected`)

Sources/AsyncHTTPClient/HTTPHandler.swift
weissi: @artemredkin Ahhh! I see, so we put in this handler just before we send `.head` and then we get inactive. That makes sense, thanks!

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: done

Sources/AsyncHTTPClient/HTTPHandler.swift
artemredkin: rewrote, added a test

weissi (approved):
very nice, thank you!

PR #350, opened 2021-03-19
Generate trust roots SecCertificate for Transport Services

### LINKED ISSUE #292: Support Non-Default TrustRoots on Darwin
Currently lots of the TLS options are not supported on Darwin - see https://github.com/swift-server/async-http-client/blob/c9a9bf061d713c91ee9974fa8a6afe413acfd0e9/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift#L115

Rough instructions of how to implement (thanks @Lukasa)

1. Set a verify block using https://developer.apple.com/documentation/security/2976289-sec_protocol_options_set_verify_. 
2. Create a SecTrust from the data you get there: https://github.com/apple/swift-nio-ssl/blob/7932e709af56d27035084ca918bdd729b4084741/Sources/NIOSSL/SecurityFrameworkCertificateVerification.swift#L47
3. Call SecTrustSetAnchorCertificates with the custom roots: https://developer.apple.com/documentation/security/1396098-sectrustsetanchorcertificates
4. Call SecTrustSetAnchorCertificatesOnly with anchorCertificatesOnly: true https://developer.apple.com/documentation/security/1399071-sectrustsetanchorcertificatesonl

adam-fowler:
This PR is a result of another #321.

In that PR I provided an alternative structure to `TLSConfiguration` for when connecting with Transport Services. 

In this one I construct the `NWProtocolTLS.Options` from `TLSConfiguration`. It does mean a little more work for whenever we make a connection, but having spoken to @weissi he doesn't seem to think that is an issue. 

Also there is no method to create a `SecIdentity` at the moment. We need to generate a pkcs#12 from the certificate chain and private key, which can then be used to create the `SecIdentity`.

This should resolve #292

### DIFF
--- a/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
+++ b/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
@@ -60,7 +60,7 @@
         ///
         /// - Parameter queue: Dispatch queue to run `sec_protocol_options_set_verify_block` on.
         /// - Returns: Equivalent NWProtocolTLS Options
-        func getNWProtocolTLSOptions() -> NWProtocolTLS.Options {
+        func getNWProtocolTLSOptions() throws -> NWProtocolTLS.Options {
             let options = NWProtocolTLS.Options()
 
             let useMTELGExplainer = """
@@ -125,12 +125,8 @@
             var secTrustRoots: [SecCertificate]?
             switch trustRoots {
             case .some(.certificates(let certificates)):
-                do {
-                    secTrustRoots = try certificates.compactMap { certificate in
-                        try SecCertificateCreateWithData(nil, Data(certificate.toDERBytes()) as CFData)
-                    }
-                } catch {
-                    // failed to load
+                secTrustRoots = try certificates.compactMap { certificate in
+                    try SecCertificateCreateWithData(nil, Data(certificate.toDERBytes()) as CFData)
                 }
             case .some(.file):
                 preconditionFailure("TLSConfiguration.trustRoots.file is not supported")
--- a/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
+++ b/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
@@ -128,7 +128,7 @@
 
             // the certificate chain
             if self.certificateChain.count > 0 {
-                preconditionFailure("TLSConfiguration.certificateChain is not supported")
+                preconditionFailure("TLSConfiguration.certificateChain is not supported. \(useMTELGExplainer)")
             }
 
             // private key
@@ -173,6 +173,7 @@
                             SecTrustSetAnchorCertificates(trust, trustRootCertificates as CFArray)
                         }
                         if #available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *) {
+                            dispatchPrecondition(condition: .onQueue(Self.tlsDispatchQueue))
                             SecTrustEvaluateAsyncWithError(trust, Self.tlsDispatchQueue) { _, result, error in
                                 if let error = error {
                                     print("Trust failed: \(error.localizedDescription)")
--- a/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
+++ b/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
@@ -128,8 +128,11 @@
                 secTrustRoots = try certificates.compactMap { certificate in
                     try SecCertificateCreateWithData(nil, Data(certificate.toDERBytes()) as CFData)
                 }
-            case .some(.file):
-                preconditionFailure("TLSConfiguration.trustRoots.file is not supported")
+            case .some(.file(let file)):
+                let certificates = try NIOSSLCertificate.fromPEMFile(file)
+                secTrustRoots = try certificates.compactMap { certificate in
+                    try SecCertificateCreateWithData(nil, Data(certificate.toDERBytes()) as CFData)
+                }
 
             case .some(.default), .none:
                 break
--- a/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
+++ b/Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
@@ -16,6 +16,7 @@
 
     import Foundation
     import Network
+    import NIO
     import NIOSSL
     import NIOTransportServices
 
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -15,9 +15,9 @@
 import Foundation
 import NIOHTTP1
 
-extension HTTPClient {
+public extension HTTPClient {
```
weissi, line 18: please don't use any qualifiers before `extension`, that's extremely dangerous because it often makes public API by accident.

weissi (changes requested):
The actual change seems to be lost in reformatting (which also reformats it in the wrong way (eg. moves qualifiers before `extension` instead of the individual members)). Could we undo everything that's not the actual change you'd like to make?

adam-fowler:
> The actual change seems to be lost in reformatting (which also reformats it in the wrong way (eg. moves qualifiers before `extension` instead of the individual members)). Could we undo everything that's not the actual change you'd like to make?

Sorry I'm running a newer version of swift format and submitted all the other files by mistake.

The `public` infront of `extension` is caused by a new rule `extensionAccessControl` which is defaulted to on by default. I have it disabled in soto because its a pain. You could use it to force the positioning of the `public` qualifier through. Do you guys intend to upgrade from v0.44.6 at some point?

Lukasa:

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -110,29 +110,59 @@
...
-                    preconditionFailure("TLSConfiguration.trustRoots != .default is not supported")
+            var secTrustRoots: [SecCertificate]?
+            switch trustRoots {
+            case .some(.certificates(let certificates)):
+                do {
+                    secTrustRoots = try certificates.compactMap { certificate in
+                        try SecCertificateCreateWithData(nil, Data(certificate.toDERBytes()) as CFData)
+                    }
+                } catch {
+                    // failed to load
```
Lukasa, line 121: I think we need some error handling here.

Lukasa (changes requested):
Aside from the error handling situation this looks basically good to me.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
adam-fowler: I stopped catching the error here and set `getNWProtocolTLSOptions()` to `throws`. Is that enough?

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -102,6 +97,11 @@
                 preconditionFailure("TLSConfiguration.keyLogCallback is not supported")
             }
 
+            // the certificate chain
+            if self.certificateChain.count > 0 {
+                preconditionFailure("TLSConfiguration.certificateChain is not supported")
```
weissi, line 102: is crashing appropriate here? I'd imagine that other libraries may be handing through their `TLSConfiguration`s?

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -110,29 +110,55 @@
...
-                guard case .default = trustRoots else {
-                    preconditionFailure("TLSConfiguration.trustRoots != .default is not supported")
+            var secTrustRoots: [SecCertificate]?
+            switch trustRoots {
+            case .some(.certificates(let certificates)):
+                secTrustRoots = try certificates.compactMap { certificate in
+                    try SecCertificateCreateWithData(nil, Data(certificate.toDERBytes()) as CFData)
                 }
+            case .some(.file):
+                preconditionFailure("TLSConfiguration.trustRoots.file is not supported")
```
weissi, line 120: same q here. @Lukasa wdyt?

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -110,29 +110,55 @@
...
+                            sec_protocol_verify_complete(true)
+                            return
+                        }
+
+                        let trust = sec_trust_copy_ref(sec_trust).takeRetainedValue()
+                        if let trustRootCertificates = secTrustRoots {
+                            SecTrustSetAnchorCertificates(trust, trustRootCertificates as CFArray)
+                        }
+                        if #available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *) {
+                            SecTrustEvaluateAsyncWithError(trust, Self.tlsDispatchQueue) { _, result, error in
```
weissi, line 143: @Lukasa wasn't there some weird catch that we need to run this on the right queue or so?

weissi (commented):
happy when @Lukasa is

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
adam-fowler: This is a slightly different case. I could probably support this case if I knew what format the file was in.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
adam-fowler: This was what it was doing before. I'm happy to throw an error instead, but that pretty much amounts to the same thing: the user of AHC has to remove the certificate chain and private key from their `TLSConfiguration`. We certainly cannot fail silently as people will be expecting their certificate chain/private key to be taken into account if we don't provide some error of some sort.

Lukasa: I think crashing is fine here until we add actual support.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
Lukasa: The file format for trust roots is just bundle-o-pem: a collection of PEM-encoded X509 certs in the same file.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
Lukasa: Yup, this must be run on `Self.tlsDispatchQueue`.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
adam-fowler: Now loading PEM info `NIOSSLCertificate` and generating DER for loading into `SecCertificateCreateWithData`. Possibly not the quickest route.

adam-fowler:
@Lukasa do we need some more tests for converting between `TLSConfiguration` and `NWProtocolTLS.Options`?

adam-fowler:
@Lukasa @weissi Anything left to do here?

Lukasa (approved):

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
weissi: @adam-fowler mind sticking a `DispatchPrecondition(.onQueue(Self.tlsDispatchQueue))` in front of this line?

weissi (approved):
Thank you, this looks good to me too! I left one comment (https://github.com/swift-server/async-http-client/pull/350/files#r631742891) which'd be nice to address. Just to be 100% sure we're on the right queue.

weissi (changes requested):
After chatting to @adam-fowler: we're not calling `SecEvaluateTrust` on the right queue at the moment.

adam-fowler:
> After chatting to @adam-fowler: we're not calling `SecEvaluateTrust` on the right queue at the moment.

And then after reviewing code I realised we were. 

Although we did come across another issue where a file load could happen on the EventLoop. To avoid this I have moved the whole `NWProtocolTLS.Options`  generation to the DispatchQueue.

Sources/AsyncHTTPClient/NIOTransportServices/TLSConfiguration.swift
```diff
@@ -109,6 +126,11 @@
                 preconditionFailure("TLSConfiguration.keyLogCallback is not supported. \(useMTELGExplainer)")
             }
 
+            // the certificate chain
+            if self.certificateChain.count > 0 {
+                preconditionFailure("TLSConfiguration.certificateChain is not supported")
```
weissi, line 131:
```suggestion
                preconditionFailure("TLSConfiguration.certificateChain is not supported. \(useMTELGExplainer)")
```

weissi (approved):
Awesome, thank you!

PR #352, opened 2021-03-29
Better backpressure management.

### LINKED ISSUE #348: FileDownloadDelegate can issue out-of-order writes
`FileDownloadDelegate` writes file body parts to the filesystem in the following code:

https://github.com/swift-server/async-http-client/blob/0dda95cffcdf3d96ca551e5701efd1661cf31374/Sources/AsyncHTTPClient/FileDownloadDelegate.swift#L83-L102

This code incorrectly assumes that only one write may be in flight at a given time. While the delegate protocol does exert backpressure, it is entirely possible for multiple `.body` chunks to be triggered on a single `channelRead`, particularly in the case of chunked transfer encoding. The delegate protocol does not protect against this and currently does not promise that this will not happen.

We should either fix the delegate, or fix the delegate protocol to guarantee that only one call to `didReceiveHead`, `didReceiveBodyPart`, and `didFinishRequest` are in flight at any given time. I'm inclined to want to clean up our use of the delegate to make it easier to program against, rather than to just fix the downloader, but I'd like to hear other opinions on the matter, particularly from @weissi and @artemredkin.

Lukasa:
Motivation:

Users of the HTTPClientResponseDelegate expect that the event loop
futures returned from didReceiveHead and didReceiveBodyPart can be used
to exert backpressure. To be fair to them, they somewhat can. However,
the TaskHandler has a bit of a misunderstanding about how NIO
backpressure works, and does not correctly manage the buffer of inbound
data.

The result of this misunderstanding is that multiple calls to
didReceiveBodyPart and didReceiveHead can be outstanding at once. This
would likely lead to severe bugs in most delegates, as they do not
expect it.

We should make things work the way delegate implementers believe it
works.

Modifications:

- Added a buffer to the TaskHandler to avoid delivering data that the
   delegate is not ready for.
- Added a new "pending close" state that keeps track of a state where
   the TaskHandler has received .end but not yet delivered it to the
   delegate. This allows better error management.
- Added some more tests.
- Documented our backpressure commitments.

Result:

Better respect for backpressure.

Resolves #348

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -419,7 +419,26 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
 /// `HTTPClientResponseDelegate` allows an implementation to receive notifications about request processing and to control how response parts are processed.
 /// You can implement this protocol if you need fine-grained control over an HTTP request/response, for example, if you want to inspect the response
 /// headers before deciding whether to accept a response body, or if you want to stream your request body. Pass an instance of your conforming
-/// class to the `HTTPClient.execute()` method and this package will call each delegate method appropriately as the request takes place.
+/// class to the `HTTPClient.execute()` method and this package will call each delegate method appropriately as the request takes place./
+///
+/// ### Backpressure
+///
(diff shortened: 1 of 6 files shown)


artemredkin (approved):

weissi (approved):
thank you! lgtm

PR #358, opened 2021-04-30
Support request specific TLS configuration

madsodgaard:
Adds support for request-specific TLS configuration:
`Request(url: "https://webserver.com", tlsConfiguration: .forClient())`

https://github.com/apple/swift-nio-ssl/pull/280 must be released, before this can be merged.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,8 +22,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.27.0"),
-        .package(url: "https://github.com/apple/swift-nio-ssl.git", .branch("main")),
-//        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.12.0"),
+        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.13.0"),
         .package(url: "https://github.com/apple/swift-nio-extras.git", from: "1.3.0"),
         .package(url: "https://github.com/apple/swift-nio-transport-services.git", from: "1.5.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
--- a/Package.swift
+++ b/Package.swift
@@ -22,7 +22,8 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.27.0"),
-        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.12.0"),
+        .package(url: "https://github.com/apple/swift-nio-ssl.git", .branch("main")),
+//        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.12.0"),
         .package(url: "https://github.com/apple/swift-nio-extras.git", from: "1.3.0"),
         .package(url: "https://github.com/apple/swift-nio-transport-services.git", from: "1.5.1"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -82,7 +82,7 @@ final class ConnectionPool {
             } else {
                 let provider = HTTP1ConnectionProvider(key: key,
                                                        eventLoop: taskEventLoop,
-                                                       configuration: self.configuration,
+                                                       configuration: key.config(overriding: self.configuration),
                                                        pool: self,
                                                        backgroundActivityLogger: self.backgroundActivityLogger)
                 let enqueued = provider.enqueue()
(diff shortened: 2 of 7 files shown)


Lukasa:

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2916,4 +2916,22 @@ class HTTPClientTests: XCTestCase {
             XCTAssertEqual(error as? ExpectedError, .expected)
         }
     }
+    
+    func testRequestSpecificTLS() throws {
```
weissi, line 2920: could we make a second request in this test and check the `RequestInfo` that the test server should return.

The `RequestInfo` has the following fields:

```swift
internal struct RequestInfo: Codable {
    var data: String
    var requestNumber: Int
    var connectionNumber: Int
}
```

so you should be able to write

```swift
        guard let response = maybeResponse, let body = response.body else {
            XCTFail("request failed")
            return
        }
        let connectionNumber = try? JSONDecoder().decode(RequestInfo.self, from: body).connectionNumber
```

and then (in a few test cases) we can validate that:

- two subsequent requests with _different_ `TLSConfiguration`s get different `connectionNumber`s
- two subsequent requests with the _same_ `TLSConfiguration` get the same `connectionNumber` for both requests.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -139,12 +139,16 @@ final class ConnectionPool {
             self.port = request.port
             self.host = request.host
             self.unixPath = request.socketPath
+            if let tls = request.tlsConfiguration {
```
weissi, line 142: @artemredkin should we add a test case that targets the pool directly here? Ie no actual connections?

Package.swift
```diff
@@ -22,7 +22,8 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.27.0"),
-        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.12.0"),
+        .package(url: "https://github.com/apple/swift-nio-ssl.git", .branch("main")),
+//        .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.12.0"),
```
weissi, line 26: once NIOSSL is released, we need to change this back (and up the version number). Just as a reminder.

weissi (changes requested):
Fab, thank you! This looks like a great start!

I left a few comments, mostly requesting changes that we don't accidentally merge the branch dependency.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
madsodgaard: Sure! Did not spot that the we could test it that way, good to know! I added the assertions to the same test, lmk if you meant something else!

weissi:
@madsodgaard hmm, weird 5.0 compiler warning (that we turn into errors):

```
[449/452] Compiling Swift Module 'AsyncHTTPClient' (15 sources)
/code/Sources/AsyncHTTPClient/BestEffortHashableTLSConfiguration.swift:16:18: error: result of call to 'bestEffortEquals' is unused
        lhs.base.bestEffortEquals(rhs.base)
                 ^               ~~~~~~~~~~
Build step 'Execute shell' marked build as failure
```

Also, async-http-client uses automatic format checking with SwiftFormat. If you run SwiftFormat over the source, then the "soundness" part should pass.

madsodgaard:
@weissi Whoops, forgot a return statement 😅

weissi:
@madsodgaard also the API breakage checker detected an API breakage:

```
08:37:54 /* Removed Decls */
08:37:54 Constructor HTTPClient.Request.init(url:method:headers:body:) has been removed
```

Note that in Swift you cannot just add a new parameter (even if it has a default value) without breaking API. Instead of adding the `tlsConfiguration:` parameter with a default value, you'll need to add a new `init` with an extra (non-defaulted) parameter. The existing `HTTPClient.Request.init(url:method:headers:body:)` should then just call the new `init` with the extra parameter.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: thank you!

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -150,17 +150,22 @@ extension NIOClientTCPBootstrap {
         let key = destination
 
         let requiresTLS = key.scheme.requiresTLS
+        // Override optional connection pool configuration.
+        var keyConfiguration = configuration
```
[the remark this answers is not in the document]
Lukasa: Also, should this merge perhaps be done at a higher level, say when we create the `HTTP1ConnectionProvider`? I am a bit nervous about having two separate configs from the perspective of the connection provider: it should always be creating the exact same connection each time.

Package.swift
weissi: @madsodgaard we got them release https://github.com/apple/swift-nio-ssl/releases/tag/2.13.0

Lukasa (approved):
Cool, this LGTM.

weissi (approved):
Fantastic, thank you very much!

PR #360, opened 2021-05-07
add 5.4 docker setup for CI

tomerd:
motivation: 5.4 is out!

changes:  add docker compose setup for ubuntu 20.04 and 5.4 toolchain

PR #361, opened 2021-05-07
add nightly docker setup for CI

tomerd:
motivation: test with nightly toolchain

changes: add docker compose setup for ubuntu 20.04 and nightly toolchain

requires #360


tomerd:
will pass after we merge the Docker fixes in #360

PR #362, opened 2021-05-08
cache NIOSSLContext (saves 27k allocs per conn)

weissi:
Motivation:

At the moment, AHC assumes that creating a `NIOSSLContext` is both cheap
and doesn't block.

Neither of these two assumptions are true.

To create a `NIOSSLContext`, BoringSSL will have to read a lot of
certificates in the trust store (on disk) which require a lot of ASN1
parsing and much much more.

On my Ubuntu test machine, creating one `NIOSSLContext` is about 27,000
allocations!!! To make it worse, AHC allocates a fresh `NIOSSLContext`
for _every single connection_, whether HTTP or HTTPS. Yes, correct.

Modification:

- Cache NIOSSLContexts per TLSConfiguration in a LRU cache
- Don't get an NIOSSLContext for HTTP (plain text) connections

Result:

New connections should be _much_ faster in general assuming that you're
not using a different TLSConfiguration for every connection.


weissi:
### before

49,640 allocations for 100 plain-text HTTP requests over 1 connection

before


### after

22,158 allocations for 100 plain-text HTTP requests over 1 connection

after

Sources/AsyncHTTPClient/SSLContextCache.swift
```diff
@@ -0,0 +1,102 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Logging
+import NIO
+import NIOConcurrencyHelpers
+import NIOSSL
+
+class SSLContextCache {
```
artemredkin, line 20: Why not just create a context in HTTPClient init instead? Currently, there is no per-request TLSConfiguration, it's only set in the client init

Sources/AsyncHTTPClient/LRUCache.swift
```diff
@@ -0,0 +1,106 @@
...
+        var minSoFar: (Index, Element)?
+
+        for indexElement in self.enumerated() {
+            if let min = minSoFar {
+                if try areInIncreasingOrder(indexElement.1, min.1) {
+                    minSoFar = indexElement
+                }
+            } else {
+                minSoFar = indexElement
+            }
```
Lukasa, line 101: Can I propose we drop this branch from the loop? This branch only needs to be taken once, at loop start, and then we don't ever take it again, so we could do it as loop setup instead (e.g. `guard var minSoFar = self.first.map { (0, $0) }`)

Sources/AsyncHTTPClient/LRUCache.swift
```diff
@@ -0,0 +1,106 @@
...
+    private var elements: [Element]
+
+    init(capacity: Int = 8) {
+        self.capacity = capacity
+        self.elements = []
+        self.elements.reserveCapacity(capacity)
+    }
+
+    private mutating func findIndex(key: Key) -> Int? {
+        self.generation += 1
```
Lukasa, line 35: It's weird to me that the `+=` is in here instead of in `find`. This would let us avoid making this `mutating`.

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
```diff
@@ -13,13 +13,14 @@
 //===----------------------------------------------------------------------===//
 
 #if canImport(Network)
```
Lukasa, line 15: Why did we move all of this?

Sources/AsyncHTTPClient/SSLContextCache.swift
```diff
@@ -0,0 +1,102 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5: Let's fix up the copyright header.

Sources/AsyncHTTPClient/SSLContextCache.swift
artemredkin: because it's blocking?

Sources/AsyncHTTPClient/LRUCache.swift
```diff
@@ -0,0 +1,106 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+struct LRUCache<Key: Equatable & Hashable,
```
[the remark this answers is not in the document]
weissi: Maybe? The license matches so they can pick it up if they like it.

Lukasa (changes requested):

Sources/AsyncHTTPClient/LRUCache.swift
weissi: `append` needs it too and `append` only uses `findIndex` because it needs the index rather than the element.

Sources/AsyncHTTPClient/SSLContextCache.swift
```diff
@@ -0,0 +1,102 @@
...
+                self.state = .active
+                self.threadPool.start()
+                return nil
+            case .active:
+                return nil
+            case .shutDown:
+                struct SSLContextCacheShutdownError: Error {}
+                return eventLoop.makeFailedFuture(SSLContextCacheShutdownError())
+            }
+        }
```
fabianfett, line 71: Returning an optional future from this block seems odd to me. Do we maybe want to return an optional `Error` and use that to fail a future outside the lock?

Sources/AsyncHTTPClient/SSLContextCache.swift
```diff
@@ -0,0 +1,102 @@
...
+class SSLContextCache {
+    private var state = State.activeNoThread
+    private let lock = Lock()
+    private var sslContextCache = LRUCache<BestEffortHashableTLSConfiguration, NIOSSLContext>()
+    private let threadPool = NIOThreadPool(numberOfThreads: 1)
+
+    enum State {
+        case activeNoThread
+        case active
+        case shutDown
```
fabianfett, line 29: Just out of interest, when do we use `shutdown` and when `shutDown`?

Sources/AsyncHTTPClient/SSLContextCache.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    init() {}
+
```
fabianfett, line 33: Maybe we should add a `deinit` that ensures that we are indeed `shutDown`? In case someone removes the call to `shutdown` by accident sometime in the future, this would bubble up.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -77,149 +78,207 @@ extension NIOClientTCPBootstrap {
...
+                                                       eventLoop: eventLoop,
+                                                       logger: logger)
+                    .flatMapThrowing { sslContext in
+                        try clientBootstrap.makeClientTCPBootstrap(host: host,
+                                                                   sslContext: sslContext,
+                                                                   useProxy: configuration.proxy != nil)
+                    }
             } else {
                 preconditionFailure("Cannot create bootstrap for the supplied EventLoop")
             }
```
fabianfett, line 124: Since this is very much the same code as L101-111, do you think we can reorganize this into something like this:

```swift
func actuallyCreateBootstrap() -> EventLoopFuture<NIOClientTCPBootstrap> {
    #if canImport(Network)
        if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *), let tsBootstrap = NIOTSConnectionBootstrap(validatingGroup: eventLoop) {
            return nioTSBootstrap
        }
        // if we can not create a nioTSBootstrap just fallthrough
    #endif

    guard let clientBootstrap = ClientBootstrap(validatingGroup: eventLoop) else {
        preconditionFailure("Cannot create bootstrap for the supplied EventLoop")
    }

    return nioBootstrap
}
```

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -77,149 +78,207 @@ extension NIOClientTCPBootstrap {
...
-                case .disabled:
-                    ()
-                case .enabled(let limit):
-                    let decompressHandler = NIOHTTPResponseDecompressor(limit: limit)
-                    try syncOperations.addHandler(decompressHandler)
-                }
+        // If we got here and we had a TLSEventsHandler in the pipeline, we can remove it ow.
+        if requiresTLS {
+            channel.pipeline.removeHandler(name: TLSEventsHandler.handlerName, promise: nil)
+        }
```
fabianfett, line 265: I think we should make this sync as well. If we want to ignore the error go with: `try?`, but I doubt that we would like to ignore the error.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -230,3 +289,17 @@ extension Connection {
...
+        #if canImport(Network)
+            if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *) {
+                return self.underlyingBootstrap is NIOTSConnectionBootstrap
+            } else {
+                return false
+            }
+        #else
+            return false
+        #endif
+    }
```
fabianfett, line 304: Is there a reason, why we don't use this pattern?
```suggestion
    var isNIOTS: Bool {
        #if canImport(Network)
            if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *) {
                return self.underlyingBootstrap is NIOTSConnectionBootstrap
            }
        #endif
        return false
    }
```

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -77,149 +78,207 @@ extension NIOClientTCPBootstrap {
...
             }
     }
 
-    static func makeHTTP1Channel(destination: ConnectionPool.Key, eventLoop: EventLoop, configuration: HTTPClient.Configuration, preference: HTTPClient.EventLoopPreference) -> EventLoopFuture<Channel> {
+    static func makeHTTP1Channel(destination: ConnectionPool.Key,
+                                 eventLoop: EventLoop,
+                                 configuration: HTTPClient.Configuration,
+                                 sslContextCache: SSLContextCache,
+                                 preference: HTTPClient.EventLoopPreference,
+                                 logger: Logger) -> EventLoopFuture<Channel> {
```
fabianfett, line 183: Since this method is the only one which is invoked from the outside, do we maybe want to move this to the top of this extension?

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -77,149 +78,207 @@ extension NIOClientTCPBootstrap {
...
+        }
     }
 
     static func makeHTTPClientBootstrapBase(
         on eventLoop: EventLoop,
         host: String,
         port: Int,
         requiresTLS: Bool,
-        configuration: HTTPClient.Configuration
-    ) throws -> NIOClientTCPBootstrap {
```
fabianfett, line 127: Method is not invoked from outside. Maybe make this `private`?

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -77,149 +78,207 @@ extension NIOClientTCPBootstrap {
...
-            channel = bootstrap.connect(host: address.host, port: address.port)
-        case .unix, .http_unix, .https_unix:
-            channel = bootstrap.connect(unixDomainSocketPath: key.unixPath)
-        }
+        let bootstrap = NIOClientTCPBootstrap.makeHTTPClientBootstrapBase(on: channelEventLoop,
+                                                                          host: key.host,
+                                                                          port: key.port,
+                                                                          requiresTLS: requiresTLS,
+                                                                          configuration: configuration,
+                                                                          sslContextCache: sslContextCache,
```
fabianfett, line 202: Why do we pass the `SSLContextCache` down here? We already have a `sslContext` (a `Future`) created before...

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -77,149 +78,207 @@ extension NIOClientTCPBootstrap {
...
-                handshakeFuture = handshakePromise.futureResult
-            } else if requiresTLS {
-                do {
-                    handshakeFuture = try channel.pipeline.syncOperations.handler(type: TLSEventsHandler.self).completionPromise.futureResult
-                } catch {
-                    return channel.eventLoop.makeFailedFuture(error)
+            return channel.flatMap { channel -> EventLoopFuture<(Channel, NIOSSLContext?)> in
+                sslContext.map { sslContext -> (Channel, NIOSSLContext?) in
+                    (channel, sslContext)
+                }
```
fabianfett, line 217: Why do we "wait" on the `SSLContext` here?

Sources/AsyncHTTPClient/SSLContextCache.swift
weissi: The problem is that in English, it can be "shutdown" (noun), "shut down", and "shut-down", eg. "Let's initiate the shutdown.", "This is a shut-down factory.", or "The factory is shut down.". I was aiming for "The cache is shut down" and because it's two words, I made it `.shutDown` :). But honestly, this totally doesn't matter and we can also make it `.shutdown` or `.closed`, I don't mind either way.

Sources/AsyncHTTPClient/SSLContextCache.swift
weissi: pretty sure it's already there?

```
    deinit {
         assert(self.state == .shutDown)
     }
```

Sources/AsyncHTTPClient/Utils.swift
weissi: You can't synchronously remove a handler. The handler is on control of removals (through the `RemovableChannelHandler` protocol and they may be async.

Sources/AsyncHTTPClient/Utils.swift
weissi: I reduced the amount of code we need here but happy to add another `} else { return false }`

Sources/AsyncHTTPClient/Utils.swift
weissi: I can but it'd make the diff larger and much harder to review to change. Up to you and the other reviewers, usually I try to keep the diff small if reasonably possible.

Sources/AsyncHTTPClient/Utils.swift
weissi: Because we now need the actual context. We're kicking off the connection creation and an SSLContext creation in parallel and this is the point where we need both.

Sources/AsyncHTTPClient/Utils.swift
weissi: That is an excellent question, I have no idea. I'll look into this, maybe we can just inject the SSLContext into that method.

Sources/AsyncHTTPClient/LRUCache.swift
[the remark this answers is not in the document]
weissi: Sure, I prefer to mention both (makes it easier to read I think) but also happy to remove the `Equatable`. I don't actually use the `Hashable` at this point in time, so can also remove that one. Up to you.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -77,149 +78,207 @@ extension NIOClientTCPBootstrap {
...
+            }.flatMap { channel, sslContext in
+                configureChannelPipeline(channel,
+                                         isNIOTS: bootstrap.isNIOTS,
+                                         sslContext: sslContext,
+                                         configuration: configuration,
+                                         key: key)
```
fabianfett, line 223: NIT: Do we maybe collapse this into one closure?

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -77,149 +78,207 @@ extension NIOClientTCPBootstrap {
...
                     let parameters = tlsConfiguration.getNWProtocolTLSOptions()
                     let tlsProvider = NIOTSClientTLSProvider(tlsOptions: parameters)
-                    bootstrap = NIOClientTCPBootstrap(tsBootstrap, tls: tlsProvider)
+                    bootstrap = eventLoop.makeSucceededFuture(NIOClientTCPBootstrap(tsBootstrap, tls: tlsProvider))
                 }
             } else if let clientBootstrap = ClientBootstrap(validatingGroup: eventLoop) {
-                bootstrap = try clientBootstrap.makeClientTCPBootstrap(host: host, requiresTLS: requiresTLS, configuration: configuration)
+                bootstrap = sslContextCache.sslContext(tlsConfiguration: configuration.tlsConfiguration ?? .forClient(),
+                                                       eventLoop: eventLoop,
+                                                       logger: logger)
```
fabianfett, line 103: This is hit with in a test that doesn't require `https`: `testUploadStreamingBackpressure`

Sources/AsyncHTTPClient/LRUCache.swift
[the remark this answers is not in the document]
weissi: removed `Hashable`.

Sources/AsyncHTTPClient/Utils.swift
weissi: great catch! Thanks you

Sources/AsyncHTTPClient/Utils.swift
fabianfett: I agree with the general goal. However the `Utils.swift` has exploded for me so much, that I have checked out your branch anyway and only come here, to post my comments.

Sources/AsyncHTTPClient/LRUCache.swift
Lukasa: Sure, but `append` can do the `+=` itself as well. It just seems weird for this method to have that side-effect.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -52,174 +53,198 @@ public final class HTTPClientCopyingDelegate: HTTPClientResponseDelegate {
...
+                                                                           tls: NIOInsecureNoTLS()))
+            } else {
+                return sslContextCache.sslContext(tlsConfiguration: tlsConfiguration,
+                                                  eventLoop: eventLoop,
+                                                  logger: logger)
+                    .flatMapThrowing { sslContext in
+                        let hostname = (host.isIPAddress || host.isEmpty) ? nil : host
+                        let tlsProvider = try NIOSSLClientTLSProvider<ClientBootstrap>(context: sslContext, serverHostname: hostname)
+                        return NIOClientTCPBootstrap(clientBootstrap, tls: tlsProvider)
+                    }
```
fabianfett, line 193: Just ran the tests on macOS and noticed that this code path is never hit.

weissi: @fabianfett yes, expected. macOS is using NIOTS.

Sources/AsyncHTTPClient/LRUCache.swift
weissi: Fair! I'll rename it.

Sources/AsyncHTTPClient/Utils.swift
weissi: it's now at the top.

Sources/AsyncHTTPClient/Utils.swift
weissi: which closures do you want me to collapse?

Sources/AsyncHTTPClient/Utils.swift
weissi: and thanks, added a test case which will trigger this code even on macOS so we don't rely on Linux.

Lukasa (approved):
Cool, this LGTM.

Sources/AsyncHTTPClient/Utils.swift
```diff
@@ -52,174 +53,207 @@ public final class HTTPClientCopyingDelegate: HTTPClientResponseDelegate {
...
+            return channel.flatMap { channel -> EventLoopFuture<(Channel, NIOSSLContext?)> in
+                sslContext.map { sslContext -> (Channel, NIOSSLContext?) in
+                    (channel, sslContext)
+                }
+            }.flatMap { channel, sslContext in
+                configureChannelPipeline(channel,
+                                         isNIOTS: bootstrap.isNIOTS,
+                                         sslContext: sslContext,
+                                         configuration: configuration,
+                                         key: key)
```
fabianfett, line 108: Nevermind. I thought we can collapse the `sslContext.map` into the following `flatMap`, but that doesn't work, since the `sslContext` is a `EventLoopFuture<NIOSSLContext?>` first and actual `NIOSSLContext?` later.

Sources/AsyncHTTPClient/Utils.swift
fabianfett: Nevermind. I thought we can collapse the `sslContext.map` into the following `flatMap`, but that doesn't work, since the `sslContext` is a `EventLoopFuture<NIOSSLContext?>` first and actual `NIOSSLContext?` later.

fabianfett (approved):
LGTM. 🎉

PR #368, opened 2021-05-13
SSLContextCache: use DispatchQueue instead of NIOThreadPool

weissi:
Motivation:

In the vast majority of cases, we'll only ever create one and only one
`NIOSSLContext`. It's therefore wasteful to keep around a whole thread
doing nothing just for that. A `DispatchQueue` is absolutely fine here.

Modification:

Run the `NIOSSLContext` creation on a `DispatchQueue` instead.

Result:

Fewer threads hanging around.

### DIFF
--- a/Sources/AsyncHTTPClient/SSLContextCache.swift
+++ b/Sources/AsyncHTTPClient/SSLContextCache.swift
@@ -12,69 +12,22 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Dispatch
 import Logging
 import NIO
 import NIOConcurrencyHelpers
 import NIOSSL
 
 class SSLContextCache {
-    private var state = State.activeNoThread
     private let lock = Lock()
     private var sslContextCache = LRUCache<BestEffortHashableTLSConfiguration, NIOSSLContext>()
-    private let threadPool = NIOThreadPool(numberOfThreads: 1)
-
-    enum State {
-        case activeNoThread
-        case active
-        case shutDown
-    }
-
-    init() {}
-
-    func shutdown() {
-        self.lock.withLock { () -> Void in
-            switch self.state {
-            case .activeNoThread:
-                self.state = .shutDown
-            case .active:
-                self.state = .shutDown
-                self.threadPool.shutdownGracefully { maybeError in
-                    precondition(maybeError == nil, "\(maybeError!)")
-                }
-            case .shutDown:
-                preconditionFailure("SSLContextCache shut down twice")
-            }
-        }
-    }
-
-    deinit {
-        assert(self.state == .shutDown)
-    }
+    private let offloadQueue = DispatchQueue(label: "io.github.swift-server.AsyncHTTPClient.SSLContextCache")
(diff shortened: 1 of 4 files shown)


Lukasa (approved):

Sources/AsyncHTTPClient/SSLContextCache.swift
```diff
@@ -12,19 +12,19 @@
...
 import NIOConcurrencyHelpers
 import NIOSSL
 
 class SSLContextCache {
-    private var state = State.activeNoThread
+    private var state = State.active
     private let lock = Lock()
     private var sslContextCache = LRUCache<BestEffortHashableTLSConfiguration, NIOSSLContext>()
-    private let threadPool = NIOThreadPool(numberOfThreads: 1)
+    private let threadPool = DispatchQueue(label: "io.github.swift-server.AsyncHTTPClient.SSLContextCache")
```
PeterAdams-A, line 25: Maybe rename the variable?

Sources/AsyncHTTPClient/SSLContextCache.swift
```diff
@@ -88,17 +79,20 @@ extension SSLContextCache {
 
         logger.debug("creating new SSL context",
                      metadata: ["ahc-tls-config": "\(tlsConfiguration)"])
-        let newSSLContext = self.threadPool.runIfActive(eventLoop: eventLoop) {
-            try NIOSSLContext(configuration: tlsConfiguration)
+        let promise = eventLoop.makePromise(of: NIOSSLContext.self)
+        self.threadPool.async {
```
glbrntt, line 83: I think NIO has API to do this for you: `threadPool.asyncWithFuture`

Sources/AsyncHTTPClient/SSLContextCache.swift
fabianfett, line 25: I'm not sure if I approve staying with the name `threadPool`. In this context threadPool implies a NIOThreadPool to me and I get high blood pressure if I don't immediately find a `threadPool.shutdown` somewhere in the file.

fabianfett: Suggestion `offloadQueue`?

weissi: done, thanks

Sources/AsyncHTTPClient/SSLContextCache.swift
fabianfett: It's not a threadPool :P The name tricked you.

glbrntt: I kept the `threadPool` name to trick you into thinking I thought it was a `NIOThreadPool` ;) 

https://github.com/apple/swift-nio/blob/d161bf658780b209c185994528e7e24376cf7283/Sources/NIO/DispatchQueue%2BWithFuture.swift#L17-L34

glbrntt (approved):

PR #370, opened 2021-05-27
publish security process

weissi:
The updated [SSWG security requirements](https://github.com/swift-server/sswg/blob/main/security/package-requirements.md) require that listed projects have a `SECURITY.md` detailing their security process.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,44 @@
+# Security
+
+This document specifies the security process for the AsyncHTTPClient project.
+
+## Disclosures
+
+### Private Disclosure Process
+
+The AsyncHTTPClient maintainers ask that known and suspected vulnerabilities be
+privately and responsibly disclosed by emailing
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -0,0 +1,45 @@
...
+[sswg-security-reports@forums.swift.org](mailto:sswg-security-reports@forums.swift.org)
+with the all the required detail.
+**Do not file a public issue.**
+
+#### When to report a vulnerability
+
+* You think you have discovered a potential security vulnerability in
+  AsyncHTTPClient.
+* You are unsure how a vulnerability affects AsyncHTTPClient.
+  projects.
```
Lukasa, line 20: This word "projects" seems wrong.

Lukasa (approved):
Clean up the typo and we should be good.

SECURITY.md
weissi: good catch, thanks, fixed.

glbrntt (approved):

PR #372, opened 2021-05-29
Update CODE_OF_CONDUCT.md

fabianfett:
When we updated to Code of Conduct 1.4, we may have forgotten AHC.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,55 +1,38 @@
 # Code of Conduct
-To be a truly great community, AsyncHTTPClient needs to welcome developers from all walks of life,
-with different backgrounds, and with a wide range of experience. A diverse and friendly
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -1,55 +1,38 @@
 # Code of Conduct
```
tomerd, line 1: speaking with @tkremenek and @DougGregor they suggested we simply link (make the content of this file into a single link) from all our packages to https://swift.org/code-of-conduct/ which is the canonical CoC for the Swift project. I think this is a good idea, WDYT?

cc @weissi @Lukasa @ktoso @yim-lee

Lukasa: Sure, WFM.

weissi: WFM

PR #374, opened 2021-06-16
Make DelayOnHeadDelegate resilient

fabianfett:
Currently the `DelayOnHeadDelegate` test utility depends on correct timings. Right now, if a request is slower than 50ms in  `testErrorAfterCloseWhileBackpressureExerted` the test will fail, since the backpressure promise is failed, before a head was received. This pr fixes this, by giving the user a callback, when the head was received.


Davidde94 (approved):

fabianfett:

Lukasa (approved):

fabianfett:
@Lukasa I just noticed: This only changes test code, but got `patch-version-bump-only`. For this reason I think we can go with label `needs-no-version-bump`, which we don't have in AHC at the moment. (Shall I create it?)

Lukasa:
Sure. I used patch version only because I was too lazy to create the label. :wink:

PR #375, opened 2021-06-16
Implement SOCKS proxy functionality

Davidde94:
Add a new Proxy type to enable a `HTTPClient` to send requests via a SOCKSv5 Proxy server.

PR #335, opened 2021-01-22
Address flakiness of testSSLHandshakeErrorPropagation

Lukasa:
Motivation:

Flaky tests are bad.

This test is flaky because the server closes the connection immediately
upon channelActive. In practice this can mean that the handshake never
even gets a chance to start: by the time the SSLHandler ends up
in the pipeline the connection is already dead. Heck, by the time we
attempt to complete the connection the connection might be dead.

Modifications:

- Change the shutdown to be on first read.
- Remove the disabled autoRead.
- Change the expected NIOTS failure mode to connectTimeout,
    which is how this manifests in NIOTS.

Result:

Test is no longer flaky.

PR #377, opened 2021-06-16
Refactor Channel creation

fabianfett:
This is the first cherry pick from #376.

- The connection creation logic has been refactored into a number of smaller methods that can be combined for the configuration. 
- Connection creation now has a logical home. We move it from `Utils.swift` into a `ConnectionFactory`
- We have explicit `ChannelHandlers` that we use for connection creation:
  - `TLSEventsHandler` got its own file and unit tests
  - `HTTP1ProxyConnectHandler` got its own file and unit tests
- Some small things are already part of this pr that will get their context later. For example: 
  - `HTTPConnectionPool` is added as a namespace to not cause major renames in follow up PRs
  - `HTTPConnectionPool.Connection.ID` and its generator were added now. (This will be used later to identify a connection during its lifetime)
  - the file `HTTPConnectionPool+Manager` was added to give `HTTPConnectionPool.Connection.ID.Generator` already its final destination.

Looking forward to your feedback.
I'm well aware that this PR is in a race with #375. I'm happy to update it once #375 has landed.


fabianfett:
Ran into a test failure in Swift 5.1: `testSSLHandshakeErrorPropagation failed`.

Looks like this one used to be flaky. But I assume  @Lukasa fixed it with #335. Investigated the test's flakyness with 100 iterations in Xcode (all passed). Will need further investigation on Linux.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,143 @@
...
+        switch self.state {
+        case .failed, .completed:
+            break
+        case .initialized, .connectSent, .headReceived:
+            preconditionFailure("Removing the handler, while connecting seems wrong")
+        }
+    }
+
+    func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        preconditionFailure("We don't support outgoing traffic during HTTP Proxy update.")
```
artemredkin, line 64: just want to make sure - clients will not be able to trigger this precondition no matter what configuration they use?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+            let decoder = ByteToMessageHandler(HTTPResponseDecoder(leftOverBytesStrategy: .dropBytes))
+            let proxyHandler = HTTP1ProxyConnectHandler(
+                targetHost: self.key.host,
+                targetPort: self.key.port,
+                proxyAuthorization: proxy.authorization,
+                connectPromise: connectPromise
+            )
+
+            do {
+                try channel.pipeline.syncOperations.addHandler(encoder)
```
artemredkin, line 106: why not add all handler at once?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    private func makePlainBootstrap(eventLoop: EventLoop) -> NIOClientTCPBootstrapProtocol {
+        #if canImport(Network)
+            if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *), let tsBootstrap = NIOTSConnectionBootstrap(validatingGroup: eventLoop) {
+                return tsBootstrap
+                    .addTimeoutIfNeeded(self.clientConfiguration.timeout)
+                    .channelInitializer { channel in
+                        do {
+                            try channel.pipeline.syncOperations.addHandler(HTTPClient.NWErrorHandler())
```
artemredkin, line 171: wouldn't async operations look cleaner here?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Manager.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIOConcurrencyHelpers
+
+extension HTTPConnectionPool.Connection.ID {
+    static var globalGenerator = Generator()
+
+    struct Generator {
+        private let atomic: NIOAtomic<Int>
```
artemredkin, line 21: maybe we can make it `UInt64`, just in case someone will be running it until our sun dies 😆

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
artemredkin: was it copied form the old handler? if yes, then ignore my comment

artemredkin (approved):
Some minor suggestions, but otherwise looks good 👍

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -234,6 +240,13 @@ class HTTP1ConnectionProvider {
         self.closePromise = eventLoop.makePromise()
         self.state = .init(eventLoop: eventLoop)
         self.backgroundActivityLogger = backgroundActivityLogger
+
+        self.factory = HTTPConnectionPool.ConnectionFactory(
+            key: self.key,
+            tlsConfiguration: tlsConfiguration ?? configuration.tlsConfiguration ?? .forClient(),
```
Davidde94, line 246: Minor cleanup, purely personal opinion, I'd refactor out this line into a `let`

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,143 @@
...
+
+    enum State {
+        case initialized(EventLoopPromise<Void>)
+        case connectSent(EventLoopPromise<Void>)
+        case headReceived(EventLoopPromise<Void>)
+        case failed(Error)
+        case completed
+    }
+
+    private var state: State
```
Davidde94, line 31: Do we not want to pull this out into a self-contained state machine?

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,143 @@
...
+    let proxyAuthorization: HTTPClient.Authorization?
+
+    init(targetHost: String,
+         targetPort: Int,
+         proxyAuthorization: HTTPClient.Authorization?,
+         connectPromise: EventLoopPromise<Void>) {
+        self.targetHost = targetHost
+        self.targetPort = targetPort
+        self.proxyAuthorization = proxyAuthorization
+
```
Davidde94, line 44:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,143 @@
...
+        self.state = .initialized(connectPromise)
+    }
+
+    func handlerAdded(context: ChannelHandlerContext) {
+        precondition(context.channel.isActive, "Expected to be added to an active channel")
+
```
Davidde94, line 50:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
Davidde94: Would it not be a good idea to buffer the data?

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,143 @@
...
+                self.state = .completed
+                promise.succeed(())
+            case .failed:
+                // ran into an error before... ignore this one
+                break
+            case .initialized, .connectSent, .completed:
+                preconditionFailure("Invalid state")
+            }
+        }
+    }
```
Davidde94, line 121: Again, personal opinion, this code is perhaps a little more difficult to read than it needs to be. Could you refactor out into separate functions to handle each http part?

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,143 @@
...
+
+    func sendConnect(context: ChannelHandlerContext) {
+        guard case .initialized(let promise) = self.state else {
+            preconditionFailure("Invalid state")
+        }
+
```
Davidde94, line 127:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+    import NIOTransportServices
+#endif
+
+extension HTTPConnectionPool {
+    enum NegotiatedProtocol {
+        case http1_1(Channel)
+        case http2_0(Channel)
+    }
+
+    final class ConnectionFactory {
```
Davidde94, line 30: Does this need to be a `class`?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+            self.key = key
+            self.clientConfiguration = clientConfiguration
+            self.sslContextCache = sslContextCache
+            self.tlsConfiguration = tlsConfiguration ?? clientConfiguration.tlsConfiguration ?? .forClient()
+        }
+    }
+}
+
+extension HTTPConnectionPool.ConnectionFactory {
+    func makeBestChannel(connectionID: HTTPConnectionPool.Connection.ID, eventLoop: EventLoop, logger: Logger) -> EventLoopFuture<(Channel, HTTPVersion)> {
```
Davidde94, line 49: What does `makeBestChannel` mean?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    private func makeChannel(eventLoop: EventLoop, logger: Logger) -> EventLoopFuture<(Channel, HTTPVersion)> {
+        switch self.key.scheme {
+        case .http, .http_unix, .unix:
+            return self.makePlainChannel(eventLoop: eventLoop).map { ($0, .http1_1) }
+        case .https, .https_unix:
+            return self.makeTLSChannel(eventLoop: eventLoop, logger: logger).map {
+                (channel, negotiated) -> (Channel, HTTPVersion) in
+                let version = negotiated == "h2" ? HTTPVersion.http2 : HTTPVersion.http1_1
```
Davidde94, line 64: Something feels off about using a string `h2` to identify the channel type here. would an enum be better? Or maybe we could switch `negotiated` to account for `H2` or something?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+                return (channel, version)
+            }
+        }
+    }
+
+    private func makePlainChannel(eventLoop: EventLoop) -> EventLoopFuture<Channel> {
+        let bootstrap = self.makePlainBootstrap(eventLoop: eventLoop)
+
```
Davidde94, line 72:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+        eventLoop: EventLoop,
+        logger: Logger
+    ) -> EventLoopFuture<(Channel, HTTPVersion)> {
+        // A proxy connection starts with a plain text connection to the proxy server. After
+        // the connection has been established with the proxy server, the connection might be
+        // upgraded to TLS before we send our first request.
+        let bootstrap = self.makePlainBootstrap(eventLoop: eventLoop)
+        return bootstrap.connect(host: proxy.host, port: proxy.port).flatMap { channel in
+            let connectPromise = channel.eventLoop.makePromise(of: Void.self)
+
```
Davidde94, line 95:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Davidde94: +1

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+            do {
+                try channel.pipeline.syncOperations.addHandler(encoder)
+                try channel.pipeline.syncOperations.addHandler(decoder)
+                try channel.pipeline.syncOperations.addHandler(proxyHandler)
+            } catch {
+                return channel.eventLoop.makeFailedFuture(error)
+            }
+
+            return connectPromise.futureResult.flatMap {
+                channel.pipeline.removeHandler(proxyHandler).flatMap {
```
Davidde94, line 114: Do we not want to do this synchronously?

I know it's nothing to do with this PR, but we should probably support multiple simultaneous removal.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
fabianfett: Because this way we would need to alloc an array. And if possible, we want to save this alloc.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+        #endif
+
+        return channelFuture
+    }
+
+    private func makeTLSBootstrap(eventLoop: EventLoop, logger: Logger)
+        -> EventLoopFuture<NIOClientTCPBootstrapProtocol> {
+        // since we can support h2, we need to advertise this in alpn
+        var tlsConfig = self.tlsConfiguration
+        tlsConfig.applicationProtocols = ["http/1.1" /* , "h2" */ ]
```
Davidde94, line 224: Just to confirm, h2 is coming in another PR? (apologies if you've said this somewhere and I missed it)

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+            tlsConfiguration: tlsConfig,
+            eventLoop: eventLoop,
+            logger: logger
+        )
+
+        let bootstrap = ClientBootstrap(group: eventLoop)
+            .addTimeoutIfNeeded(self.clientConfiguration.timeout)
+            .channelInitializer { channel in
+                sslContextFuture.flatMap { (sslContext) -> EventLoopFuture<Void> in
+                    let sync = channel.pipeline.syncOperations
```
Davidde94, line 262: Tbh it's probably neater to just move this inside the `do`

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,314 @@
...
+        }
+        return self.connectTimeout(connectTimeamount)
+    }
+}
+
+private extension String {
+    var isIPAddress: Bool {
+        var ipv4Addr = in_addr()
+        var ipv6Addr = in6_addr()
+
```
Davidde94, line 308:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
fabianfett: > just want to make sure - clients will not be able to trigger this precondition no matter what configuration they use?

Yes. We have basically two modes for a channel. Setup and Running. The channel is only returned to the user once it is running. (meaning 100% setup). If someone writes to the `ProxyChannelHandler` this is def. wrong :)

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
```diff
@@ -0,0 +1,56 @@
...
+            return .init(host: host, port: port, authorization: nil)
+        }
+
+        /// Create proxy.
+        ///
+        /// - parameters:
+        ///     - host: proxy server host.
+        ///     - port: proxy server port.
+        ///     - authorization: proxy server authorization.
+        public static func server(host: String, port: Int, authorization: HTTPClient.Authorization? = nil) -> Proxy {
```
Davidde94, line 52: I might be completely wrong, but shouldn't the SOCKS stuff be here?

Tests/AsyncHTTPClientTests/HTTP1ProxyConnectHandlerTests.swift
```diff
@@ -0,0 +1,205 @@
...
+import NIO
+import NIOHTTP1
+import XCTest
+
+class HTTP1ProxyConnectHandlerTests: XCTestCase {
+    func testProxyConnectWithoutAuthorizationSuccess() {
+        let embedded = EmbeddedChannel()
+        defer { XCTAssertNoThrow(try embedded.finish(acceptAlreadyClosed: false)) }
+
+        let socketAddress = try! SocketAddress(ipAddress: "127.0.0.1", port: 3000)
```
Davidde94, line 25: Can we make this `localhost`, as it's possible to have IPv6-only systems

Davidde94: And everywhere else, I won't spam you saying the same thing repeatedly haha!

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
fabianfett: It would look nicer, I agree. 

However with the current impl. we don't need to alloc an `EventLoopFuture` at all, since we can return the prealloced `return channel.eventLoop.makeSucceededFuture(())`.

This would not necessarily be the case when adding the handlers async.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
fabianfett: Sadly `removeHandler` always has to be async, based on how it works today. 
https://github.com/apple/swift-nio/blob/f7918a80f16cecc06d9980f65019300fdeb59e77/Sources/NIO/ChannelHandler.swift#L320

The channel handler get's a callback with a removalToken.

```
func removeHandler(context: ChannelHandlerContext, removalToken: ChannelHandlerContext.RemovalToken)
```

The handler might stay in the pipeline for longer and call 

```
context.leavePipeline(removalToken: removalToken)
```

at a later point. For this reason, handler removals always have to be async. :(

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
fabianfett: Yes, it is coming in one of the follow up PRs (there will be a lot of them)

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
fabianfett: Not merged yet :) Comming...

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
fabianfett: > Would it not be a good idea to buffer the data?

@Davidde94 We would develop a feature that we don't use in this case. If we were to put this into `NIOExtras`, I would agree. In AHC it would be purely dead code.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1ProxyConnectHandler.swift
fabianfett: I'm not sure if this is necessary here? @glbrntt @Lukasa opinions?

glbrntt: I think either is fine, this looks like a fairly simple state machine. However, with the lack of explicit state machine I think it would be useful to document each state `case` and their valid transitions.

Sources/AsyncHTTPClient/ConnectionPool.swift
fabianfett: This is totally unneeded here. Removed :)

Lukasa (changes requested):

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,355 @@
...
+            tlsConfig.applicationProtocols = ["http/1.1" /* , "h2" */ ]
+            let tlsEventHandler = TLSEventsHandler()
+
+            let sslContextFuture = self.sslContextCache.sslContext(
+                tlsConfiguration: tlsConfig,
+                eventLoop: channel.eventLoop,
+                logger: logger
+            )
+
+            return sslContextFuture.flatMap { sslContext -> EventLoopFuture<String?> in
```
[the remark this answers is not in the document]
fabianfett: well. the SSLContextCache creates the future with the eventLoop we pass in (L176). Further all syncOperations assert on the correct eventLoop. Do you still want me to add a precondition here?

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SOCKSEventsHandler.swift
```diff
@@ -0,0 +1,117 @@
...
+        // transitions to socksEstablished or failed
+        case channelActive(Scheduled<Void>)
+        // final success state
+        case socksEstablished
+        // final success state
+        case failed(Error)
+    }
+
+    private var socksEstablishedPromise: EventLoopPromise<Void>?
+    var socksEstablishedFuture: EventLoopFuture<Void>! {
```
Davidde94, line 33: SSWG rule

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SOCKSEventsHandler.swift
```diff
@@ -0,0 +1,117 @@
...
+        self.socksEstablishedPromise = context.eventLoop.makePromise(of: Void.self)
+
+        if context.channel.isActive {
+            self.connectionStarted(context: context)
+        }
+    }
+
+    func handlerRemoved(context: ChannelHandlerContext) {
+        struct NoResult: Error {}
+        self.socksEstablishedPromise!.fail(NoResult())
```
Davidde94, line 54: How can we make sure that the right error is thrown if the error is private to a function?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Davidde94: Same question

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Davidde94: Yep this still applies @fabianfett

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Davidde94: Makes sense.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Davidde94: @Lukasa any opinion?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,356 @@
...
+        eventLoop: EventLoop,
+        logger: Logger
+    ) -> EventLoopFuture<(Channel, HTTPVersion)> {
+        // A proxy connection starts with a plain text connection to the proxy server. After
+        // the connection has been established with the proxy server, the connection might be
+        // upgraded to TLS before we send our first request.
+        let bootstrap = self.makePlainBootstrap(eventLoop: eventLoop)
+        return bootstrap.connect(host: proxy.host, port: proxy.port).flatMap { channel in
+            let socksConnectHandler = SOCKSClientHandler(targetAddress: .domain(self.key.host, port: self.key.port))
+            let socksEventHandler = SOCKSEventsHandler(deadline: .now() + .seconds(10))
```
Davidde94, line 141: I'd recommend adding a small comment re why 10 seconds?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,356 @@
...
+                                try channel.pipeline.syncOperations.addHandler(TLSEventsHandler(deadline: nil))
+                                return channel.eventLoop.makeSucceededFuture(())
+                            } catch {
+                                return channel.eventLoop.makeFailedFuture(error)
+                            }
+                        } as NIOClientTCPBootstrapProtocol
+                }
+                return bootstrapFuture
+            }
+        #endif
```
Davidde94, line 289: This section is kind of rough to read, could it be refactored a little to have less indentation?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,356 @@
...
+        }
+    }
+}
+
+extension NIOClientTCPBootstrapProtocol {
+    func addTimeoutIfNeeded(_ config: HTTPClient.Configuration.Timeout?) -> Self {
+        guard let connectTimeamount = config?.connect else {
+            return self
+        }
+
```
Davidde94, line 341:
```suggestion

```

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
```diff
@@ -0,0 +1,85 @@
...
+            case http(HTTPClient.Authorization?)
+            case socks
+        }
+
+        /// Specifies Proxy server host.
+        public var host: String
+        /// Specifies Proxy server port.
+        public var port: Int
+        /// Specifies Proxy server authorization.
+        public var authorization: HTTPClient.Authorization? {
```
Davidde94, line 40: Can we specific specifically that this is HTTP authorization only?

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
```diff
@@ -0,0 +1,85 @@
...
+        public static func server(host: String, port: Int) -> Proxy {
+            return .init(host: host, port: port, type: .http(nil))
+        }
+
+        /// Create a HTTP proxy.
+        ///
+        /// - parameters:
+        ///     - host: proxy server host.
+        ///     - port: proxy server port.
+        ///     - authorization: proxy server authorization.
```
Davidde94, line 72: Can we specific specifically that this is HTTP authorization only?

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SOCKSEventsHandler.swift
Davidde94: Also I don't think this is actually safe as the underlying promise is a "real" optional.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SOCKSEventsHandler.swift
fabianfett: This error is only really used, if a developer removes the Handler from the pipeline before a success or failure has happened (This might happen in HappyEyeballs for example). Promises must be fulfilled, before they can be released (otherwise crash!). To ensure we don't crash, this trick is used here. This is part of the dirty tricks in the transport handshake state to somewhere outside the pipeline. I'm not sure if we want to test for this error. Since this is all internal API only, I would think we can leave it as is.

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
fabianfett: This is only moved from a previous file (HTTPClientProxyHandler). Since we don't want to change public API at all, the only adjustments we can make here is in documentation.

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
fabianfett: This is only moved from a previous file (HTTPClientProxyHandler). Since we don't want to change public API at all, the only adjustments we can make here is in documentation.

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
Davidde94: Yeh this is what I mean - previously there was a comment

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
Davidde94: Yeh this is what I mean - previously there was a comment

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,355 @@
...
+            tlsConfiguration: tlsConfig,
+            eventLoop: eventLoop,
+            logger: logger
+        )
+
+        let bootstrap = ClientBootstrap(group: eventLoop)
+            .addTimeoutIfNeeded(self.clientConfiguration.timeout)
+            .channelInitializer { channel in
+                sslContextFuture.flatMap { (sslContext) -> EventLoopFuture<Void> in
+                    do {
```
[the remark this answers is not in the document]
fabianfett: well. the SSLContextCache creates the future with the eventLoop we pass in (L319). Further all syncOperations assert on the correct eventLoop. Do you still want me to add a precondition here?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
fabianfett: Refactored the magic numbers into their own struct:
```swift
        struct Timeouts {
            /// the `TimeAmount` waited before the TCP connection creation is failed with a timeout error
            var connect: TimeAmount = .seconds(10)

            /// the `TimeAmount` waited before the SOCKS proxy connection creation is failed with a timeout error
            var socksProxyHandshake: TimeAmount = .seconds(10)

            /// the `TimeAmount` waited before the HTTP proxy connection creation is failed with a timeout error
            var httpProxyHandshake: TimeAmount = .seconds(10)

            /// the `TimeAmount` waited before we the TLS handshake is failed with a timeout error
            var tlsHandshake: TimeAmount = .seconds(10)
        }
```
The 10 seconds are totally arbitrary. We need a timeout to not fill up our ConnectionPool with starting connections that never come up. Eventually we could make those publicly changeable.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,204 @@
...
+        // final success state
+        case completed
+    }
+
+    private var state: State = .initialized
+
+    let targetHost: String
+    let targetPort: Int
+    let proxyAuthorization: HTTPClient.Authorization?
+    let deadline: NIODeadline
```
Lukasa, line 41: Is there any reason for these not to be private?

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,204 @@
...
+            return
+        }
+
+        let timeout = context.eventLoop.scheduleTask(deadline: self.deadline) {
+            switch self.state {
+            case .initialized:
+                preconditionFailure("How can we have a scheduled timeout, if the connection is not even up?")
+
+            case .connectSent(let scheduled), .headReceived(let scheduled):
+                scheduled.cancel()
```
Lukasa, line 178: We can omit the call to `cancel` here, we know the scheduled task fired if we're in this function!

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Lukasa: In general I'm inclined to make things values when they can be.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Lukasa: I'm happy to do these one at a time, as-is currently implemented.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,386 @@
...
+
+            /// the `TimeAmount` waited before the SOCKS proxy connection creation is failed with a timeout error
+            var socksProxyHandshake: TimeAmount = .seconds(10)
+
+            /// the `TimeAmount` waited before the HTTP proxy connection creation is failed with a timeout error
+            var httpProxyHandshake: TimeAmount = .seconds(10)
+
+            /// the `TimeAmount` waited before we the TLS handshake is failed with a timeout error
+            var tlsHandshake: TimeAmount = .seconds(10)
+        }
```
Lukasa, line 44: Hmm, this is why I don't like timeouts. Is it intentional that we would allow up to 30s of delay (10s for TCP, 10s for HTTP proxy, 10s for TLS)? It seems like it would be much clearer if we just had a hard deadline to hit instead, irregardless of what protocols we had to run.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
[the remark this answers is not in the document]
Lukasa: Ah, ok. Fine then.

Tests/AsyncHTTPClientTests/HTTPConnectionPool+FactoryTests.swift
```diff
@@ -0,0 +1,119 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,204 @@
...
+         deadline: NIODeadline) {
+        self.targetHost = targetHost
+        self.targetPort = targetPort
+        self.proxyAuthorization = proxyAuthorization
+        self.deadline = deadline
+    }
+
+    func handlerAdded(context: ChannelHandlerContext) {
+        self.proxyEstablishedPromise = context.eventLoop.makePromise(of: Void.self)
+
```
Davidde94, line 60:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,204 @@
...
+        if context.channel.isActive {
+            self.sendConnect(context: context)
```
Davidde94, line 62: If the channel is already active, won't `channelActive` have already been called, meaning `sendConnect` will have been called? If that's the case, this whole `if` is redundant as it should probably be guarded in the `sendConnect` function.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,204 @@
...
+            preconditionFailure("How can we receive a channelInactive before a channelActive?")
+        case .connectSent(let timeout), .headReceived(let timeout):
+            timeout.cancel()
+            let error = HTTPClientError.remoteConnectionClosed
+            self.state = .failed(error)
+            self.proxyEstablishedPromise?.fail(error)
+            context.fireErrorCaught(error)
+        case .failed:
+            break
+        case .completed:
```
Davidde94, line 93: Nit, match `failed` and `completed` together

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,204 @@
...
+
+            case .failed:
+                // ran into an error before... ignore this one
+                break
+            case .completed, .connectSent, .initialized:
+                preconditionFailure("Invalid state: \(self.state)")
+            }
+
+        case .end:
+            switch self.state {
```
Davidde94, line 151: I still think it would be really beneficial to break this big-boy up a little... Interested to see if @Lukasa agrees

(this isn't a blocker in any way)

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Davidde94: We can save a Heap allocation by making it a struct. It probably doesn't matter, but I'd say it makes sense to use a `struct` in this case.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,386 @@
...
+
+        init(key: ConnectionPool.Key,
+             tlsConfiguration: TLSConfiguration?,
+             clientConfiguration: HTTPClient.Configuration,
+             sslContextCache: SSLContextCache,
+             timeouts: Timeouts = .init()) {
+            self.key = key
+            self.clientConfiguration = clientConfiguration
+            self.sslContextCache = sslContextCache
+            self.tlsConfiguration = tlsConfiguration ?? clientConfiguration.tlsConfiguration ?? .forClient()
```
Davidde94, line 60: So if we have both a `tlsconfigurataion` and `clientConfiguration.tlsConfiguration`, we drop the client configuration on the floor? Why? Might be useful to have a comment

Tests/AsyncHTTPClientTests/SOCKSEventsHandlerTests+XCTest.swift
```diff
@@ -0,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
Davidde94, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```

Tests/AsyncHTTPClientTests/TLSEventsHandlerTests+XCTest.swift
```diff
@@ -0,0 +1,34 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
Davidde94, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
Lukasa: No. This is `handlerAdded`: if the channel is active, we're being added _after_ `channelActive` fired, and so we'll never see it.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
Lukasa: Yeah, probably a good idea.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Lukasa: This prefers _either_ an explicit `tlsConfiguration`, or if that's not available, one from `clientConfiguration`, or if there isn't one there either it constructs a plain one.

Tests/AsyncHTTPClientTests/TLSEventsHandlerTests+XCTest.swift
Lukasa: This is an autogenerated file, and the header shouldn't change.

Tests/AsyncHTTPClientTests/SOCKSEventsHandlerTests+XCTest.swift
Lukasa: This is an autogenerated file, and so the header shouldn't be changed.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
Davidde94: Yeh, I'm saying we can remove the `if` because it will always be true.

Lukasa: It won't always be true: we might be added _before_ `channelActive` is called, in which case this will be `false`.

Nonetheless I agree that the state machine is already policing this.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
fabianfett: We capture this factory in _a lot_ of EventLoopFuture `flatMap`s. Yes we save one heap allocation per connection pool, but we will start to copy around four config properties for every closure. Of course we could encapsulate the config in a storage class to offset the copying... But is this really worth here?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1052,4 +973,16 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
...
+    public static let connectTimeout = HTTPClientError(code: .connectTimeout)
+    /// The socks handshake timed out.
+    public static let socksHandshakeTimeout = HTTPClientError(code: .socksHandshakeTimeout)
+    /// The http proxy connection creation timed out.
+    public static let httpProxyHandshakeTimeout = HTTPClientError(code: .httpProxyHandshakeTimeout)
+    /// The tls handshake timed out.
+    public static let tlsHandshakeTimeout = HTTPClientError(code: .tlsHandshakeTimeout)
+    /// The remote server only offered an unsupported application protocol
+    public static func serverOfferedUnsupportedApplicationProtocol(_ proto: String) -> HTTPClientError {
+        return HTTPClientError(code: .serverOfferedUnsupportedApplicationProtocol(proto))
```
fabianfett, line 986: This has become a patch that requires a minor.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
Lukasa: I think this represents a premature optimisation. So far I've only seen a few flatMaps where this is the _only_ object captured (i.e. where we'd remove an allocation by having this be a `class`). We can reduce the copying if needed by extracting to temporaries and only closing over the specific thing we need.

In general I care much more about getting semantics right than about initial performance numbers. As you say, we can always refactor to a class-backed struct if we want the performance profile of a `class`. The question is, conceptually, _should_ this object be a value? In general I'm inclined to make objects that should be values into values.

fabianfett: I agree. Thanks for the explanation. Fixed.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
artemredkin: nit: then we should probably return `eventLoop.makeSucceededVoidFuture()`

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
Davidde94: !

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    func channelInactive(context: ChannelHandlerContext) {
+        switch self.state {
+        case .initialized:
+            preconditionFailure("How can we receive a channelInactive before a channelActive?")
+        case .connectSent(let timeout), .headReceived(let timeout):
+            timeout.cancel()
+            self.failWithError(HTTPClientError.remoteConnectionClosed, context: context, closeConnection: false)
+
```
Davidde94, line 86:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        let timeout = context.eventLoop.scheduleTask(deadline: self.deadline) {
+            switch self.state {
+            case .initialized:
+                preconditionFailure("How can we have a scheduled timeout, if the connection is not even up?")
+
+            case .connectSent, .headReceived:
+                self.failWithError(HTTPClientError.httpProxyHandshakeTimeout, context: context)
+
```
Davidde94, line 120:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+        guard case .initialized = self.state else {
+            // we might run into this handler twice, once in handlerAdded and once in channelActive.
+            return
+        }
+
+        let timeout = context.eventLoop.scheduleTask(deadline: self.deadline) {
+            switch self.state {
+            case .initialized:
+                preconditionFailure("How can we have a scheduled timeout, if the connection is not even up?")
+
```
Davidde94, line 117:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+        switch head.status.code {
+        case 200..<300:
+            // Any 2xx (Successful) response indicates that the sender (and all
+            // inbound proxies) will switch to tunnel mode immediately after the
+            // blank line that concludes the successful response's header section
+            self.state = .headReceived(scheduled)
+        case 407:
+            self.failWithError(HTTPClientError.proxyAuthenticationRequired, context: context)
+
```
Davidde94, line 154:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }
+
+    private func handleHTTPEndReceived(context: ChannelHandlerContext) {
+        switch self.state {
+        case .headReceived(let timeout):
+            timeout.cancel()
+            self.state = .completed
+            self.proxyEstablishedPromise?.succeed(())
+
```
Davidde94, line 182:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SOCKSEventsHandler.swift
```diff
@@ -0,0 +1,117 @@
...
+    private let deadline: NIODeadline
+    private var state: State = .initialized
+
+    init(deadline: NIODeadline) {
+        self.deadline = deadline
+    }
+
+    func handlerAdded(context: ChannelHandlerContext) {
+        self.socksEstablishedPromise = context.eventLoop.makePromise(of: Void.self)
+
```
Davidde94, line 46:
```suggestion

```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -0,0 +1,404 @@
...
+        eventLoop: EventLoop,
+        logger: Logger
+    ) -> EventLoopFuture<(Channel, HTTPVersion)> {
+        switch self.key.scheme {
+        case .http, .http_unix, .unix:
+            return self.makePlainChannel(deadline: deadline, eventLoop: eventLoop).map { ($0, .http1_1) }
+        case .https, .https_unix:
+            return self.makeTLSChannel(deadline: deadline, eventLoop: eventLoop, logger: logger).flatMapThrowing {
+                channel, negotiated in
+
```
Davidde94, line 103:
```suggestion

```

Davidde94 (approved):

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -440,12 +453,31 @@ class HTTP1ConnectionProvider {
...
                              logger: Logger) -> EventLoopFuture<Channel> {
-        return NIOClientTCPBootstrap.makeHTTP1Channel(destination: self.key,
-                                                      eventLoop: self.eventLoop,
-                                                      configuration: self.configuration,
-                                                      sslContextCache: self.pool.sslContextCache,
-                                                      preference: preference,
-                                                      logger: logger)
+        let connectionID = HTTPConnectionPool.Connection.ID.globalGenerator.next()
+        let eventLoop = preference.bestEventLoop ?? self.eventLoop
+        let deadline = NIODeadline.now() + (self.configuration.timeout.connect ?? .seconds(10))
```
Lukasa, line 458: This hardcoded default is a bit wild. I'd rather we didn't put it here but instead modified the behaviour of the timeout structure.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -440,12 +453,31 @@ class HTTP1ConnectionProvider {
...
+
+            switch self.configuration.decompression {
+            case .disabled:
+                ()
+            case .enabled(let limit):
+                let decompressHandler = NIOHTTPResponseDecompressor(limit: limit)
+                try syncOperations.addHandler(decompressHandler)
+            }
+
+            return channel
```
Lukasa, line 479: I'm a bit surprised to find this function here. Why is it being placed here, as opposed to being the responsibility of the factory object?

fabianfett: Because this is functionality that is tied to the "current" implementation. This code will shall be removed as soon as we have the real deal lands. To keep the traffic in the factory low, I put it here. Happy to move though if you want me to.

Lukasa: I think I'd like to keep it logically in the right place, even though I acknowledge it's only temporary.

fabianfett: Fixed.

Sources/AsyncHTTPClient/ConnectionPool.swift
fabianfett: Fixed!

Lukasa (approved):
LGTM.

PR #379, opened 2021-06-21
Fix test for misbehaving server

Davidde94:
The misbehaving SOCKS server test was previously passing, but not for the right reason. The invalid data we were sending was never making it off the server thanks to the SOCKSServerHandshakeHandler.

In fact, as long as SOCKSServerHandshakeHandler is present, it isn't possible to send invalid data. So to test the client now we have to add a completely new handler, which exists only to write some nonsense bytes.

The test still passes, but for the right reason, and we now assert the type of error the client throws too.

### DIFF
--- a/Tests/AsyncHTTPClientTests/SOCKSTestUtils.swift
+++ b/Tests/AsyncHTTPClientTests/SOCKSTestUtils.swift
@@ -24,7 +24,7 @@ struct MockSOCKSError: Error, Hashable {
 
 class TestSOCKSBadServerHandler: ChannelInboundHandler {
     typealias InboundIn = ByteBuffer
-    
+
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
         // just write some nonsense bytes
         let buffer = context.channel.allocator.buffer(bytes: [0xAA, 0xBB, 0xCC, 0xDD, 0xEE])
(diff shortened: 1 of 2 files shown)


Tests/AsyncHTTPClientTests/SOCKSTestUtils.swift
```diff
@@ -124,7 +139,16 @@ class TestHTTPServer: ChannelInboundHandler {
...
 
     func errorCaught(context: ChannelHandlerContext, error: Error) {
+//        if self.misbehave {
+//            // We've landed here because of a state error
+//            // which we want when we're misbehaving
+//            // so in this case ignore the error, and send forward
+//            // some invalid socks bytes.
+//            let buffer = context.channel.allocator.buffer(bytes: [0xFF, 0xFF, 0xFF, 0xFF])
+//            context.writeAndFlush(self.wrapOutboundOut(buffer), promise: nil)
+//        } else {
```
fabianfett, line 149: Remove?

Tests/AsyncHTTPClientTests/SOCKSTestUtils.swift
```diff
@@ -86,6 +97,10 @@ class SOCKSTestHandler: ChannelInboundHandler, RemovableChannelHandler {
             }
         }
     }
+
+    func errorCaught(context: ChannelHandlerContext, error: Error) {
+        context.fireErrorCaught(error)
+    }
```
fabianfett, line 103: I don't think we need this.

fabianfett (commented):
Thanks! Looks good. Let's remove the dead code.

Tests/AsyncHTTPClientTests/SOCKSTestUtils.swift
Davidde94: Oh, oops

fabianfett (approved):
Thanks

PR #380, opened 2021-06-23
Make `testSSLHandshakeErrorPropagation` less flaky

fabianfett:
### Motivation

- The `testSSLHandshakeErrorPropagation` has a very short connect timeout (100ms). The reason for this is, that a failing TLS handshake manifest in NIOTS as a connect timeout and we don’t want to wait for the timeout for to long in our tests.
- When using `NIOSSL` however, we expect a proper `NIOSSLError`. Sometimes though the connection is not setup fast enough (because of the very short 100ms connect timeout), that we get a `connectTimeout` error instead. See build logs: https://ci.swiftserver.group/job/async-http-client-swift52-prb/480/console
    ```
    Test Case 'HTTPClientTests.testSSLHandshakeErrorPropagation' started at 2021-06-22 19:14:49.426
    /code/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:2735: error: 
    HTTPClientTests.testSSLHandshakeErrorPropagation : failed - Handshake failed with unexpected error: connectTimeout(NIO.TimeAmount(nanoseconds: 100000000))
    ```

### Modifications

- Setting the short connect timeout to 100ms when using NIOTS only. This will make sure we get the proper NIOSSLError when using NIOSSL

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -2715,9 +2715,14 @@ class HTTPClientTests: XCTestCase {
             XCTAssertNoThrow(try server.close().wait())
         }
 
-        // We set the connect timeout down very low here because on NIOTS this manifests as a connect
-        // timeout.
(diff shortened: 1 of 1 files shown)


Davidde94 (approved):
Nice!

Issue #371, opened 2021-05-27
HTTPClientTests.testRaceNewRequestsVsShutdown flaky (crash in `assert(self.pool.count == 0)`)


weissi:
Failed in CI on 5.0 job https://ci.swiftserver.group/job/async-http-client-swift50-prb/884/console

```
Test Case 'HTTPClientTests.testRaceNewRequestsVsShutdown' started at 2021-05-27 13:09:25.562
Assertion failed: file /code/Sources/AsyncHTTPClient/HTTPClient.swift, line 117
Current stack trace:
0    libswiftCore.so                    0x00007f7d115b7830 _swift_stdlib_reportFatalErrorInFile + 115
1    libswiftCore.so                    0x00007f7d114f294c <unavailable> + 3463500
2    libswiftCore.so                    0x00007f7d114f2a3e <unavailable> + 3463742
3    libswiftCore.so                    0x00007f7d112ee189 <unavailable> + 1347977
4    libswiftCore.so                    0x00007f7d114c6a32 <unavailable> + 3283506
5    libswiftCore.so                    0x00007f7d112ed569 <unavailable> + 1344873
6    async-http-clientPackageTests.xctest 0x000055a276b5bc1b <unavailable> + 838683
```

this is distinct but probably related to #290 . The line in question is line 117 in `HTTPClient.swift`.

```

    deinit {
        assert(self.pool.count == 0)   <<<<<< THIS ONE
        assert(self.state == .shutDown, "Client not shut down before the deinit. Please call client.syncShutdown() when no longer needed.")
    }
```

weissi:
CC @artemredkin  / @fabianfett

fabianfett:
https://ci.swiftserver.group/job/async-http-client-swift53-prb/215/console

fabianfett:
Implementation replaced in #427. Not seen another instance of this in CI for now. If this comes up again it should be a new issue. (property count does not exist on connection pool anymore).

PR #384, opened 2021-06-28
Add HTTPScheduledRequest and HTTPExecutingRequest

fabianfett:
#### Motivation

- In the future we don't want to modify the `ChannelPipeline` for each request.
- Our current API supports generic Response types: `HTTPClient.Task<Response>`.
  - For this reason our current `TaskHandler<Delegate: HTTPClientResponseDelegate>` is generic over the request's delegate.
- If we don't want to modify our `ChannelPipeline` for each request in the future we need remove the generic response types somehow. The only way to achieve this is by boxing our current API in a protocol, that we use as an existential.

#### Changes

- Introducing three protocols:
  - `HTTPClientTask` (I hate this name. Suggestions are highly welcome. Especially since the potential conflict with the current type `HTTPClient.Task<Response>`): An abstraction of a request, that handles back-pressure for sending the request body as well as receiving the response body.
  - `HTTPTaskExecutor` a protocol that can be used from the `HTTPClientTask` abstracting away functionality that will normally be implemented by a `ChannelHandler`
  - `HTTPTaskQueuer` a protocol that can be used from the `HTTPClientTask` abstracting away functionality that will normally be implemented by a `ConnectionPool`
- An implementation of the `HTTPClientTask` called `RequestBag`. This shows how we can support our current API, with the new `HTTPClientTask` protocol

#### Notes for reviewers:

- We should first agree on the three new protocols and then progress to the `RequestBag` implementation. I propose those changes not as two PRs to show the usage of those protocols right away.
- This code is not triggered by any existing component. It purely exist to allow follow up prs and will enabled eventually.
- For a new `async`/`await` compatible API, we would add a new implement that conforms to `HTTPClientTask`


Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,843 @@
...
+            case .executing(let executor, let requestState, let responseState):
+                switch requestState {
+                case .initialized:
+                    preconditionFailure("Request stream must be started before it can be paused")
+                case .producing:
+                    return .write(part, executor, self.task.eventLoop.makeSucceededFuture(()))
+
+                case .paused(var promises):
+                    // backpressure is signaled to the writer using unfulfilled futures. let's
+                    // create a new one for this write
```
Lukasa, line 304: Why do we need to create new ones each time, instead of using a single promise?

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,843 @@
...
+
+                case .paused(let promises):
+                    switch result {
+                    case .success:
+                        self._state = .executing(executor, .finished, responseState)
+                        return .forwardStreamFinished(executor, promises)
+                    case .failure(let error):
+                        self._state = .finished(error: error)
+                        return .forwardStreamFailureAndFailTask(executor, error, promises)
+                    }
```
Lukasa, line 384: Several of these compute on unwrapped `requestState`. Should they be `mutating func`s on the request state object itself?

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,843 @@
...
+            writer.cancelRequest(self)
+            promises?.forEach { $0.fail(error) }
+            self.failTask(error)
+        }
+    }
+
+    // MARK: Request delegate calls
+
+    func didSendRequestHead(_ head: HTTPRequestHead) {
+        guard self.task.eventLoop.inEventLoop else {
```
Lukasa, line 415: Recommend continuing the `if` pattern instead of the `guard` one.

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,843 @@
...
+            return self.task.eventLoop.execute {
+                self.didSendRequestHead(head)
+            }
+        }
+
+        self.delegate.didSendRequestHead(task: self.task, head)
+    }
+
+    func didSendRequestPart(_ part: IOData) {
+        guard self.task.eventLoop.inEventLoop else {
```
Lukasa, line 425: Same here.

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,843 @@
...
+            return self.task.eventLoop.execute {
+                self.didSendRequestPart(part)
+            }
+        }
+
+        self.delegate.didSendRequestPart(task: self.task, part)
+    }
+
+    func didSendRequest() {
+        guard self.task.eventLoop.inEventLoop else {
```
Lukasa, line 435: Same here.

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,843 @@
...
+            }
+        }
+
+        self.delegate.didSendRequest(task: self.task)
+    }
+
+    func failTask(_ error: Error) {
+        self.task.promise.fail(error)
+
+        guard self.task.eventLoop.inEventLoop else {
```
Lukasa, line 447: Same here.

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,843 @@
...
+            return self.task.eventLoop.execute {
+                self.delegate.didReceiveError(task: self.task, error)
+            }
+        }
+
+        self.delegate.didReceiveError(task: self.task, error)
```
Lukasa, line 453: Do you want to fail the promise first, or after this?

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,843 @@
...
+        }
+
+        queuer?.cancelRequest(self)
+        executor?.cancelRequest(self)
+
+        if forward {
+            self.failTask(error)
+        }
+    }
+}
```
Lukasa, line 837: I think this object would get _substantially_ cleaner if the request/response mini-state machines were codified and made explicit. This would shift much of the functionality out, which would make our lives a good deal easier.

Sources/AsyncHTTPClient/ConnectionPool/HTTPExecutingRequest.swift
```diff
@@ -0,0 +1,250 @@
...
+///
+///   1. The request executor will call `willExecuteRequest(_: HTTPRequestExecutor) -> Bool` on the
+///     request. The request is expected to return `true`, if it wants to be executed and false if
+///     it was cancelled or failed, before this checkpoint. If the request returns `true` it is
+///     expected to keep a reference to the `HTTPRequestExecutor` that was passed to the request
+///     for further communication. If the request returned false, the executor shall inform the
+///     scheduling mechanism that it is ready to execute the a new request.
+///   2. The request sending is started by the executor accessing the `HTTPExecutingRequest`'s
+///     property `requestHead: HTTPRequestHead`. Based on the `requestHead` the executor can
+///     determine if the request has a body (Is a "content-length" or "transfer-encoding"
```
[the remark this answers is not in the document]
fabianfett: We expect that the ChannelHandler will call `HTTPHeaders.validate(method: HTTPMethod, body: HTTPClient.Body?)` to ensure the headers are set correctly. For reference:

https://github.com/swift-server/async-http-client/blob/main/Sources/AsyncHTTPClient/RequestValidation.swift

Sources/AsyncHTTPClient/ConnectionPool/HTTPExecutingRequest.swift
```diff
@@ -0,0 +1,249 @@
...
+/// queuer and executor. The client's methods will be called synchronously on an `EventLoop` by the
+/// executor. For this reason it is very important that the implementation of these functions never blocks.
+protocol HTTPScheduledRequest: AnyObject {
+    /// The task's logger
+    var logger: Logger { get }
+
+    /// A connection to run this task on needs to be found before this deadline!
+    var connectionDeadline: NIODeadline { get }
+
+    /// The task's eventLoopPreference
```
glbrntt, line 152:
```suggestion
    /// The task's `EventLoop` preference
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPExecutingRequest.swift
```diff
@@ -0,0 +1,249 @@
...
+}
+
+protocol HTTPExecutingRequest: AnyObject {
+    /// The request's head.
+    ///
+    /// Based on the content of the request head the task executor will call `startRequestBodyStream`
+    /// after `requestHeadSent` was called.
+    var requestHead: HTTPRequestHead { get }
+
+    /// The maximal `TimeAmount` that is allowed to pass between reads from the Channel.
```
glbrntt, line 199: Presumably `fireChannelRead` events rather than issuing `read` events?

Sources/AsyncHTTPClient/ConnectionPool/HTTPExecutingRequest.swift
```diff
@@ -0,0 +1,249 @@
...
+    var idleReadTimeout: TimeAmount? { get }
+
+    /// Will be called by the ChannelHandler to indicate that the request is going to be send.
```
glbrntt, line 202:
```suggestion
    /// Will be called by the ChannelHandler to indicate that the request is going to be sent.
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPExecutingRequest.swift
```diff
@@ -0,0 +1,249 @@
...
+    ///
+    /// This will be called on the Channel's EventLoop. Do **not block** during your execution!
+    ///
+    /// - Returns: A bool indicating if the request should really be started. Return false if the request has already been cancelled.
+    ///            If the request is cancelled after this method call `executor.cancel()` to stop request execution.
```
glbrntt, line 207: Who's responsible for calling `executor.cancel()`? It seems like the caller is responsible, if so I'm not sure this comment belongs here.

Sources/AsyncHTTPClient/ConnectionPool/HTTPExecutingRequest.swift
```diff
@@ -0,0 +1,249 @@
...
+
+    /// Pause request streaming
+    ///
+    /// This will be called on the Channel's EventLoop. Do **not block** during your execution!
+    func pauseRequestBodyStream()
+
+    /// Resume request streaming
+    ///
+    /// This will be called on the Channel's EventLoop. Do **not block** during your execution!
+    func resumeRequestBodyStream()
```
glbrntt, line 228: Why do we need to distinguish between start and resume here? Does it make a difference to the implementor?

Sources/AsyncHTTPClient/ConnectionPool/HTTPExecutingRequest.swift
```diff
@@ -0,0 +1,249 @@
...
+    /// - Returns: A bool indicating if the request should really be started. Return false if the request has already been cancelled.
+    ///            If the request is cancelled after this method call `executor.cancel()` to stop request execution.
+    func willExecuteRequest(_: HTTPRequestExecutor) -> Bool
+
+    /// Will be called by the ChannelHandler to indicate that the request head has been sent.
+    ///
+    /// This will be called on the Channel's EventLoop. Do **not block** during your execution!
+    func requestHeadSent(_: HTTPRequestHead)
+
+    /// Start request streaming
```
glbrntt, line 215: Should we document when we expect these to be called (and how often). I.e. can the implementor safely assume that `startRequestBodyStream()` will be called at most once?

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -0,0 +1,529 @@
...
+
+    enum WriteAction {
+        case write(IOData, HTTPRequestExecutor, EventLoopFuture<Void>)
+
+        case failTask(Error)
+        case failFuture(Error)
+    }
+
+    mutating func writeNextRequestPart(_ part: IOData, taskEventLoop: EventLoop) -> WriteAction {
+        // this method is invoked with bodyPart and returns a future that signals that
```
glbrntt, line 198: this seems out of place.

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -0,0 +1,529 @@
...
+            preconditionFailure("Invalid state: \(self.state)")
+        case .executing(let executor, let requestState, let responseState):
+            switch requestState {
+            case .initialized:
+                preconditionFailure("Request stream must be started before it can be paused")
+            case .producing:
+                return .write(part, executor, taskEventLoop.makeSucceededFuture(()))
+
+            case .paused(.none):
+                let promise = taskEventLoop.makePromise(of: Void.self)
```
glbrntt, line 213: This confused me a little until I read the comment in the next case. Worth adding a comment here to say why we're making this promise.

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -0,0 +1,529 @@
...
+
+        case .executing(let executor, let requestState, .buffering(var buffer, next: .eof)):
+            self.state = .modifying
+
+            if let byteBuffer = buffer.popFirst() {
+                self.state = .executing(executor, requestState, .buffering(buffer, next: .eof))
+                return .consume(byteBuffer)
+            }
+
+            // buffer is empty, wait for more
```
glbrntt, line 462: Doesn't seem like we're waiting for more!

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -0,0 +1,529 @@
...
+            return .failTask(queuer, nil)
+        case .executing(let executor, let requestState, .buffering(_, next: .eof)):
+            self.state = .executing(executor, requestState, .buffering(.init(), next: .error(error)))
+            return .cancelExecutor(executor)
+        case .executing(let executor, let requestState, .buffering(_, next: .askExecutorForMore)):
+            self.state = .executing(executor, requestState, .buffering(.init(), next: .error(error)))
+            return .cancelExecutor(executor)
+        case .executing(let executor, _, .buffering(_, next: .error(_))):
+            // this would override another error, let's keep the first one
+            return .cancelExecutor(executor)
```
glbrntt, line 509: If `next:` is `.error` already, might we have already cancelled the executor? And does that matter? Does executors support being cancelled multiple times?

fabianfett: The executor supports being cancelled a bazillion times :)

Sources/AsyncHTTPClient/ConnectionPool/HTTPExecutingRequest.swift
fabianfett: I tried to be more specific.

Sources/AsyncHTTPClient/RequestBag.swift
Lukasa: You thumbs up'd but didn't do it. :wink:

Sources/AsyncHTTPClient/RequestBag.swift
Lukasa: Still not done.

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,418 @@
...
+            break
+        case .consume(let buffer):
+            self.delegate.didReceiveBodyPart(task: self.task, buffer)
+                .hop(to: self.task.eventLoop)
+                .whenComplete {
+                    switch $0 {
+                    case .success:
+                        self.consumeMoreBodyData0(resultOfPreviousConsume: $0)
+                    case .failure(let error):
+                        self.fail(error)
```
glbrntt, line 243: Why are we failing here?

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -0,0 +1,418 @@
...
+    private func receiveResponseHead0(_ head: HTTPResponseHead) {
+        self.task.eventLoop.assertInEventLoop()
+
+        // runs most likely on channel eventLoop
+        let forwardToDelegate = self.state.receiveResponseHead(head)
+
+        guard forwardToDelegate else { return }
+
+        self.delegate.didReceiveHead(task: self.task, head)
+            .hop(to: self.task.eventLoop)
```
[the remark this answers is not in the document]
fabianfett: Chatted about offline: A if-in-event-loop-else dance™️ also incurs an allocation, if we are not on the correct eventLoop. The closure context allocates.

fabianfett:
Adds a new error type.

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -0,0 +1,521 @@
...
+        case .redirected:
+            return .none
+        case .finished(error: _):
+            return .none
+        case .modifying:
+            preconditionFailure("Invalid state")
+        }
+    }
+
+    mutating func receiveResponseHead(_ head: HTTPResponseHead) -> Bool {
```
glbrntt, line 258: Nit: document what the return value represents

glbrntt (approved):
LGTM

Lukasa (approved):
You love to see it.

fabianfett:
CI failure because of known flaky test #371.

PR #385, opened 2021-06-29
Always access Task’s connection and cancelled property through a lock

fabianfett:
#### Motivation

- The task's properties `connection` and `cancelled` are sometime accessed through a lock and sometimes not.

#### Modifications

- Let's always go through a lock

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -633,17 +633,25 @@ extension HTTPClient {
 
         let promise: EventLoopPromise<Response>
         var completion: EventLoopFuture<Void>
-        var connection: Connection?
-        var cancelled: Bool
-        let lock: Lock
+        private let lock = Lock()
+        // protected by lock
+        private var _connection: Connection?
+        // protected by lock
+        private var _cancelled: Bool = false
         let logger: Logger // We are okay to store the logger here because a Task is for only one request.
+
+        var isCancelled: Bool {
+            self.lock.withLock { self._cancelled }
+        }
+
+        var connection: Connection? {
+            self.lock.withLock { self._connection }
+        }
 
         init(eventLoop: EventLoop, logger: Logger) {
             self.eventLoop = eventLoop
             self.promise = eventLoop.makePromise()
             self.completion = self.promise.futureResult.map { _ in }
-            self.cancelled = false
-            self.lock = Lock()
             self.logger = logger
         }
 
@@ -669,9 +677,9 @@ extension HTTPClient {
         /// Cancels the request execution.
         public func cancel() {
             let channel: Channel? = self.lock.withLock {
-                if !self.cancelled {
-                    self.cancelled = true
-                    return self.connection?.channel
+                if !self._cancelled {
+                    self._cancelled = true
+                    return self._connection?.channel
                 } else {
                     return nil
                 }
@@ -681,13 +689,16 @@ extension HTTPClient {
 
         @discardableResult
         func setConnection(_ connection: Connection) -> Connection {
-            return self.lock.withLock {
-                self.connection = connection
-                if self.cancelled {
-                    connection.channel.triggerUserOutboundEvent(TaskCancelEvent(), promise: nil)
-                }
-                return connection
+            let cancelled = self.lock.withLock { () -> Bool in
+                self._connection = connection
+                return self._cancelled
             }
+
+            if cancelled {
+                connection.channel.triggerUserOutboundEvent(TaskCancelEvent(), promise: nil)
+            }
+
+            return connection
         }
 
         func succeed<Delegate: HTTPClientResponseDelegate>(promise: EventLoopPromise<Response>?,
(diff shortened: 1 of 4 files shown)


fabianfett:
Failure in known flaky test #371. @swift-server-bot test this please

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
```diff
@@ -403,37 +408,68 @@ class HTTPClientInternalTests: XCTestCase {
...
             XCTAssertNoThrow(try httpBin.shutdown())
             XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully())
         }
 
         let request = try Request(url: "http://localhost:\(httpBin.port)/custom")
-
-        let requestFuture = httpClient.execute(request: request, delegate: delegate).futureResult
+        let logger = Logger(label: "test-connection")
+
+        let clientFactory = HTTPConnectionPool.ConnectionFactory(
```
Lukasa, line 428: Why has this dropped down to be so low-level?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -633,17 +633,25 @@ extension HTTPClient {
...
+        private var _connection: Connection?
+        // protected by lock
+        private var _cancelled: Bool = false
         let logger: Logger // We are okay to store the logger here because a Task is for only one request.
 
+        var isCancelled: Bool {
+            self.lock.withLock { self._cancelled }
+        }
+
+        var connection: Connection? {
```
Lukasa, line 647: It seems like this is only used in tests: can we mark it clearly for that purpose?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -702,7 +713,9 @@ extension HTTPClient {
 
         func fail<Delegate: HTTPClientResponseDelegate>(with error: Error,
                                                         delegateType: Delegate.Type) {
-            if let connection = self.connection {
+            let maybeConnection = self.lock.withLock { self._connection }
```
glbrntt, line 716: nit: why not use `self.connection` here?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -681,13 +689,16 @@ extension HTTPClient {
 
         @discardableResult
         func setConnection(_ connection: Connection) -> Connection {
```
glbrntt, line 691: not related to this PR, but why do we also return the connection that we set here?

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
fabianfett: I would really like to not expose the connection on the task in the future. In order to have control over the channel, we need to create it ourselves here.

fabianfett: There is only one other test remaining that accesses the `task.connection`:

`testWeCanActuallyExactlySetTheEventLoops`

We should be able to remove `task.connection` from `testWeCanActuallyExactlySetTheEventLoops`, once the new connection pool has landed. We would test only the delegate calls, since all connection eventLoop requirements can be tested with unit tests and will be tested with unit tests.

Sources/AsyncHTTPClient/HTTPHandler.swift
fabianfett: Will do!

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
Lukasa: Ok, can we encapsulate this new code in a function that clearly explains what it does? The code here is _very_ aware of implementation details of AHC, which suggests the test as a whole needs rewriting, but in the near term I'd be ok with just hiding the implementation details in a function.

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: Sorry, I wasn't clear: I meant changing the name to `var _testOnly_connection`

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
Lukasa: This is still open.

Sources/AsyncHTTPClient/HTTPHandler.swift
glbrntt: This is used on L715, maybe make it private and add an internal `_testOnly_connection` per Cory's suggestion?

Davidde94 (approved):

artemredkin (approved):

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
Lukasa: Bump. :smile:

fabianfett: Yes, ignored it for now, since I think we should be able to just remove this test, once other PRs have landed. The functionality will be tested by unit tests.

PR #386, opened 2021-06-29
Add `HTTPRequestStateMachine`

fabianfett:
This PR adds an explicit `HTTPRequestStateMachine`. It is intended to be used in two instances in the future:

1. As a substate machine for a `HTTP1ConnectionStateMachine`, when the connection is in a request.
2. As a state machine for a `HTTP2RequestHandler` that operates on a single http2 stream created by the multiplexer.

It remains to be seen, if we want to handle buffer body parts for backpressure directly in the `HTTPExecutingRequest` (see also #384) or directly the `HTTPRequestStateMachine`.


Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+    fileprivate enum State {
+        case initialized
+        case running(RequestState, ResponseState)
+        case finished
+
+        case failed(Error)
+    }
+
+    fileprivate enum RequestState {
+        enum ExpectedBody {
```
Lukasa, line 28: This is entirely unused.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOHTTP1
+
+struct HTTPRequestStateMachine {
```
Lukasa, line 18: All of these inner enums need comments on them explaining what their various sub-states mean, and why we would be in them.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+            case waiting
+        }
+
+        case initialized
+        case receivingBody(StreamControlState)
+        case endReceived
+    }
+
+    enum Action {
+        enum AfterHeadContinueWith {
```
Lukasa, line 56: This needs a new name. Maybe:

```suggestion
        enum NextMessageToSend {
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    mutating func readEventCaught() -> Action {
+        return .read
+    }
+
+    mutating func errorHappened(_ error: Error) -> Action {
+        switch self.state {
+        case .initialized:
+            preconditionFailure("After the state machine has been initialized, start must be called immidiatly. Thus this state is unreachable")
```
Lukasa, line 140:
```suggestion
            preconditionFailure("After the state machine has been initialized, start must be called immediately. Thus this state is unreachable")
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+        case .running:
+            self.state = .failed(error)
+            return .failRequest(error, closeStream: true)
+        case .finished, .failed:
+            preconditionFailure("If the request is finished or failed, we expect the connection state machine to remove the request immidiatly from its state. Thus this state is unreachable.")
```
Lukasa, line 145:
```suggestion
            preconditionFailure("If the request is finished or failed, we expect the connection state machine to remove the request immediately from its state. Thus this state is unreachable.")
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+        return .verifyRequest
+    }
+
+    mutating func requestVerified(_ head: HTTPRequestHead) -> Action {
+        guard case .running(.verifyRequest, .initialized) = self.state else {
+            preconditionFailure("Invalid state")
+        }
+
+        guard self.isChannelWritable else {
+            preconditionFailure("Unimplemented. Wait with starting the request here!")
```
Lukasa, line 164: Probably best not to precondition here?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+        }
+
+        if let value = head.headers.first(name: "content-length"), let length = Int(value), length > 0 {
```
Lukasa, line 167: This code should deal with the possibility the user has provided multiple, inconsistent transport header fields (i.e. multiple content-lengths _and_ a transfer-encoding: chunked).

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    mutating func requestStreamPartReceived(_ part: IOData) -> Action {
+        switch self.state {
+        case .initialized:
+            preconditionFailure("Invalid state: \(self.state)")
+
+        case .running(.verifyRequest, _),
+             .running(.endSent, _):
+            preconditionFailure("Invalid state: \(self.state)")
```
Lukasa, line 195: Can we collapse these with the one above?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+
+        case .running(.streaming(let expectedBodyLength, var sentBodyBytes, let producerState), let responseState):
+            // More streamed data is accepted, even though the producer should stop. However
+            // there might be thread syncronisations situations in which the producer might not
+            // be aware that it needs to stop yet.
```
Lukasa, line 200: This comment seems to be in the wrong place.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+
+            if let expected = expectedBodyLength {
+                if sentBodyBytes + part.readableBytes > expected {
+                    let error = HTTPClientError.bodyLengthMismatch
+
+                    switch responseState {
+                    case .initialized, .receivingBody:
+                        self.state = .failed(error)
+                    case .endReceived:
+                        #warning("TODO: This needs to be fixed. @Cory: What does this mean here?")
```
Lukasa, line 210: As discussed, this represents an error condition, and we should fail the request.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+
+            self.state = .running(requestState, responseState)
+
+            return .sendBodyPart(part)
+
+        case .failed:
+            return .wait
+
+        case .finished:
+            // a request may be finished, before we send all parts. We may still receive something
```
Lukasa, line 234: How can a request be finished before we sent all parts?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+            preconditionFailure("Invalid state")
+        case .running(.streaming(let expectedBodyLength, let sentBodyBytes, _), let responseState):
+            if let expected = expectedBodyLength, expected != sentBodyBytes {
+                let error = HTTPClientError.bodyLengthMismatch
+
+                switch responseState {
+                case .initialized, .receivingBody:
+                    self.state = .failed(error)
+                case .endReceived:
+                    #warning("TODO: This needs to be fixed. @Cory: What does this mean here?")
```
Lukasa, line 252: Same as before, this is an error as well.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+
+            self.state = .running(.endSent, responseState)
+            return .sendRequestEnd(startReadTimeoutTimer: self.idleReadTimeout)
+
+        case .running(.verifyRequest, _),
+             .running(.endSent, _):
+            preconditionFailure("Invalid state")
+
+        case .finished:
+            return .wait
```
Lukasa, line 267: Under what circumstance is this acceptable?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+    // MARK: - Response
+
+    mutating func receivedHTTPResponseHead(_ head: HTTPResponseHead) -> Action {
+        switch self.state {
+        case .initialized:
+            preconditionFailure("How can we receive a response head before sending a request head ourselves")
+
+        case .running(let requestState, .initialized):
+            switch requestState {
+            case .verifyRequest:
```
Lukasa, line 310: Can we fold this into the case above? I.e

```
switch self.state {
case .initialized, .running(.verifyRequest, _):
}
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+                preconditionFailure("How can we receive a response head before sending a request head ourselves")
+            case .streaming, .endSent:
+                break
+            }
+
+            self.state = .running(requestState, .receivingBody(.waiting))
+            return .forwardResponseHead(head)
+
+        case .running(_, .receivingBody), .running(_, .endReceived), .finished:
+            preconditionFailure("How can we sucessfully finish the request, before having received a head")
```
Lukasa, line 320: We can in principle receive multiple `head`s for 1XX responses, which this code does not tolerate. That's fine, but we should probably mark it here as an enhancement point for later.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+        }
+    }
+
+    mutating func receivedHTTPResponseBodyPart(_ body: ByteBuffer) -> Action {
+        switch self.state {
+        case .initialized:
+            preconditionFailure("How can we receive a response head before sending a request head ourselves")
+
+        case .running(_, .initialized):
+            preconditionFailure("How can we receive a response body, if we haven't a received a head")
```
Lukasa, line 332:
```suggestion
            preconditionFailure("How can we receive a response body, if we haven't received a head")
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+            case .waiting, .readEventPending:
+                break
+            case .downstreamHasDemand:
+                self.state = .running(requestState, .receivingBody(.waiting))
+            }
+
+            return .forwardResponseBodyPart(body, resetReadTimeoutTimer: self.idleReadTimeout)
+
+        case .running(_, .endReceived), .finished:
+            preconditionFailure("How can we sucessfully finish the request, before having received a head")
```
Lukasa, line 345: This comment seems out of place.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+        }
+    }
+
+    mutating func receivedHTTPResponseEnd() -> Action {
+        switch self.state {
+        case .initialized:
+            preconditionFailure("How can we receive a response head before sending a request head ourselves")
+
+        case .running(_, .initialized):
+            preconditionFailure("How can we receive a response body, if we haven't a received a head")
```
Lukasa, line 357: We're receiving an `end`, technically.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+
+        case .running(.streaming, .receivingBody(let streamState)):
+            preconditionFailure("Unimplemented")
+            #warning("@Fabian: We received response end, before sending our own request's end.")
```
Lukasa, line 361: We've discussed this.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+        }
+    }
+
+    mutating func idleReadTimeoutTriggered() -> Action {
+        guard case .running(.endSent, let responseState) = self.state else {
+            preconditionFailure("We only schedule idle read timeouts after we have sent the complete request")
+        }
+
+        if case .endReceived = responseState {
+            preconditionFailure("Invalid state: If we have received everything, we must not schedule further timeout timers")
```
Lukasa, line 408: This seems a bit harsh: it's possible we failed to cancel the timer in time.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+            case waiting
+        }
+
+        case initialized
+        case receivingBody(StreamControlState)
+        case endReceived
+    }
+
+    enum Action {
+        enum NextMessageToSend {
```
glbrntt, line 56: This doesn't appear to be used

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+            case readEventPending
+            case waiting
+        }
+
+        case initialized
+        case receivingBody(StreamControlState)
+        case endReceived
+    }
+
+    enum Action {
```
glbrntt, line 55: FWIW, in the past I've found it slightly awkward when debugging code which relies on a generic `Action` (rather than one action type per function on the state machine) since it's not clear at the call site what actions may occur as a result of prodding the state machine. I.e. `writabilityChanged` can only result in three of the 12 defined actions (`wait`, `resumeRequestBodyStream` and `pauseRequestBodyStream`) but that isn't clear when calling `writabilityChanged`. This is entirely personal preference though.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+                return .pauseRequestBodyStream
+            } else {
+                // no state change needed
+                return .wait
+            }
+        }
+    }
+
+    mutating func readEventCaught() -> Action {
+        return .read
```
glbrntt, line 134: Why do we unconditionally read? Does this not just apply to the `running` state?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+            self.state = .failed(error)
+            return .failRequest(error, closeStream: true)
+        case .finished, .failed:
+            preconditionFailure("If the request is finished or failed, we expect the connection state machine to remove the request immediately from its state. Thus this state is unreachable.")
+        }
+    }
+
+    mutating func start() -> Action {
+        guard case .initialized = self.state else {
+            preconditionFailure("Invalid state")
```
glbrntt, line 151: A comment indicating that `start()` must be called first and exactly once would be useful.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
glbrntt: Also can you add comments on why they're invalid states?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+            return ".verifyRequest"
+        case .streaming(expectedBodyLength: let expected, let sent, producer: let producer):
+            return ".sendingHead(sent: \(expected != nil ? String(expected!) : "-"), sent: \(sent), producer: \(producer)"
+        case .endSent:
+            return ".endSent"
+        }
+    }
+}
+
+extension HTTPRequestStateMachine.RequestState.ProducerControlState {
```
Davidde94, line 445: Missing `CustomStringConvertible`?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+        case failed(Error)
+    }
+
+    fileprivate enum RequestState {
+        enum ExpectedBody {
+            case length(Int)
+            case stream
+        }
+
+        enum ProducerControlState: Equatable {
```
Davidde94, line 33: Why not just make this type `:String`, and you can avoid the `CustomStringConvertible` below?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
Davidde94: +1 there's a lot going on here

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+                    producer: .paused
+                )
+                self.state = .running(requestState, responseState)
+                return .pauseRequestBodyStream
+            } else {
+                // no state change needed
+                return .wait
+            }
+        }
+    }
```
Davidde94, line 131: Let's break this function out a little - it does far too much IMHO

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
Davidde94: +1

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,467 @@
...
+
+        case .failed:
+            return .wait
+
+        case .finished:
+            // a request may be finished, before we send all parts. We may still receive something
+            // here because of a thread race
+            return .wait
+        }
+    }
```
Davidde94, line 238: I really think we should split this function up a bit. This is just personal opinion, but IMHO each case in a switch should have a few lines maximum, otherwise it can quickly become difficult to read and maintain.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
fabianfett: My idea here was that the state machine triggers the action: `return .verifyRequest`. It is the callers task to then call `HTTPHeaders.validate` that is currently in `RequestValidation.swift`. This should ensure that the request head and headers are valid. We could of course move the request validation directly into the state machine. (I didn't be cause it may change the http request, which I would consider a side effect).

fabianfett: For reference: https://github.com/swift-server/async-http-client/blob/main/Sources/AsyncHTTPClient/RequestValidation.swift

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
fabianfett: My feeling is that this function is only massive because of the comments now. Do you still think we should split up?

Tests/AsyncHTTPClientTests/HTTPRequestStateMachineTests.swift
```diff
@@ -0,0 +1,137 @@
...
+        let part1 = IOData.byteBuffer(ByteBuffer(bytes: [1]))
+        let part2 = IOData.byteBuffer(ByteBuffer(bytes: [2]))
+        let part3 = IOData.byteBuffer(ByteBuffer(bytes: [3]))
+        XCTAssertEqual(state.requestStreamPartReceived(part0), .sendBodyPart(part0))
+        XCTAssertEqual(state.requestStreamPartReceived(part1), .sendBodyPart(part1))
+
+        // oh the channel reports... we should slow down producing...
+        XCTAssertEqual(state.writabilityChanged(writable: false), .pauseRequestBodyStream)
+
+        // but we issued a .produceMoreRequestBodyData before... Thus, we must accept more produced
```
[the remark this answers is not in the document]
fabianfett: Yes, this might happen because of race conditions, if we receive `channelNotWritable` in thread A and we need to communicate it to thread B, B may have already created new data. For this reason, we need to accept more data even though we asked the producer to pause.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+
+    /// A sub state for a running request. More specifically for sending a request body.
+    fileprivate enum RequestState {
+        /// A sub state for sending a request body. Stores whether a producer should produce more
+        /// bytes or should pause.
+        enum ProducerControlState: Equatable {
+            /// The request body producer should produce more body bytes. The channel is writable,
+            case producing
+            /// The request body producer should pause producing more bytes. The channel is not writable,
```
Lukasa, line 52: Nit: 
```suggestion
            /// The request body producer should pause producing more bytes. The channel is not writable.
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+        /// The request has failed
+        case failed(Error)
+    }
+
+    /// A sub state for a running request. More specifically for sending a request body.
+    fileprivate enum RequestState {
+        /// A sub state for sending a request body. Stores whether a producer should produce more
+        /// bytes or should pause.
+        enum ProducerControlState: Equatable {
+            /// The request body producer should produce more body bytes. The channel is writable,
```
Lukasa, line 50: Nit: 
```suggestion
            /// The request body producer should produce more body bytes. The channel is writable.
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+            case producing
+            /// The request body producer should pause producing more bytes. The channel is not writable,
+            case paused
+        }
+
+        /// The request is streaming its request body. `expectedBodyLength` has a value, if the request header contained
+        /// a `"content-length"` header field. It is the request header contained a `"transfer-encoding" = "chunked"`
+        /// header field.
```
Lukasa, line 58: The second sentence here is something I don't understand: can you rewrite it?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+        /// A sub state for receiving a response. Stores whether the consumer has either signaled demand for more data or
+        /// is busy consuming the so far forwarded bytes
+        enum ConsumerControlState {
+            case downstreamIsConsuming(readPending: Bool)
+            case downstreamHasDemand
+        }
+
+        /// A response head has not been received yet.
+        case waitingForHead
+        /// A response head has been received and we are ready to consume more data of the wire
```
Lukasa, line 74: Nit: 
```suggestion
        /// A response head has been received and we are ready to consume more data off the wire
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+        case receivingBody(HTTPResponseHead, ConsumerControlState)
+        /// A response end has been received and we are ready to consume more data of the wire
```
Lukasa, line 76: Should be saying "ready to consume more data off the wire" here?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+
+    enum Action {
+        /// A action to execute, when we consider a request "done".
+        enum FinalStreamAction {
+            /// close the connection
+            case close
+            /// trigger a read event
+            case read
+            /// do nothing
+            case none
```
Lukasa, line 88: It would be useful to clarify when we do `none` here: namely, when we hit an error validating the request and haven't sent it yet.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+             .waitForChannelToBecomeWritable,
+             .running(_, .waitingForHead),
+             .running(_, .endReceived),
+             .finished,
+             .failed:
+            // If we are not in the middle of streaming the response body, we always want to get
+            // more data...
+            return .read
+        case .running(_, .receivingBody(_, .downstreamIsConsuming(readPending: true))):
+            preconditionFailure("It should not be possible to receive two reads after each other, if the first one hasn't been forwarded.")
```
Lukasa, line 206: Why is that?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+        case .running(let requestState, .receivingBody(let responseHead, .downstreamIsConsuming(readPending: false))):
+            self.state = .running(requestState, .receivingBody(responseHead, .downstreamIsConsuming(readPending: true)))
+            return .wait
+        case .running(let requestState, .receivingBody(let responseHead, .downstreamHasDemand)):
+            self.state = .running(requestState, .receivingBody(responseHead, .downstreamHasDemand))
```
Lukasa, line 211: There's no state change here. Is that intentional?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+            // No matter if the `producerState` is either `.producing` or `.paused` any bytes we
+            // receive shall be forwarded to the Channel right away. As long as we have not received
+            // a response with status >= 300.
+            //
+            // More streamed data is accepted, even though the producer may have been asked to
+            // pause. The reason for this is as follows: There might be thread synchronization
+            // situations in which the producer might not have received the plea to pause yet.
+
+            if let expected = expectedBodyLength {
+                if sentBodyBytes + part.readableBytes > expected {
```
Lukasa, line 273: These two if statements can be collapsed.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+                    let error = HTTPClientError.bodyLengthMismatch
+
+                    var clearReadTimeoutTimer = false
+                    if case .receivingBody = responseState, self.idleReadTimeout != nil {
+                        clearReadTimeoutTimer = true
+                    }
+
+                    return .failRequest(error, .close, clearReadTimeoutTimer: clearReadTimeoutTimer)
```
Lukasa, line 281: Is there any reason this state machine has to keep track of whether we need to clear the read timeout timer? Presumably someone has set the timer, and they know if they have one or not, so should unconditionally clear it.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+
+        case .running(.streaming(let expectedBodyLength, let sentBodyBytes, _), .endReceived):
+            if let expected = expectedBodyLength, expected != sentBodyBytes {
+                let error = HTTPClientError.bodyLengthMismatch
+                self.state = .failed(error)
+                return .failRequest(error, .close, clearReadTimeoutTimer: false)
+            }
+
+            self.state = .finished
+            return .succeedRequest(.none, clearReadTimeoutTimer: false)
```
Lukasa, line 353: Yeah, I don't know why the read timeout isn't getting cleared here. I think whether the read timeout is cleared or reset should not be the responsibility of this state machine: the read timeout should get set when either we've sent our `.end` or received our first `.head`, and from that point on it shouldn't be our problem. It adds too many responsibilities to this object.

Lukasa: Alternatively, if you _really_ want it to be composed into this state machine, factor it out into a little state machine of its own and give it an interface that you can call from here. It'll make it much easier to validate its correctness. But honestly I think it's overcomplicating this object.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+            assert(head.status.code >= 300)
+            assert(producerState == .paused, "Expected to have paused the request body stream, when the head was received. Invalid state: \(self.state)")
+            self.state = .finished
+            return .succeedRequest(.close, clearReadTimeoutTimer: self.idleReadTimeout != nil)
+
+        case .running(.endSent, .receivingBody(_, let streamState)):
+            let finalAction: Action.FinalStreamAction
+            switch streamState {
+            case .downstreamIsConsuming(readPending: true):
+                finalAction = .read
```
Lukasa, line 499: I do not understand why the consumer has any control over what we do with the HTTP connection at this point.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -0,0 +1,625 @@
...
+            return .read
+        }
+    }
+
+    mutating func errorHappened(_ error: Error) -> Action {
+        switch self.state {
+        case .initialized:
+            preconditionFailure("After the state machine has been initialized, start must be called immediately. Thus this state is unreachable")
+        case .verifyingRequest, .waitForChannelToBecomeWritable:
+            // the request failed, before it was send onto the wire.
```
glbrntt, line 221:
```suggestion
            // the request failed, before it was sent onto the wire.
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
glbrntt: IMHO I think this is fine as is, the amount of code in each case is pretty small.

glbrntt (commented):
Nothing substantive from me. The comments were helpful though, thanks for adding them all!

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
fabianfett: Given that reads are triggered after a `readComplete` (we don't modify any channel options), I would assume that a new read event can only happen, if in the mean time data was actually read from the wire. To read from the wire the consumer needs to signal demand: this will transform `.downstreamIsConsuming(readPending: true)` -> `.downstreamHasDemand`. Once we get new data: `.downstreamIsConsuming(readPending: false)`, only then we should get into a state where a new `read` event might happen.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
fabianfett: Yes.

fabianfett: Yes. Added a comment.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
fabianfett: I'm not sure I get your point here?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
Lukasa: A `read` event is not required to trigger any non-zero number of calls to `channelRead`: it's only required to trigger a single call to `channelReadComplete`. More broadly, there's no real reason to forbid this, I don't think.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
Lukasa: Why does our `finalAction` depend on what the consumer is doing with the data?

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
Davidde94: Bump

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
Davidde94: Yeh this is fine now

Davidde94 (approved):

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
fabianfett: The `finalAction` depends on whether backpressure was used. If we receive the response end and we have a read event buffered, I think we should make sure this read event is passed forward. This way we ensure that the channel is in a "normal" state (no buffered read event), once we start the next request. I see it as a cleanup job.

Lukasa (approved):
LGTM, I think I'm happy with this as it stands.

PR #388, opened 2021-07-06
Buffer `channelRead`s in `ResponseStreamState` until the next `channelReadComplete`

fabianfett:
### Motivation

After a lengthy offline discussion @Lukasa, we concluded that it would be best, if the `HTTPRequestStateMachine` would buffer up all reads until the next `channelReadComplete`. The main reason is, we minimize potential cross thread communication, by passing collected byteBuffers onward.

### Modifications

- `ConsumerControlState` was replaced with `ResponseStreamState `
- `channelRead`s are buffered up to a `channelReadComplete`.
- on `channelReadComplete` all buffered body parts are forwarded to the consumer


Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -58,17 +60,19 @@ struct HTTPRequestStateMachine {
...
+            /// The state machines expects further writes to `channelRead`. The writes are appended to the buffer.
+            case waitingForBytes(CircularBuffer<ByteBuffer>)
+            /// The state machines expects a call to `demandMoreResponseBodyParts` or `read`. The buffer is
+            /// empty. It is preserved for performance reasons.
+            case waitingForReadOrDemand(CircularBuffer<ByteBuffer>)
+            /// The state machines expects a call to `read`. The buffer is empty. It is preserved for performance reasons.
+            case waitingForRead(CircularBuffer<ByteBuffer>)
+            /// The state machines expects a call to `demandMoreResponseBodyParts`. The buffer is empty. It is
+            /// preserved for performance reasons.
+            case waitingForDemand(CircularBuffer<ByteBuffer>)
```
Lukasa, line 75: By making this a mini-state-machine, I recommend going the whole way: pull this out to a file and give it a bunch of methods that correspond to the events that can occur to it (e.g. `func read`). Then, the bigger state machine can just call those.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine+Demand.swift
```diff
@@ -0,0 +1,149 @@
...
+
+        enum Action {
+            case read
+            case wait
+        }
+
+        private var state: State
+
+        init(expectingBody: Bool) {
+            self.state = .waitingForBytes(CircularBuffer(initialCapacity: expectingBody ? 16 : 0))
```
Lukasa, line 44: Given that there is no such thing as an empty circular buffer (`initialCapacity: 0` still allocates) we may as well skip the branch and just always set `16`.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine+Demand.swift
```diff
@@ -0,0 +1,154 @@
...
+
+            case .modifying:
+                preconditionFailure("Invalid state: \(self.state)")
+            }
+        }
+
+        mutating func end() -> CircularBuffer<ByteBuffer> {
+            switch self.state {
+            case .waitingForBytes(let buffer):
+                // This should never happen. But we don't want to precondition this behavior. Let's just
```
Lukasa, line 140: This comment is wrong.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -376,56 +358,138 @@ struct HTTPRequestStateMachine {
...
             return .wait
         }
 
+        var expectingBody: Bool = false
+        if let length = head.headers.first(name: "content-length").flatMap({ Int($0) }) {
+            if length > 0 {
+                expectingBody = true
+            }
+        } else if head.headers.contains(name: "transfer-encoding") {
+            expectingBody = true
```
Lukasa, line 453: We can delete this code per my above comment.

Lukasa (approved):

PR #396, opened 2021-07-07
Add HTTP1ConnectionStateMachine

fabianfett:
This PR adds an `HTTP1ConnectionStateMachine` that shall be used in an upcoming `HTTP1ChannelHandler`.

This PR doesn't have any tests yet. I would like to get some early feedback on the general form and will then add tests.


Lukasa (commented):
This generally seems reasonable to me.

Lukasa (approved):

PR #397, opened 2021-07-07
Add a waiter box for the connection pool

fabianfett:
### Motivation

For the new `HTTPConnectionPool.StateMachine` we need a new `Waiter` type.

### Modifications

- Add a new `HTTPConnectionPool.Waiter`


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Waiter.swift
```diff
@@ -0,0 +1,46 @@
...
+            private let objectIdentifier: ObjectIdentifier
+
+            init(_ request: HTTPScheduledRequest) {
+                self.objectIdentifier = ObjectIdentifier(request)
+            }
+        }
+
+        let id: ID
+        let request: HTTPScheduledRequest
+        let eventLoopRequirement: EventLoop?
```
Lukasa, line 29: In general for good hygiene I'd recommend these being `var` instead of `let`.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Waiter.swift
```diff
@@ -0,0 +1,46 @@
...
+
+        init(request: HTTPScheduledRequest, eventLoopRequirement: EventLoop?) {
```
Lukasa, line 31: Is there a reason we take the preference separately from the `HTTPScheduledRequest`, which encodes a preference?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Waiter.swift
```diff
@@ -0,0 +1,64 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+extension HTTPConnectionPool {
+    struct Waiter {
+        struct ID: Hashable {
```
glbrntt, line 19: What a great idea ;)

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Waiter.swift
```diff
@@ -0,0 +1,64 @@
...
+
+        var id: ID {
+            ID(self.request)
+        }
+
+        var request: HTTPScheduledRequest {
+            didSet {
+                self.updateEventLoopRequirement()
+            }
+        }
```
glbrntt, line 35: Hmm, this feels a bit off. If we can the `request` then the `id` of the waiter also changes. Is that the behaviour we actually want or should it really be called `requestID`?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Waiter.swift
```diff
@@ -0,0 +1,64 @@
...
+    struct Waiter {
+        var requestID: RequestID {
+            RequestID(self.request)
+        }
+
+        var request: HTTPScheduledRequest {
+            didSet {
+                self.updateEventLoopRequirement()
+            }
+        }
```
fabianfett, line 35: @glbrntt Change to `RecordID`. But I moved the `RecordID` type out of the waiter.

Lukasa: This still says `RequestID`. Have I missed something?

fabianfett: It used to be called `ID` and was on the type `Waiter`, but dependent on the `HTTPScheduledRequest`

Lukasa: I see, I was confused by this comment saying the change was to `RecordID`.

fabianfett: https://github.com/swift-server/async-http-client/pull/397#discussion_r666253889

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Waiter.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    struct Waiter {
+        var requestID: RequestID {
+            RequestID(self.request)
+        }
+
+        var request: HTTPScheduledRequest {
+            didSet {
```
Lukasa, line 32: Might it be easier to make `eventLoopRequirement` a computed property?

fabianfett: Since `HTTPScheduledRequest` is a protocol, I wanted to keep the number of calls going through the lookup table as small as possible.

Lukasa (approved):

glbrntt (approved):

PR #398, opened 2021-07-07
Add HTTPConnectionPool Connection as a box type

fabianfett:
### Motivation

In our `HTTPConnectionPool.StateMachine` we want to manage HTTPConnection states. However, to ensure that our state machine reflects the outside world it is beneficial if the state machine would hold references to the actual existing connections. Exposing our Connection types directly to the state machine could lead to abuse though. Developers might call actions directly on the Connection within the state machine.

### Modifications

- Add a `HTTPConnectionPool.Connection` type that is a box around an actual connection. The purpose of the box is to ensure no actions are invoked on the connection within the state machine. Further the box can be used for testing the state machine without creating actual connections.

### Caveats

- This PR doesn't add any tests. This code will be thoroughly tested once the actual state machines land. (This code is not hit for actual uses as of today)

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
@@ -15,7 +15,7 @@
 import NIO
 
 enum HTTPConnectionPool {
-    struct Connection: Equatable {
+    struct Connection: Hashable {
         typealias ID = Int
 
         private enum Reference {
@@ -91,5 +91,38 @@ enum HTTPConnectionPool {
 //                return false
             }
         }
+
+        func hash(into hasher: inout Hasher) {
+            switch self._ref {
+            case .__testOnly_connection(let id, let eventLoop):
+                hasher.combine(id)
+                hasher.combine(eventLoop.id)
+            }
+        }
     }
+}
+
+struct EventLoopID: Hashable {
+    private var id: Identifier
+
+    private enum Identifier: Hashable {
+        case objectIdentifier(ObjectIdentifier)
+        case __testOnly_fakeID(Int)
+    }
+
+    init(_ eventLoop: EventLoop) {
+        self.init(.objectIdentifier(ObjectIdentifier(eventLoop)))
+    }
+
+    private init(_ id: Identifier) {
+        self.id = id
+    }
+
+    static func __testOnly_fakeID(_ id: Int) -> EventLoopID {
+        return EventLoopID(.__testOnly_fakeID(id))
+    }
+}
+
+extension EventLoop {
+    var id: EventLoopID { EventLoopID(self) }
 }
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
@@ -12,8 +12,84 @@
 //
 //===----------------------------------------------------------------------===//
 
+import NIO
+
 enum HTTPConnectionPool {
-    struct Connection {
+    struct Connection: Equatable {
         typealias ID = Int
+
+        private enum Reference {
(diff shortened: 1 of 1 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -12,8 +12,114 @@
 //
 //===----------------------------------------------------------------------===//
 
+import NIO
+
 enum HTTPConnectionPool {
-    struct Connection {
+    struct Connection: Equatable {
```
Lukasa, line 18: This should be `Hashable`.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -12,8 +12,114 @@
...
         typealias ID = Int
+
+        private enum Reference {
+//            case http1_1(HTTP1Connection)
+
+            #if DEBUG
```
Lukasa, line 24: We sometimes want to run tests in release mode: we should probably just define this always and name it to indicate its use in testing.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -12,8 +12,114 @@
...
 
+import NIO
+
 enum HTTPConnectionPool {
-    struct Connection {
+    struct Connection: Equatable {
         typealias ID = Int
+
+        private enum Reference {
+//            case http1_1(HTTP1Connection)
```
Lukasa, line 22: Can you provide explanations for why all of these are commented out?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -12,8 +12,114 @@
...
+//            case .http1_1(let connection):
+//                return connection.channel.eventLoop
+            #if DEBUG
+                case .testing(_, let eventLoop):
+                    return eventLoop
+            #endif
+            }
+        }
+
+        #if DEBUG
```
Lukasa, line 63: Don't hide this behind debug: use underscores and `testOnly` (e.g. `__testOnly_channel`.)

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
fabianfett: Yes. The the sad and simple reason is: We don't have a HTTP1Connection yet. All of those will be enabled as soon as the `HTTP1Connection` has landed. Landing the connection box now enables us to already review the ConnectionPool StateMachines.

Lukasa: Sorry, I meant in the code. :wink:

fabianfett: Damn it. Fixed.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -12,8 +12,124 @@
...
-    struct Connection {
+    struct Connection: Hashable {
         typealias ID = Int
+        
+        // PLEASE NOTE:
+        // The HTTP/1.1 connection code here is commented out, for a sad and simple reason: We
+        // don't have a HTTP1Connection yet. As soon as the HTTP1Connection has landed
+        // (https://github.com/swift-server/async-http-client/pull/400) we will enable
+        // HTTP1Connections here. Landing the connection box now enables us to already review the
+        // ConnectionPool StateMachines.
```
fabianfett, line 26: @Lukasa

Lukasa (approved):

PR #401, opened 2021-07-08
Add HTTP2Connection

fabianfett:
This pr adds an `HTTP2Connection` that we want to use in our connection pool going forward.

It has no tests at all. I would like to get some early feedback before adding those.


Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    // MARK: - Run Actions
+
+    func run(_ action: HTTPRequestStateMachine.Action, context: ChannelHandlerContext) {
+        switch action {
+        case .sendRequestHead(let head, let startBody):
+            if startBody {
+                context.writeAndFlush(self.wrapOutboundOut(.head(head)), promise: nil)
+                self.request!.resumeRequestBodyStream()
```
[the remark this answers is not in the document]
glbrntt: also, SSWG nit, we need to say why we can bang the request.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -0,0 +1,170 @@
...
+
+            self.state = .connected
+        }
+
+        mutating func settingsReceived(_ settings: HTTP2Settings) -> Action {
+            guard case .connected = self.state else {
+                preconditionFailure("Invalid state")
+            }
+
+            let maxStream = settings.first(where: { $0.parameter == .maxConcurrentStreams })?.value ?? 100
```
glbrntt, line 123: I don't believe it's invalid for there to be more than one entry per setting, i.e. it's possible for `.maxConcurrentStreams` to appear more than once in this array. Settings must however be processed in order so we should use `last(where:)` instead.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -0,0 +1,170 @@
...
+
+            self.state = .active(openStreams: 0, maxStreams: maxStream)
```
glbrntt, line 125: > A SETTINGS frame MUST be sent by both endpoints at the start of a connection and MAY be sent at any other time by either endpoint over the lifetime of the connection.

We need to handle this otherwise we'll blow away our open stream count. Moreover we can't always assume 100 as the default when no setting is present (since a mid-connection settings update may change another setting and we don't want to inadvertently overwrite the previous value of max concurrent streams).

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -0,0 +1,242 @@
...
+class HTTP2ClientRequestHandler: ChannelDuplexHandler {
+    typealias OutboundIn = HTTPExecutingRequest
+    typealias OutboundOut = HTTPClientRequestPart
+    typealias InboundIn = HTTPClientResponsePart
+
+    private let eventLoop: EventLoop
+
+    private(set) var channelContext: ChannelHandlerContext?
+    private(set) var state: HTTPRequestStateMachine = .init(isChannelWritable: false)
+    private(set) var request: HTTPExecutingRequest?
```
Lukasa, line 29: Any reason for these not to be fully `private`?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -0,0 +1,242 @@
...
+
+    init(eventLoop: EventLoop) {
+        self.eventLoop = eventLoop
+    }
+
+    func handlerAdded(context: ChannelHandlerContext) {
+        self.channelContext = context
+
+        let isWritable = context.channel.isActive && context.channel.isWritable
+        let action = self.state.writabilityChanged(writable: isWritable)
```
Lukasa, line 39: Will the state machine tolerate this event before the first `write`?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -0,0 +1,146 @@
...
+
+        // 1. Modify channel pipeline and add http2 handlers
+        let sync = channel.pipeline.syncOperations
+
+        let http2Handler = NIOHTTP2Handler(mode: .client, initialSettings: nioDefaultSettings)
+        let idleHandler = HTTP2IdleHandler(connection: self, logger: self.logger)
+
+        try sync.addHandler(http2Handler, position: .last)
+        try sync.addHandler(idleHandler, position: .last)
+        try sync.addHandler(self.multiplexer, position: .last)
```
Lukasa, line 90: Why are we building the HTTP/2 handlers ourselves?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -0,0 +1,170 @@
...
+            case .initialized:
+                preconditionFailure("Invalid state")
+            case .connected:
+                self.state = .goAwayReceived(openStreams: 0, maxStreams: 0)
+                return .notifyConnectionGoAwayReceived
+            case .active(let openStreams, let maxStreams):
+                self.state = .goAwayReceived(openStreams: openStreams, maxStreams: maxStreams)
+                return .notifyConnectionGoAwayReceived
+            case .goAwayReceived:
+                preconditionFailure("Invalid state")
```
glbrntt, line 140: IIRC it's valid for a server to send more than one GOAWAY frame.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
fabianfett: Because we need to add the IdleHandler in the middle. This trick was imported from grpc based on @glbrntt's suggestion.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
fabianfett: Added a test case to the `HTTPRequestStateMachineTests`: `testConnectionBecomesWritableBeforeFirstRequest`.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -0,0 +1,146 @@
...
+import Logging
+import NIO
+import NIOHTTP2
+
+protocol HTTP2ConnectionDelegate {
+    func http2ConnectionStreamClosed(_: HTTP2Connection, availableStreams: Int)
+    func http2ConnectionClosed(_: HTTP2Connection)
+}
+
+class HTTP2Connection {
```
Lukasa, line 24: Any reason this shouldn't be `final`?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
Lukasa: Can we add a comment to that effect please?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -0,0 +1,146 @@
...
+            inboundStreamInitializer: { (channel) -> EventLoopFuture<Void> in
+                struct HTTP2PushNotsupportedError: Error {}
+                return channel.eventLoop.makeFailedFuture(HTTP2PushNotsupportedError())
+            }
+        )
+        self.delegate = delegate
+        self.state = .starting(readyToAcceptConnectionsPromise)
+        self.readyToAcceptConnectionsFuture = readyToAcceptConnectionsPromise.futureResult
+
+        // 1. Modify channel pipeline and add http2 handlers
```
Lukasa, line 82: It's not clear to me that we should do this on `init`. I tend to be a bit nervous about `init`s that do a lot of work. Would it be better to move this to a, say, `start` method?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -0,0 +1,146 @@
...
+
+        let readyToAcceptConnectionsPromise = channel.eventLoop.makePromise(of: Void.self)
+
+        self.channel = channel
+        self.id = connectionID
+        self.logger = logger
+        self.multiplexer = HTTP2StreamMultiplexer(
+            mode: .client,
+            channel: channel,
+            targetWindowSize: 65535,
```
Lukasa, line 70: We should almost certainly default this to a substantially larger number. 65535 is the protocol default and it's very conservative.

Lukasa: I'd normally suggest something closer to 8MB.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -0,0 +1,224 @@
...
+internal final class HTTP2IdleHandler: ChannelDuplexHandler {
+    typealias InboundIn = HTTP2Frame
+    typealias InboundOut = HTTP2Frame
+    typealias OutboundIn = HTTP2Frame
+    typealias OutboundOut = HTTP2Frame
+
+    let logger: Logger
+    let connection: HTTP2Connection
+
+    var state: StateMachine = .init()
```
glbrntt, line 28: Any reason this and others can't be private?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -0,0 +1,224 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Logging
+import NIO
+import NIOHTTP2
+
+internal final class HTTP2IdleHandler: ChannelDuplexHandler {
```
glbrntt, line 19: Looks like this is actually just an inbound handler?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -0,0 +1,224 @@
...
+            }
+        }
+
+        mutating func settingsReceived(_ settings: HTTP2Settings) -> Action {
+            switch self.state {
+            case .initialized, .closed:
+                preconditionFailure("Invalid state: \(self.state)")
+
+            case .connected:
+                let maxStreams = settings.last(where: { $0.parameter == .maxConcurrentStreams })?.value ?? 100
```
glbrntt, line 143: Probably worth adding a comment to explain `last` and the default value

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -0,0 +1,224 @@
...
+                self.state = .closing(openStreams: openStreams, maxStreams: maxStreams)
+                return .nothing
+
+            case .closed, .closing:
+                return .nothing
+            }
+        }
+
+        mutating func streamCreated() -> Action {
+            guard case .active(var openStreams, let maxStreams) = self.state else {
```
glbrntt, line 200: nit: switch instead of guard

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -0,0 +1,224 @@
...
+            assert(openStreams <= maxStreams)
+
+            self.state = .active(openStreams: openStreams, maxStreams: maxStreams)
+            return .nothing
+        }
+
+        mutating func streamClosed() -> Action {
+            guard case .active(var openStreams, let maxStreams) = self.state else {
+                preconditionFailure("Invalid state")
+                // TODO: What happens, if we received a go away?!??!
```
glbrntt, line 214: We should handle this in a similar way to active but we should close the connection when `openStreams` drops to zero.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -136,20 +127,25 @@ class HTTP2Connection {
 
         self.multiplexer.createStreamChannel(promise: createStreamChannelPromise) { channel -> EventLoopFuture<Void> in
             do {
+                // We only support http/2 over an https connection – using the Application-Layer
+                // Protocol Negotiation (ALPN). For this reason it is save to fix this to `.https`.
```
glbrntt, line 131:
```suggestion
                // Protocol Negotiation (ALPN). For this reason it is safe to fix this to `.https`.
```

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -197,28 +233,41 @@ extension HTTP2IdleHandler {
...
+            switch self.state {
+            case .active(var openStreams, let maxStreams):
+                openStreams += 1
+                self.state = .active(openStreams: openStreams, maxStreams: maxStreams)
+                return .nothing
 
-            openStreams += 1
-            assert(openStreams <= maxStreams)
+            case .closing(var openStreams, let maxStreams):
+                openStreams += 1
```
glbrntt, line 243: In what situations can we open a stream when we're closing?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2IdleHandler.swift
```diff
@@ -197,28 +233,41 @@ extension HTTP2IdleHandler {
...
-                // TODO: What happens, if we received a go away?!??!
-            }
+            switch self.state {
+            case .active(var openStreams, let maxStreams):
+                openStreams -= 1
+                self.state = .active(openStreams: openStreams, maxStreams: maxStreams)
+                return .notifyConnectionStreamClosed(currentlyAvailable: maxStreams - openStreams)
 
-            openStreams -= 1
-            assert(openStreams >= 0)
```
glbrntt: Why did we drop the assertion? It seems relevant at least in the `active` case.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -243,7 +303,11 @@ extension HTTPConnectionPool.ConnectionFactory {
         case .https:
             var tlsConfig = self.tlsConfiguration
             // since we can support h2, we need to advertise this in alpn
-            tlsConfig.applicationProtocols = ["http/1.1" /* , "h2" */ ]
+            if self.allowHTTP2Connections {
+                tlsConfig.applicationProtocols = ["http/1.1", "h2"]
```
glbrntt, line 307: The protocols offered in ALPN are in descending preference order. Should we offer h2 first and prefer it over http/1.1? (https://datatracker.ietf.org/doc/html/rfc7301#section-3.1)

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -341,7 +405,11 @@ extension HTTPConnectionPool.ConnectionFactory {
         -> EventLoopFuture<NIOClientTCPBootstrapProtocol> {
         // since we can support h2, we need to advertise this in alpn
         var tlsConfig = self.tlsConfiguration
-        tlsConfig.applicationProtocols = ["http/1.1" /* , "h2" */ ]
+        if self.allowHTTP2Connections {
+            tlsConfig.applicationProtocols = ["http/1.1", "h2"]
```
glbrntt, line 409: Same comment re: ordering

glbrntt (commented):
I didn't get a chance to look at the tests yet but leaving these comments for the time being.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -0,0 +1,330 @@
...
+
+final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
+    typealias OutboundIn = HTTPExecutableRequest
+    typealias OutboundOut = HTTPClientRequestPart
+    typealias InboundIn = HTTPClientResponsePart
+
+    private let eventLoop: EventLoop
+
+    private var state: HTTPRequestStateMachine = .init(isChannelWritable: false) {
+        didSet {
```
Lukasa, line 28: This should be `willSet`, or this won't fire early enough.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    private var idleReadTimeoutStateMachine: IdleReadStateMachine?
+    private var idleReadTimeoutTimer: Scheduled<Void>?
+
+    init(eventLoop: EventLoop) {
+        self.eventLoop = eventLoop
+    }
+
+    func handlerAdded(context: ChannelHandlerContext) {
```
Lukasa, line 53: Can I propose validating that the event loop associated with our channel and the loop passed to us in `init` are the same here?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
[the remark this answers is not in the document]
Lukasa: This nit still applies.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -0,0 +1,330 @@
...
+    func errorCaught(context: ChannelHandlerContext, error: Error) {
+        let action = self.state.errorHappened(error)
+        self.run(action, context: context)
+    }
+
+    // MARK: Channel Outbound Handler
+
+    func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+        let request = self.unwrapOutboundIn(data)
+        self.request = request
```
Lukasa, line 107: Should we be validating that `self.request` is `nil` here?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -0,0 +1,301 @@
...
+        var id: ID {
+            ID(self.channel)
+        }
+
+        init(_ channel: Channel) {
+            self.channel = channel
+        }
+
+        static func == (lhs: Self, rhs: Self) -> Bool {
+            lhs.channel === rhs.channel
```
Lukasa, line 67: Make the equatable/hashable implementations the same.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
fabianfett: Added a comment, that this check is done by the statemachine.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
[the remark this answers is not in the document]
fabianfett: See L142-145

Lukasa (approved):

PR #403, opened 2021-07-09
Connection Pool HTTP1 StateMachine

fabianfett:
- Adds a new connection pool state machine
- Adds a substate machine for HTTP1 remotes that support HTTP1

Minor version bump:

- Added a new error code


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                        if connReturn > existingMatchReturn {
+                            goodMatch = (index, connReturn)
+                        }
+                    }
+                }
+            }
+
+            // if we found an eventLoopMatch, we can execute the request right away
+            if let (index, _) = eventLoopMatch {
+                assert(self.waiters.isEmpty, "If a connection is available, why are there any waiters")
```
Lukasa, line 239: Is this code guaranteed to be re-entrancy free?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                var connectionState = self.connections[index]
+                let connection = connectionState.lease()
+                self.connections[index] = connectionState
```
Lukasa, line 242: Prefer:

```swift
let connection = self.connections[index].lease()
```

it performs better.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+            }
+
+            // if we found a good match, let's use this
+            if let (index, _) = goodMatch {
+                assert(self.waiters.isEmpty, "If a connection is available, why are there any waiters")
+                if !required {
+                    var connectionState = self.connections[index]
+                    let connectionID = connectionState.connectionID
+                    let connection = connectionState.lease()
+                    self.connections[index] = connectionState
```
Lukasa, line 256: Same note here.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                    return .init(
+                        .executeRequest(request, connection, cancelWaiter: nil),
+                        .cancelTimeoutTimer(connectionID)
+                    )
+                } else {
+                    assert(self.connections.count - self.maximumConcurrentConnections == 0)
+                    var oldConnectionState = self.connections[index]
+                    let newConnectionID = self.idGenerator.next()
+                    let newWaiter = Waiter(request: request)
+                    self.connections[index] = .init(connectionID: newConnectionID, eventLoop: prefferedEL, waiter: newWaiter)
```
Lukasa, line 266: For this one consider `swap`:

```
var replacement = Connection(connectionID: newConnectionID, eventLoop: preferredEL, waiter: newWaiter)
swap(&replacement, &self.connections[index])
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+            }
+
+            // all connections are busy, and there is no more room to create further connections
+            let waiter = Waiter(request: request)
+            self.waiters.append(waiter)
+            return .init(
+                .scheduleWaiterTimeout(waiter.requestID, request, on: prefferedEL),
+                .none
+            )
+        }
```
Lukasa, line 293: This method is 140 lines long. It needs to be refactored into several smaller methods or it will be impossible to follow.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                    )
+                }
+            }
+
+            // we didn't find any match at all... Let's create a new connection, if there is room
+            // left
+            if self.connections.count < self.maximumConcurrentConnections {
+                let newConnectionID = self.idGenerator.next()
+                let newWaiter = Waiter(request: request)
+                self.connections.append(.init(connectionID: newConnectionID, eventLoop: prefferedEL, waiter: newWaiter))
```
Lukasa, line 279: It is unwise to force a waiter to own a _specific_ connection attempt. Instead, you want to queue up waiters and serve them using a FIFO as connections become available. This helps prevent issues where either a) waiter timeouts throw away connections we could use to serve other waiters, or b) problems where connections become active in a different order to the order of waiters arriving, risking more timeouts than necessary.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                    connectionState.lease()
+                    self.connections[index] = connectionState
+                    return .init(
+                        .executeRequest(waiter.request, connection, cancelWaiter: waiter.requestID),
+                        .none
+                    )
+                }
+
+                // 2. if we don't have an associated waiter for this connection, pick the first one
+                //    from the queue
```
Lukasa, line 317: Hmm, it's not clear to me why we pick the first one from the queue instead of the first one _that can run on this loop_. What's the reasoning?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                preconditionFailure("We tried to create a new connection, that we know nothing about?")
+            }
+
+            var connectionState = self.connections[index]
+
+            switch self.state {
+            case .running:
+                var requestAction: StateMachine.RequestAction = .none
+                if let failedWaiter = connectionState.failed() {
+                    requestAction = .failRequest(failedWaiter.request, error, cancelWaiter: failedWaiter.requestID)
```
Lukasa, line 380: This should not be necessary: if we failed to connect, we should try again (on an exponential backoff) until we either succeed or have more connection attempts in flight than waiters.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+
+            assert(self.state == .running, "If we are shutting down, we must not have any idle connections")
+
+            var connectionState = self.connections[index]
+            guard connectionState.isAvailable else {
+                // connection is not available anymore, we may have just leased it for a request
+                return .init(.none, .none)
+            }
+
+            assert(self.waiters.isEmpty, "We have an idle connection, that times out, but waiters? Something is very wrong!")
```
Lukasa, line 437: Remember that timers can fire during the same tick as other work, so this is absolutely possible. This should be checked for and ignored.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                preconditionFailure("A connection that we don't know was released? Something is very wrong...")
+            }
+
+            var connectionState = self.connections[index]
+            connectionState.release()
+
+            switch self.state {
+            case .running:
+                guard let nextWaiter = self.waiters.popFirst() else {
+                    // there is no more work todo immidiatly
```
Lukasa, line 454:
```suggestion
                    // there is no more work to do immediately
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+
+                guard nextWaiter.canBeRun(on: connectionState.eventLoop) else {
+                    let eventLoop = nextWaiter.eventLoopRequirement!
+                    let newConnection = HTTP1ConnectionState(
+                        connectionID: self.idGenerator.next(),
+                        eventLoop: eventLoop,
+                        waiter: nextWaiter
+                    )
+                    self.connections[index] = newConnection
+                    return .init(.none, .replaceConnection(connectionState.close(), with: newConnection.connectionID, on: eventLoop))
```
Lukasa, line 470: This logic appears to be repeated in a number of places. Consider factoring it out.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                if self.connections.isEmpty {
+                    self.state = .shutDown
+                    return .init(.none, .cleanupConnection(close: [], cancel: [], isShutdown: .yes(unclean: unclean)))
+                } else {
+                    return .init(.none, .none)
+                }
+
+            case .shutDown:
+                preconditionFailure("The pool is already shutdown all connections must already been torn down")
+            }
```
Lukasa, line 540: This entire method seems extremely common to `http1ConnectionReleased`. I think this continues to speak to the question of whether some of this code can be factored out.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+                if let waiter = connectionState.removeStartWaiter() {
+                    requestAction = .failRequest(waiter.request, HTTPClientError.cancelled, cancelWaiter: waiter.requestID)
+                }
+                self.connections[connectionIndex] = connectionState
+
+                return .init(requestAction, .none)
+            }
+
+            // 2. check waiters in queue
+            let waiterIndex = self.waiters.firstIndex(where: { $0.requestID == requestID })
```
Lukasa, line 604: Cancel and timeout here appear to be identical except for error type. Refactor to be common.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+extension HTTPConnectionPool {
+    struct HTTP1ConnectionState {
+        enum State {
+            case starting(Waiter?)
+            case active(Connection, isAvailable: Bool, lastReturn: NIODeadline)
```
glbrntt, line 21: What's `lastReturn`, the point in time the connection was returned to the pool?

glbrntt: Is `lastReturn` ever relevant when `isAvailable` is `false`?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+        mutating func started(_ connection: Connection) -> Waiter? {
+            guard case .starting(let waiter) = self.state else {
+                preconditionFailure("Invalid state: \(self.state)")
+            }
+            self.state = .active(connection, isAvailable: true, lastReturn: .now())
+            return waiter
+        }
+
+        mutating func failed() -> Waiter? {
+            guard case .starting(let waiter) = self.state else {
```
glbrntt, line 83: This looks like it only applies to a newly started connection; what about if the connection drops while active?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+        private(set) var waiters: CircularBuffer<Waiter>
+        private(set) var state: State = .running
+
+        init(idGenerator: Connection.ID.Generator, maximumConcurrentConnections: Int) {
+            self.idGenerator = idGenerator
+            self.maximumConcurrentConnections = maximumConcurrentConnections
+            self.connections = []
+            self.connections.reserveCapacity(self.maximumConcurrentConnections)
+            self.waiters = []
+            self.waiters.reserveCapacity(32)
```
glbrntt, line 156: nit: `CircularBuffer(initialCapacity: 32)`

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+        mutating func executeRequest(_ request: HTTPSchedulableRequest, onPreffered prefferedEL: EventLoop, required: Bool) -> Action {
+            var eventLoopMatch: (Int, NIODeadline)?
+            var goodMatch: (Int, NIODeadline)?
+
+            switch self.state {
+            case .running:
+                break
+            case .shuttingDown, .shutDown:
+                // it is fairly unlikely that this condition is met, since the ConnectionPoolManager
+                // also fails new requests immidiatly, if it is shutting down. However there might
```
glbrntt, line 168:
```suggestion
                // also fails new requests immediately, if it is shutting down. However there might
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
glbrntt: +1

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
glbrntt: I'm wondering if this is better expressed as

```swift
enum Availability {
  case available(since: NIODeadline)
  case unavailable
}

enum State {
  case active(Connection, Availability)
  // ...
}
```

The `availableAndLastReturn` could then be `availableSince` (I found the use of "return" in the various names in `executeRequest` quite unclear)

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+State.swift
```diff
@@ -0,0 +1,242 @@
...
+                preconditionFailure("Invalid state. Use closure to modify state")
+            }
+        }
+        return try closure(&self)
+    }
+
+    mutating func modify(with action: HTTPConnectionPool.StateMachine.Action)
+        -> HTTPConnectionPool.StateMachine.Action {
+        return action
+    }
```
glbrntt, line 229: What's the point of this?

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
fabianfett: I changed it to this, which I like even better:

```swift
        enum State {
            case starting(Waiter?)
            case available(Connection, since: NIODeadline)
            case leased(Connection)
            case failed
            case closed
        }
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
```diff
@@ -0,0 +1,694 @@
...
+        @discardableResult
+        mutating func lease() -> Connection {
+            guard case .active(let conn, isAvailable: true, let lastReturn) = self.state else {
+                preconditionFailure("Invalid state")
+            }
+            self.state = .active(conn, isAvailable: false, lastReturn: lastReturn)
+            return conn
+        }
+
+        mutating func release() {
```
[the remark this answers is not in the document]
fabianfett: because further down we pick the connection based on its id:

```swift
        mutating func http1ConnectionReleased(_ connectionID: Connection.ID) -> Action {
            guard let index = self.connections.firstIndex(where: { $0.connectionID == connectionID }) else {
                preconditionFailure("A connection that we don't know was released? Something is very wrong...")
            }

            var connectionState = self.connections[index]
            connectionState.release()
```

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+State.swift
```diff
@@ -0,0 +1,242 @@
...
+                }
+
+            case .modify:
+                preconditionFailure("Invalid state")
+            }
+        }
+
+        mutating func http1ConnectionReleased(_ connectionID: Connection.ID) -> Action {
+            guard case .http1(var http1StateMachine) = self.state else {
+                preconditionFailure("Invalid state")
```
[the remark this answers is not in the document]
Lukasa: I would generally recommend throwing in cases where we believe it's possible to trigger the event accidentally: i.e. because someone using the program got confused, or from the network. In this case I believe the precondition is here because an outer state machine should prevent us getting here at all.

Nonetheless, I would recommend we switch exhaustively over the states, even if we crash in the other cases.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+HTTP1State.swift
fabianfett: The action is also different. In the timeout case, we need to cancel the request. If the request was cancelled, we don't need to fail the request anymore. I personally would like to keep them separate. But not a hill I want to die on.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1.1/HTTP1ConnectionStateMachine.swift
```diff
@@ -387,7 +387,7 @@ extension HTTP1ConnectionStateMachine.State {
             case .initialized:
                 preconditionFailure("Invalid state")
             case .idle:
-                preconditionFailure("How can we fail a task, if we are idle")
+                preconditionFailure("How can we fail a request, if we are idle")
```
Davidde94, line 390: This is such an incredible nit, give me a slap if you want

```suggestion
                preconditionFailure("Can't fail an idle request")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+        var availableAndLastReturn: NIODeadline? {
+            switch self.state {
+            case .available(_, since: let lastReturn):
+                return lastReturn
+            case .backingOff, .starting, .leased, .failed, .closed:
+                return nil
+            }
+        }
+
+        mutating func started(_ connection: Connection) {
```
Davidde94, line 85: I think this reads nicer as something thing `state.connectionStarted(connection)`. You're telling the state machine that something started, so tell it what.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+            case .starting:
+                self.state = .available(connection, since: .now())
+            case .backingOff, .available, .leased, .failed, .closed:
+                preconditionFailure("Invalid state: \(self.state)")
+            }
+        }
+
+        /// The connection failed to start
+        /// - Returns: How often the connection failed to start. Use this int to calculate backoff intervals.
+        mutating func failedToStart() {
```
Davidde94, line 96: Same here - `connectionFailedToStart`.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+            for connectionState in self.connections {
+                if connectionState.isConnecting {
+                    stats.connecting += 1
+                } else if connectionState.isBackingOff {
+                    stats.backingOff += 1
+                } else if connectionState.isLeased {
+                    stats.leased += 1
+                } else if connectionState.isAvailable {
+                    stats.idle += 1
+                }
```
Davidde94, line 197: There must be a nicer way to do this!

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+
+        mutating func createNewConnection(on eventLoop: EventLoop) -> Connection.ID {
+            let connection = HTTP1ConnectionState(connectionID: self.generator.next(), eventLoop: eventLoop)
+            self.connections.append(connection)
+            self.overflowIndex = self.connections.index(after: self.overflowIndex)
+            return connection.connectionID
+        }
+
+        mutating func createNewOverflowConnection(on eventLoop: EventLoop) -> Connection.ID {
+            preconditionFailure()
```
Davidde94, line 232: We should probably have a message here

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+        }
+
+        mutating func newHTTP1ConnectionCreated(_ connection: Connection) -> Connection.ID {
+            guard let index = self.connections.firstIndex(where: { $0.connectionID == connection.id }) else {
+                preconditionFailure("There is a new connection, that we didn't request!")
```
Davidde94, line 237:
```suggestion
                preconditionFailure("There is a new connection that we didn't request!")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+            }
+
+            self.connections[index].started(connection)
+            return connection.id
+        }
+
+        mutating func backoffNextConnectionAttempt(_ connectionID: Connection.ID) -> EventLoop {
+            guard let index = self.connections.firstIndex(where: { $0.connectionID == connectionID }) else {
+                preconditionFailure("We tried to create a new connection, that we know nothing about?")
```
Davidde94, line 246:
```suggestion
                preconditionFailure("We tried to create a new connection that we know nothing about?")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+            }
+
+            let newConnection = HTTP1ConnectionState(
+                connectionID: self.generator.next(),
+                eventLoop: self.connections[index].eventLoop
+            )
+
+            self.connections[index] = newConnection
+            return (newConnection.connectionID, newConnection.eventLoop)
```
Davidde94, line 264: Should we first check if the connection is actually failed or closed?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+                    // the current best match. We continue the iteration.
+                    continue
+                }
+            }
+
+            if let (index, _) = match {
+                return index
+            }
+
+            return nil
```
Davidde94, line 428: Very subjective in terms of neatness, but can't you just `return match?.0`?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1StateMachine.swift
```diff
@@ -0,0 +1,327 @@
...
+                // also fails new requests immediately, if it is shutting down. However there might
+                // be race conditions in which a request passes through a running connection pool
+                // manager, but hits a connection pool that is already shutting down.
+                //
+                // (Order in one lock does not guarantee order in the next lock!)
+                return .init(.failRequest(request, HTTPClientError.alreadyShutdown, cancelWaiter: nil), .none)
+            }
+
+            if required {
+                preconditionFailure("EventLoop requirements not supported as of today.")
```
Davidde94, line 63: I feel like it's a little pointless to have `required` as a parameter if we can't do anything with it

Sources/AsyncHTTPClient/ConnectionPool/State Machine/StateMachine.swift
```diff
@@ -0,0 +1,275 @@
...
+            let connection: ConnectionAction
+
+            init(_ request: RequestAction, _ connection: ConnectionAction) {
+                self.request = request
+                self.connection = connection
+            }
+
+            static var none: Action {
+                return .init(.none, .none)
+            }
```
Davidde94, line 31: What's the difference between this and just allowing returned actions to be `nil`?

fabianfett: In my opinion, making this an optional make everything harder. The option to switch like I do on an enum was also added more recently. Before you would need to do this:

```swift
enum Action {
  case foo
  case bar
}

func run(_ action: Action?) {
  switch action {
  case .some(.foo):
    break
  case .some(.bar):
    break
  case .none:
    break
  }
}
```

Which is ugly. Also autocomplete works much nicer, if it is not an optional, but an enum with a `none` case.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
fabianfett: Yes, I could do this. However my brain is often very slow. And parsing the shorthand syntax and translating it into the above takes a bit of time. For this reason, I personally prefer to go the verbose way. Not a hill I want to die on.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
fabianfett: If a connection fails in startup or is closed, I currently don't bother to set its state. Since the connections are in an ever changing array this would mean iterating over the connections twice, or exposing the array index. Neither of which do I like right now. In the `MockConnectionPool` (the structure I use for testing), I very much verify the state on every step.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+
+            // To find an appropriate connection we iterate all existing connections.
+            // While we do this we try to find the best fitting connection for our request.
+            //
+            // A perfect match, runs on the same eventLoop and has been idle the shortest amount
+            // of time.
+            //
+            // An okay match is not on the same eventLoop, and has been idle for the shortest
+            // time.
+            for (index, conn) in self.connections.enumerated() {
```
glbrntt, line 353: (fwiw: another place where we rely on `Array` indexing)

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
fabianfett: I guess normally I would agree. However we are down here to a single connection. and an invocation would therefore read like this `connectionState.connectionStarted(connectionBox)`. This is quite a number of connections in one line.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
fabianfett: I guess normally I would agree. However we are down here to a single connection. and an invocation would therefore read like this connectionState.connectionFailedToStart(). This is quite a number of connections in one line.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTP1Connections.swift
```diff
@@ -0,0 +1,438 @@
...
+
+        private func findAvailableConnection(onPreferred preferredEL: EventLoop) -> Int? {
+            var eventLoopMatch: (Int, NIODeadline)?
+            var goodMatch: (Int, NIODeadline)?
+
+            // To find an appropriate connection we iterate all existing connections.
+            // While we do this we try to find the best fitting connection for our request.
+            //
+            // A perfect match, runs on the same eventLoop and has been idle the shortest amount
+            // of time.
```
glbrntt, line 349: Might be worth adding that "idle the shortest amount of time" means "returned the most recently".

Tests/AsyncHTTPClientTests/Mocks/MockConnectionPool.swift
```diff
@@ -0,0 +1,633 @@
...
+            case closed
+        }
+
+        let id: ID
+        let eventLoop: EventLoop
+
+        private(set) var state: State = .starting
+        private(set) var isParked: Bool = false
+
+        init(id: ID, eventLoop: EventLoop) {
```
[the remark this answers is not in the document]
fabianfett: No, because I want to have a lot of different ELs that all belong to the same EventLoopGroup, to test the best connections are picked.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/StateMachine.swift
```diff
@@ -0,0 +1,275 @@
...
+            case scheduleWaiterTimeout(RequestID, HTTPSchedulableRequest, on: EventLoop)
+            case cancelWaiterTimeout(RequestID)
+
+            case none
+        }
+
+        enum HTTPTypeStateMachine {
+            case http1(HTTP1StateMachine)
+
+            case modify
```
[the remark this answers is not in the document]
fabianfett: I take both!

PR #404, opened 2021-07-23
Cleanup: Connection cancel -> shutdown

fabianfett:
### Motivation

- It is unclear what it means to cancel a connection

### Modifications

- Rename `cancel` connection to `shutdown`, with an explicit code comment
- Add `.http1_1` and `.http2` cases to `HTTPConnectionPool.Connection`

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
@@ -60,10 +60,10 @@ enum HTTPConnectionPool {
             }
         }
 
-        fileprivate func execute(request: HTTPExecutableRequest) {
+        fileprivate func executeRequest(_ request: HTTPExecutableRequest) {
             switch self._ref {
             case .http1_1(let connection):
-                return connection.execute(request: request)
+                return connection.executeRequest(request)
             case .http2(let connection):
                 return connection.executeRequest(request)
             case .__testOnly_connection:
@@ -83,7 +83,8 @@ enum HTTPConnectionPool {
             }
         }
 
-        /// Closes the connection without cancelling running requests.
+        /// Closes the connection without cancelling running requests. Use this when you are sure, that the
+        /// connection is currently idle.
         fileprivate func close() -> EventLoopFuture<Void> {
             switch self._ref {
             case .http1_1(let connection):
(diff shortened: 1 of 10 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -18,89 +18,102 @@ enum HTTPConnectionPool {
...
             switch self._ref {
-//            case .http1_1(let connection):
-//                return connection.close()
-
-            case .__testOnly_connection(_, let eventLoop):
-                return eventLoop.makeSucceededFuture(())
+            case .http1_1(let connection):
+                return connection.execute(request: request)
+            case .http2(let connection):
+                return connection.executeRequest(request)
```
glbrntt, line 68: That's some mildly irritating asymmetry :)

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -18,89 +18,102 @@ enum HTTPConnectionPool {
...
+                return connection.shutdown()
+            case .http2(let connection):
+                return connection.shutdown()
             case .__testOnly_connection:
                 break
             }
         }
 
-        fileprivate func cancel() {
+        /// Closes the connection without cancelling running requests.
```
glbrntt, line 86: What happens to the requests if we don't cancel them? Why would do use this over shutdown?

glbrntt (approved):
LGTM.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
fabianfett: Fixed.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
fabianfett: In the pool state machine, we know if a connection is idle or not. For this reason, we can use the direct way to close, if we are sure nothing is running on the connection.

glbrntt (approved):

Lukasa (approved):

PR #405, opened 2021-08-11
Fix bi directional streaming test

fabianfett:
### Motivation

The `HTTPEchoHandler` has a property `var promises: CircularBuffer<EventLoopPromise<Void>>` that is currently accessed without lock from different eventLoops (`self.serverGroup` and `self.defaultClient.eventLoopGroup`). This in itself may not be a problem. However testing bi directional streaming should not depend on shared promises. Instead the client should write something to an echo server, wait for the body to be echoed and write some more.

### Changes

- bi directional streaming, writes and reads from an echo server.

### Result

No weird shared state in promises between client and server.

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
@@ -1149,23 +1149,144 @@ struct CollectEverythingLogHandler: LogHandler {
     }
 }
 
+class StreamDelegate: HTTPClientResponseDelegate {
+    typealias Response = Void
+
+    enum State {
+        case idle
+        case waitingForBytes(EventLoopPromise<ByteBuffer?>)
+        case buffering(ByteBuffer, done: Bool)
+        case failed(Error)
+        case finished
+    }
+
+    let eventLoop: EventLoop
+    private var state: State = .idle
+
+    init(eventLoop: EventLoop) {
+        self.eventLoop = eventLoop
+    }
+
+    func next() -> EventLoopFuture<ByteBuffer?> {
+        if self.eventLoop.inEventLoop {
+            return self.next0()
+        } else {
+            return self.eventLoop.flatSubmit {
+                self.next0()
+            }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


Lukasa:
FWIW the shared state isn't an issue: NIO's multithreading rules mean there is no data race here.

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -1149,23 +1149,144 @@ struct CollectEverythingLogHandler: LogHandler {
...
+    func didSendRequestHead(task: HTTPClient.Task<Response>, _ head: HTTPRequestHead) {
+        XCTAssert(self.eventLoop.inEventLoop)
+    }
+
+    func didSendRequestPart(task: HTTPClient.Task<Response>, _ part: IOData) {
+        XCTAssert(self.eventLoop.inEventLoop)
+    }
+
+    func didSendRequest(task: HTTPClient.Task<Response>) {
+        XCTAssert(self.eventLoop.inEventLoop)
```
Lukasa, line 1218: Can we remove this assertion? `inEventLoop` is validly allowed to return false negatives, so we shouldn't assert that it won't. We can add `preconditionInEventLoop` instead if you want the check.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2815,40 +2815,65 @@ class HTTPClientTests: XCTestCase {
...
+            func writeLoop(_ writer: HTTPClient.Body.StreamWriter, index: Int) {
+                XCTAssert(writeEL.inEventLoop, "Always write from unexpected el")
+
+                if index >= 30 {
+                    return finalPromise.succeed(())
+                }
+
+                let sent = ByteBuffer(integer: index)
+                writer.write(.byteBuffer(sent)).flatMap { () -> EventLoopFuture<ByteBuffer?> in
+                    XCTAssert(delegateEL.inEventLoop, "Always dispatch back to delegate el")
```
Lukasa, line 2848: Same note here about false positives.

Lukasa (approved):

PR #406, opened 2021-08-16
[RFC] Add async/await proposal

fabianfett:
In this pr we outline our vision for an `async/await` API. We would love to hear your thoughts.

Related work: https://github.com/grpc/grpc-swift/pull/1231

### DIFF
--- a/docs/async-await.md
+++ b/docs/async-await.md
@@ -12,12 +12,12 @@ This proposal describes what these APIs could look like and explores some of the
 
 ## Proposed API additions
 
-### New `AsyncRequest` type
+### New `HTTPClientRequest` type
 
-The proposed new `AsyncRequest` shall be a simple swift structure.
+The proposed new `HTTPClientRequest` shall be a simple swift structure.
 
 ```swift
-struct AsyncRequest {
+struct HTTPClientRequest {
   /// The requests url.
   var url: String
   
@@ -39,7 +39,7 @@ struct AsyncRequest {
 }
 ```
 
-A notable change from the current [`HTTPRequest`][HTTPRequest] is that the url is not of type `URL`. This makes the creation of a request non throwing. Existing issues regarding current API: 
+A notable change from the current [`HTTPClient.Request`][HTTPClient.Request] is that the url is not of type `URL`. This makes the creation of a request non throwing. Existing issues regarding current API: 
 
 - [HTTPClient.Request.url is a let constant][issue-395]
 - [refactor to make request non-throwing](https://github.com/swift-server/async-http-client/pull/56)
@@ -51,18 +51,18 @@ In normal try/catch flows this should not change the control flow:
 
 ```swift
 do {
-  var request = AsyncRequest(url: "invalidurl")
+  var request = HTTPClientRequest(url: "invalidurl")
   try await httpClient.execute(request, deadline: .now() + .seconds(3))
 } catch {
   print(error)
 }
 ```
 
-If the library code throws from the AsyncRequest creation or the request invocation the user will, in normal use cases, handle the error in the same catch block. 
+If the library code throws from the `HTTPClientRequest` creation or the request invocation the user will, in normal use cases, handle the error in the same catch block. 
 
-#### Body streaming
+#### Request body streaming
 
-The new AsyncRequest has a new body type, that is wrapper around an internal enum. This allows us to evolve this type for use-cases that we are not aware of today. 
+The new `HTTPClientRequest` has a new body type, that is wrapper around an internal enum. This allows us to evolve this type for use-cases that we are not aware of today. 
 
 ```swift
 public struct Body {
@@ -74,16 +74,16 @@ public struct Body {
 }
 ```
 
-The main difference to today's `Request.Body` type is the lack of a `StreamWriter` for streaming scenarios. The existing StreamWriter offered the user an API to write into (thus the user was in control of when writing happened). The new `AsyncRequest.Body` uses `AsyncSequence`s to stream requests. By iterating over the provided AsyncSequence, the HTTPClient is  in control when writes happen, and can ask for more data efficiently. 
+The main difference to today's `Request.Body` type is the lack of a `StreamWriter` for streaming scenarios. The existing StreamWriter offered the user an API to write into (thus the user was in control of when writing happened). The new `HTTPClientRequest.Body` uses `AsyncSequence`s to stream requests. By iterating over the provided AsyncSequence, the HTTPClient is  in control when writes happen, and can ask for more data efficiently. 
 
 Using the `AsyncSequence` from the Swift standard library as our upload stream mechanism dramatically reduces the learning curve for new users.
 
-### New `AsyncResponse` type
+### New `HTTPClientResponse` type
 
-The `AsyncResponse` looks more similar to the existing `Response` type. The biggest difference is again the `body` property, which is now an `AsyncSequence` of `ByteBuffer`s instead of a single optional `ByteBuffer?`. This will make every response on AsyncHTTPClient streaming by default.
+The `HTTPClientResponse` looks more similar to the existing [`HTTPClient.Response`][HTTPClient.Response] type. The biggest difference is again the `body` property, which is now an `AsyncSequence` of `ByteBuffer`s instead of a single optional `ByteBuffer?`. This will make every response on AsyncHTTPClient streaming by default. As with `HTTPClientRequest`, we dropped the namespacing on `HTTPClient` to allow easier discovery with autocompletion.
 
 ```swift
-public struct AsyncResponse {
+public struct HTTPClientResponse {
   /// the used http version
   public var version: HTTPVersion
   /// the http response status
@@ -94,7 +94,7 @@ public struct AsyncResponse {
   public var body: Body
 }
 
-extension AsyncResponse {
+extension HTTPClientResponse {
   public struct Body: AsyncSequence {
     public typealias Element = ByteBuffer
     public typealias AsyncIterator = Iterator
@@ -110,7 +110,11 @@ extension AsyncResponse {
 }
 ```
 
-At a later point we could add trailers to the AsyncResponse as effectful properties.
+Note: The user must consume the `Body` stream or drop the `HTTPClientResponse`, to ensure that the 
+internal HTTPClient connection can move forward. Dropping the `HTTPClientResponse` would lead to a 
+request cancellation which in turn would lead to a close of an exisiting HTTP/1.1 connection.
+
+At a later point we could add trailers to the `HTTPClientResponse` as effectful properties:
 
 ```swift
     public var trailers: HTTPHeaders? { async throws }
@@ -120,15 +124,16 @@ However we will need to make sure that the user has consumed the body stream com
 
 ```swift
 do {
-  var request = AsyncRequest(url: "https://swift.org/")
+  var request = HTTPClientRequest(url: "https://swift.org/")
   let response = try await httpClient.execute(request, deadline: .now() + .seconds(3))
   
   var trailers = try await response.trailers // can not move forward since body must be consumed before.
 } catch {
   print(error)
 }
-
 ```
+
+In such a case we can either throw an error or crash.
 
 ### New invocation
 
@@ -136,7 +141,7 @@ The new way to invoke a request shall look like this:
 
 ```swift
 extension HTTPClient {
-  func execute(_ request: AsyncRequest, deadline: NIODeadline) async throws -> AsyncResponse
+  func execute(_ request: HTTPClientRequest, deadline: NIODeadline) async throws -> HTTPClientResponse
 }
 ```
 
@@ -165,4 +170,5 @@ extension HTTPClient {
 [issue-392]: https://github.com/swift-server/async-http-client/issues/392
 [issue-395]: https://github.com/swift-server/async-http-client/issues/395
 
-[HTTPRequest]: https://github.com/swift-server/async-http-client/blob/main/Sources/AsyncHTTPClient/HTTPHandler.swift#L96-L318
+[HTTPClient.Request]: https://github.com/swift-server/async-http-client/blob/main/Sources/AsyncHTTPClient/HTTPHandler.swift#L96-L318
+[HTTPClient.Response]: https://github.com/swift-server/async-http-client/blob/main/Sources/AsyncHTTPClient/HTTPHandler.swift#L320-L364
--- a/docs/async-await.md
+++ b/docs/async-await.md
@@ -150,7 +150,7 @@ extension HTTPClient {
           let response = try await httpClient.execute(request, deadline: .distantFuture)
         }
         
-        Task.sleep(500 * 1000 * 1000) // wait half a second
+        await Task.sleep(nanosecond: 500 * 1000 * 1000) // wait half a second
         task.cancel() // cancel the task after half a second
         ```
 
--- a/docs/async-await.md
+++ b/docs/async-await.md
@@ -65,12 +65,14 @@ If the library code throws from the `HTTPClientRequest` creation or the request 
 The new `HTTPClientRequest` has a new body type, that is wrapper around an internal enum. This allows us to evolve this type for use-cases that we are not aware of today. 
 
 ```swift
-public struct Body {
-  static func bytes<S: Sequence>(_ sequence: S) -> Body where S.Element == UInt8
-
-  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == ByteBuffer
-
-  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == UInt8
+extension HTTPClientRequest {
+  public struct Body {
+    static func bytes<S: Sequence>(_ sequence: S) -> Body where S.Element == UInt8
+  
+    static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == ByteBuffer
+  
+    static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == UInt8
+  }
 }
 ```
 
@@ -142,6 +144,27 @@ The new way to invoke a request shall look like this:
 ```swift
 extension HTTPClient {
   func execute(_ request: HTTPClientRequest, deadline: NIODeadline) async throws -> HTTPClientResponse
+}
+```
+
+Usage example:
+
+```swift
+var request = HTTPClientRequest(url: "https://swift.org")
+request.method = .POST
+request.headers = [
+  "content-type": "text/plain; charset=UTF-8"
+  "x-my-fancy-header": "super-awesome"
+]
+request.body = .sequence("Hello world!".utf8)
+
+var response = try await client.execute(request, deadline: .now() + .seconds(5))
+
+switch response.status {
+case .ok:
+  let body = try await response.body.collect(maxBytes: 1024 * 1024)
+default:
+  throw MyUnexpectedHTTPStatusError
 }
 ```
 
--- a/docs/async-await.md
+++ b/docs/async-await.md
@@ -113,7 +113,7 @@ extension AsyncResponse {
 At a later point we could add trailers to the AsyncResponse as effectful properties.
 
 ```swift
-    public var trailers: HTTPHeaders { async throws }
+    public var trailers: HTTPHeaders? { async throws }
 ```
 
 However we will need to make sure that the user has consumed the body stream completely before, calling the trailers, because otherwise we might run into a situation from which we can not progress forward:
--- a/docs/async-await.md
+++ b/docs/async-await.md
@@ -1,0 +1,168 @@
+# Proposal: Async-await support
+
+## Introduction
+
+With the introduction of [async/await][SE-0296] in Swift 5.5, it is now possible to write asynchronous code without the need for callbacks. 
+
+Language support for [`AsyncSequence`][SE-0298] also allows for writing functions that return values over time.
+
+We would like to explore how we could offer APIs that make use of these new language features to allow users to run HTTPRequest using these new idioms.
+
+This proposal describes what these APIs could look like and explores some of the potential usability concerns.
+
+## Proposed API additions
+
+### New `AsyncRequest` type
+
+The proposed new `AsyncRequest` shall be a simple swift structure.
+
+```swift
+struct AsyncRequest {
+  /// The requests url.
+  var url: String
+  
+  /// The request's HTTPMethod
+  var method: HTTPMethod
+  
+  /// The request's headers
+  var headers: HTTPHeaders
+
+  /// The request's body
+  var body: Body?
+    
+  init(url: String) {
+    self.url = url
+    self.method = .GET
+    self.headers = .init()
+    self.body = .none
+  }
+}
+```
+
+A notable change from the current [`HTTPRequest`][HTTPRequest] is that the url is not of type `URL`. This makes the creation of a request non throwing. Existing issues regarding current API: 
+
+- [HTTPClient.Request.url is a let constant][issue-395]
+- [refactor to make request non-throwing](https://github.com/swift-server/async-http-client/pull/56)
+- [improve request validation](https://github.com/swift-server/async-http-client/pull/67)
+
+The url validation will become part of the normal request validation that occurs when the request is scheduled on the `HTTPClient`. If the user supplies a request with an invalid url, the http client, will reject the request.
+
+In normal try/catch flows this should not change the control flow:
+
+```swift
+do {
+  var request = AsyncRequest(url: "invalidurl")
+  try await httpClient.execute(request, deadline: .now() + .seconds(3))
+} catch {
+  print(error)
+}
+```
+
+If the library code throws from the AsyncRequest creation or the request invocation the user will, in normal use cases, handle the error in the same catch block. 
+
+#### Body streaming
+
+The new AsyncRequest has a new body type, that is wrapper around an internal enum. This allows us to evolve this type for use-cases that we are not aware of today. 
+
+```swift
+public struct Body {
+  static func bytes<S: Sequence>(_ sequence: S) -> Body where S.Element == UInt8
+
+  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == ByteBuffer
+
+  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == UInt8
+}
+```
+
+The main difference to today's `Request.Body` type is the lack of a `StreamWriter` for streaming scenarios. The existing StreamWriter offered the user an API to write into (thus the user was in control of when writing happened). The new `AsyncRequest.Body` uses `AsyncSequence`s to stream requests. By iterating over the provided AsyncSequence, the HTTPClient is  in control when writes happen, and can ask for more data efficiently. 
+
+Using the `AsyncSequence` from the Swift standard library as our upload stream mechanism dramatically reduces the learning curve for new users.
+
+### New `AsyncResponse` type
+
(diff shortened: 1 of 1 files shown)


fabianfett:
cc @toddvarland @ktoso @tomerd @tachyonics @0xTim @gwynne @adam-fowler

fabianfett:
cc @slashmo @kmahar @kneekey23

adam-fowler:
In general, I'm quite happy with this. It's a good idea to use `AsyncSequence` as the stream primitive and that you don't plan to wrap it in your own AHC type. This allows for much more user flexibility in how user can provide the data to AHC. I had planned to do the same for Soto. It makes my life easier now that AHC is doing the heavy lifting of reading from `AsyncSequence`. 

The only thing I think is missing is `AsyncResponse` could do with a helper function to collate all the response body buffers into one. I know this is only a couple of lines of code but this would be such a common piece of functionality that I think it is required.

There are still questions to be asked about task locals vs passing the Logger about and also the tier system but I guess this document isn't about those.

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+
+The new AsyncRequest has a new body type, that is wrapper around an internal enum. This allows us to evolve this type for use-cases that we are not aware of today. 
+
+```swift
+public struct Body {
+  static func bytes<S: Sequence>(_ sequence: S) -> Body where S.Element == UInt8
+
+  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == ByteBuffer
+
+  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == UInt8
```
dnadoba, line 73: How do you imagine to store the sequence in the internal enum? I think we would probably need to erase it with some thing like `AnyAsyncSequence`, similar to what `AnySequence` does for `Sequence`. But as discussed in [this thread](https://forums.swift.org/t/anyasyncsequence/50828), there is no `AnyAsyncSequence` in the standard library and `AnySequence` is already slow. `AnyAsyncSequence` would probably be even slower. I therefore think we can't get reasonable performance in the case where the element is of type `UInt8` because of the overhead introduced by erasing the type of the sequence. They only way I can currently think of making this fast would require making AsyncRequest generic over the actual Body type, which has it own downsides.

fabianfett:
@adam-fowler 

> The only thing I think is missing is AsyncResponse could do with a helper function to collate all the response body buffers into one. I know this is only a couple of lines of code but this would be such a common piece of functionality that I think it is required.

https://github.com/apple/swift-nio/pull/1939

docs/async-await.md
Lukasa: The way we make it fast in case of `AsyncSequence where Element == UInt8` is by having an inlinable transformer from UInt8 to buffer. That is, a wrapper sequence that builds up buffers of, say, 1kB in size (or whatever) before it passes it on. We then wrap _that_ sequence in the body.

So long as the transformer is inlinable we can have it specialized by the caller. We can then wrap it in our type erasing logic (simply a callback of type `() async throws -> ByteBuffer?`) and voila, we get pretty good straight line performance. It's not _perfect_, but it should be pretty good.

fabianfett: A proof of concept has been implemented here... https://github.com/fabianfett/async-http-client/blob/ff-async/Sources/AsyncHTTPClient/AsyncAwait/AsyncRequest.swift#L74-L87

In the `static` factory methods we translate the `AsyncSequence` into a callback, that makes use of the specialization.
```swift
static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == UInt8 {
    var iterator = sequence.makeAsyncIterator()
    let body = self.init(.asyncSequence { allocator -> IOData? in
        var buffer = allocator.buffer(capacity: 1024) // TODO: Magic number
        while buffer.writableBytes > 0, let byte = try await iterator.next() {
            buffer.writeInteger(byte)
        }
        if buffer.readableBytes > 0 {
            return .byteBuffer(buffer)
        }
        return nil
    })
    return body
}
```

ktoso: > The way we make it fast in case of AsyncSequence where Element == UInt8 is by having an inlinable transformer from UInt8 to buffer. That is, a wrapper sequence that builds up buffers of, say, 1kB in size (or whatever) before it passes it on. We then wrap that sequence in the body.

Yeah that's what _any_ stream (IO related at least) will have to do and I've mentioned this repeatedly to folks working on IO and streams. The problem with our async story is that it's not clear who and where must do this chunking. So we may double-chunk un-necessarily... Say the "async bytes" abstraction https://developer.apple.com/documentation/foundation/filehandle/3766681-bytes does exactly that. So feeding a request from a file by doing `.bytes` on it and putting it into this request body _will work_ but we may end up with two such buffers that aggregate where we might have wanted one... but I digress, yes such buffers are exactly what we need to do at asynchronous boundaries of streams (such as here, where we get passed one).

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+}
+```
+
+The main difference to today's `Request.Body` type is the lack of a `StreamWriter` for streaming scenarios. The existing StreamWriter offered the user an API to write into (thus the user was in control of when writing happened). The new `AsyncRequest.Body` uses `AsyncSequence`s to stream requests. By iterating over the provided AsyncSequence, the HTTPClient is  in control when writes happen, and can ask for more data efficiently.
```
ktoso, line 77: We don't lose this capability actually, we just express it the right way (well, "right" in my brain at least) now :-)

One should always accept "some" async sequence. The way this sequence is produced matters not to the consumer of it (this API).

Now, the way a "writer" is expressed, is by the end-user creating an `Swift.AsyncStream` (which actually is more like "async sequence factory" though the name stuck). That stream then can be yielded into: `stream.yield(some bytes) -> YieldResult` and this way the end user may construct their own "writer style" API, and we don't have to prepare anything for it, since we only deal with it in terms of `AsyncSequence` 👍

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+
+  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == ByteBuffer
+
+  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == UInt8
+}
+```
+
+The main difference to today's `Request.Body` type is the lack of a `StreamWriter` for streaming scenarios. The existing StreamWriter offered the user an API to write into (thus the user was in control of when writing happened). The new `AsyncRequest.Body` uses `AsyncSequence`s to stream requests. By iterating over the provided AsyncSequence, the HTTPClient is  in control when writes happen, and can ask for more data efficiently. 
+
+Using the `AsyncSequence` from the Swift standard library as our upload stream mechanism dramatically reduces the learning curve for new users.
```
ktoso, line 79: I propose to reword the above about "losing" the writer approach, since we still have it -- thanks to AsyncStream, this way this is only simplification without feature loss 👍

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+
+### New `AsyncResponse` type
```
ktoso, line 81: (Again not in love with not mentioning HTTP here... AsyncHTTPResponse is nice and clear)

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+```swift
+public struct AsyncResponse {
+  /// the used http version
+  public var version: HTTPVersion
+  /// the http response status
+  public var status: HTTPResponseStatus
+  /// the response headers
+  public var headers: HTTPHeaders
+  /// the response payload as an AsyncSequence
+  public var body: Body
```
ktoso, line 94: It might be worth exploring an akka http entity inspired design here before we commit to this API. 
Specially for the response object it matters because as we create it we _know_ if there will be more bytes or not, so we can immediately construct the right "entity object", e.g. Empty or Strict(<all the bytes in mem>) or a stream.

Representing the body as an abstract `HTTPEntity` which may be "strict" or not allows us then to create store the strict bytes via reference to it directly, we don't have to use the stream then at all other than lazy creating it when someone pulls from it.

--- 

The entire topic of "there's a stream of bytes, but noone subscribed to it" is quite hellish in general.

What is the approach of this implementation about it? If we're doing a proper back-pressured stream it means that not reading it is directly connected to not reading from the wire, which may or may not be what the user intended. In akka we had to invent subscription timeouts to resolve such zombie streams.

In case I'm not clear about the case: response has streamed body; we're a nice streaming API and people expect back-pressure via such stream API. If we _never_ consume this stream, what happens?

(In Akka's case we were very loud that this is user error and one has to do `.discardBytes()` (but it's annoying))

--- 

Note though that I don't yet have good performance intuitions about AsyncSequences... perhaps it is not worth optimizing for knowing that there's a `.single(<bytes>)` at all since setup and logic costs are comparable between this and an iterator... So mostly sharing as an "are we _sure_ and have we considered alternatives?" note.

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+    public func makeAsyncIterator() -> Iterator
+  }
+}
+```
+
+At a later point we could add trailers to the AsyncResponse as effectful properties.
+
+```swift
+    public var trailers: HTTPHeaders { async throws }
+```
```
ktoso, line 117: Yeah just an async property is likely fine... not worth making those a stream as well I guess, as it's only a few of those if any.

Nitpick, `HTTPHeaders?` I guess, right?

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+
+However we will need to make sure that the user has consumed the body stream completely before, calling the trailers, because otherwise we might run into a situation from which we can not progress forward:
+
+```swift
+do {
+  var request = AsyncRequest(url: "https://swift.org/")
+  let response = try await httpClient.execute(request, deadline: .now() + .seconds(3))
+  
+  var trailers = try await response.trailers // can not move forward since body must be consumed before.
```
ktoso, line 126: Probably throw if it is known that body is not consumed yet?

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+        ```swift
+        let task = Task {
+          let response = try await httpClient.execute(request, deadline: .distantFuture)
+        }
+        
+        Task.sleep(500 * 1000 * 1000) // wait half a second
+        task.cancel() // cancel the task after half a second
+        ```
+
+- **What happens with all the other configuration options?** Currently users can configure a TLSConfiguration on a request.  This API doesn't expose this option. We hope to create a three layer model in the future. For this reason, we currently don't want to add per request configuration on the request invocation. More info can be found in the issue: [RFC: design suggestion: Make this a "3-tier library"][issue-392]
```
ktoso, line 157: Hah yeah sorry I promised to get to that one as welll... will try to make the time. But agreed on the more layers approach in general!

docs/async-await.md
fabianfett: > In case I'm not clear about the case: response has streamed body; we're a nice streaming API and people expect back-pressure via such stream API. If we never consume this stream, what happens?
>
> (In Akka's case we were very loud that this is user error and one has to do .discardBytes() (but it's annoying))

docs/async-await.md
fabianfett: Fixed.

docs/async-await.md
Lukasa: We don't have a choice, we _have_ to do it. The reason we have to do it is that we have to type-erase the sequence, and the moment we do that the compiler can't see through the loop. This means we can only iterate on the chunks. The better move, I think, would be for `.bytes` to be augmented by a `.chunkedBytes` API that returned some appropriate buffer type. This would allow adopters building pipelines to delay chunking until the last moment.

Of course, now we get into an argument about what that buffer type is. :wink:

docs/async-await.md
Lukasa: We do lose it: `AsyncStream.yield` is not isomorphic to `StreamWriter`. Importantly, `StreamWriter.write` _returns an `EventLoopFuture`_, which allows it to exert backpressure on the producer. `AsyncStream.Continuation.yield` is not an async function, and so it cannot exert backpressure in the same way. Users could retrofit it in a ghetto kinda way by using a buffered stream and observing the return values from `yield`. They can also retrofit it by using `AsyncStream(unfolding:onCancel:)` but that forces them to construct a closure and essentially "invert" their backpressure (from push-based to pull-based) which is fairly non-obvious.

gRPC is proposing an alternative API that keeps the `StreamWriter`-like design, which tbh I prefer.

docs/async-await.md
Lukasa: I think we should just crash.

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+    self.method = .GET
+    self.headers = .init()
+    self.body = .none
+  }
+}
+```
+
+A notable change from the current [`HTTPRequest`][HTTPRequest] is that the url is not of type `URL`. This makes the creation of a request non throwing. Existing issues regarding current API: 
+
+- [HTTPClient.Request.url is a let constant][issue-395]
```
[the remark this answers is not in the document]
fabianfett: Yes. It does. The actual link is in the end of the document.

docs/async-await.md
ktoso: We talked offline a little bit -- I think we're in a better place here, because the request's dropping can cancel (and does, but was not documented in this document), which leaves us without the nasty "hanging stream" issue 👍 

The must always consume rule is fine, we'll have to be explicit about it 👍

docs/async-await.md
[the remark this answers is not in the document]
ktoso: ah! got it, thanks

docs/async-await.md
ktoso: That's the entire reason behind the yielded values which we made sure to include for exactly this reason.

You're saying that it's dependent on the intermediate buffer being full or not, and not on the actual "end that does the write being full or not" but that's just the reality of dealing with buffered streams -- and _any_ stream may be buffered so that's it, you won't get the exact same semantics but these are the semantics you work with in stream APIs.

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+  var url: String
+  
+  /// The request's HTTPMethod
+  var method: HTTPMethod
+  
+  /// The request's headers
+  var headers: HTTPHeaders
+
+  /// The request's body
+  var body: Body?
```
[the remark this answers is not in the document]
weissi: @ktoso hmm, the above `HTTPEntity.*` list seems extremely HTTP/1.1 specific and even exposes some really bad things like "close delimited" which we don't need. Also a user shouldn't care if it's sent using chunked encoding or not, right? That's a choice the HTTP lib can make itself. If the user doesn't tell the number of bytes ahead of time, then in HTTP/1.1 it should use chunked encoding. If the user promises the size ahead of time, it can use content-length instead. It should never use EOF-framing (aka "close delimited").

- `Empty` seems like an (unnecessary?) specialisation of a body with a pre-announced length (of 0 bytes)
- `indefiniteLength` seems like an (unnecessary?) specialisation of a body without a pre-announced length

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+This proposal describes what these APIs could look like and explores some of the potential usability concerns.
+
+## Proposed API additions
+
+### New `AsyncRequest` type
+
+The proposed new `AsyncRequest` shall be a simple swift structure.
+
+```swift
+struct AsyncRequest {
```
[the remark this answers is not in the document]
fabianfett: Since `HTTPRequest`/`HTTPResponse` has a fairly high chance of naming conflicts, we opt for: `HTTPClientRequest` and `HTTPClientResponse`.

docs/async-await.md
adam-fowler: There is a minor issue implementing `StreamWriter` using `AsyncStream`. If I am looking for the following api
```swift
let stream = StreamWriter()
stream.write(buffer1)
stream.write(buffer2)
stream.write(nil)
```
I need the `StreamWriter` to be ready to accept ByteBuffers immediately. The construction of `AsyncStream` takes a `build` closure where you setup your stream. The api does not make it obvious that this closure will have been called by the time `AsyncStream.init` has finished. Having looked at the code I know this is the case but can I assume that will always be the case. Here is the best I could do to implement above
```swift
class StreamWriter {
    var write: (ByteBuffer?) -> ()
    let stream: AsyncStream<ByteBuffer>
    
    init() {
        var write: ((ByteBuffer?) -> ())?
        self.stream = .init { cont in
            write = { buffer in
                guard let buffer = buffer else { cont.finish(); return }
                cont.yield(buffer)
            }
        }
        self.write = write!
    }
}
```
I had to force unwrap the `write` closure which makes me unhappy.

weissi: I assume we'd fix the `StreamWriter` API, right? There's already  #291

weissi: problem statements in #194 and #264

docs/async-await.md
ktoso: Yeah agreed that we have to do it here. Interesting idea with offering a chunkedBytes to allow more control to users there, might be worth a radar :-)

docs/async-await.md
adam-fowler: Aren't we discussing removing it and replacing with `AsyncSequence` which in the end I'm all for. I do think supplying a stream implementation with back pressure would be helpful as `AsyncStream` isn't up to the job yet.

fabianfett:
The proposal was updated with new naming. `HTTPClient.AsyncRequest` has been replaced with `HTTPClientRequest` and `HTTPClient.AsyncResponse` has been replaced with `HTTPClientResponse`. We acknowledge that the new naming is very close to the current naming (just dropping a `.`). However by dropping the dot we will get types that are findable with autocompletion (just type "HTTP").

docs/async-await.md
ktoso: We know about that limitation and now ticketified the specific case for  it: https://bugs.swift.org/browse/SR-15076

docs/async-await.md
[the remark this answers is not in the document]
fabianfett: I spoke with @ktoso offline that there might be performance optimization potential in here. @ktoso mentioned, that in akka they could optimize for cases where they received the response body in the same read as the response head. In those cases they would not be required to start a stream, but just copy the bytes out.

We might be able to achieve the same if we changed NIOHTTP1. However we should not expose this optimizations to the user with an API option. For this reason we should stay with the currently proposed API.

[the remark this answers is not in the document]
weissi: @fabianfett you can do the same in `NIOHTTP1`. If you see `channelRead(.head(...))`, `channelRead(.body(...))`, `channelRead(.end(...))`, `channelReadComplete`, then you have everything in one go and you don't need to stream. If you get `channelReadComplete` before `channelRead(.end(...))` then you'd need to stream.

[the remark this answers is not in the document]
fabianfett: I agree. We don't need to change `NIOHTTP1` in this case. And we even use `channelReadComplete` today. This would entail some internal changes though. But could be a great optimization.

docs/async-await.md
fabianfett: I think I'm on the crash side as well. If the user consumes the trailers before the body, it will never be right. Crash seems to be the right tradeoff.

ktoso: Yeah that's true, unlikely it'll be sometimes correct, and even if it is people really must fix it. 👍

docs/async-await.md
```diff
@@ -0,0 +1,174 @@
...
+} catch {
+  print(error)
+}
+```
+
+In such a case we can either throw an error or crash.
+
+### New invocation
+
+The new way to invoke a request shall look like this:
```
glbrntt, line 140: Not exactly related the proposal, but will convenience methods be added similar to the existing api (e.g. `client.get(...)` etc.)?

docs/async-await.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+If the library code throws from the `HTTPClientRequest` creation or the request invocation the user will, in normal use cases, handle the error in the same catch block. 
+
+#### Request body streaming
+
+The new `HTTPClientRequest` has a new body type, that is wrapper around an internal enum. This allows us to evolve this type for use-cases that we are not aware of today. 
+
+```swift
+public struct Body {
```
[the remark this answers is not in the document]
fabianfett: the later, will fix...

docs/async-await.md
```diff
@@ -0,0 +1,174 @@
...
+  
+  var trailers = try await response.trailers // can not move forward since body must be consumed before.
+} catch {
+  print(error)
+}
+```
+
+In such a case we can either throw an error or crash.
+
+### New invocation
```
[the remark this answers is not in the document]
fabianfett: added

docs/async-await.md
```diff
@@ -65,12 +65,14 @@ If the library code throws from the `HTTPClientRequest` creation or the request
...
 
 ```swift
-public struct Body {
-  static func bytes<S: Sequence>(_ sequence: S) -> Body where S.Element == UInt8
-
-  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == ByteBuffer
-
-  static func stream<S: AsyncSequence>(_ sequence: S) -> Body where S.Element == UInt8
+  public struct Body {
```
ktoso, line 69: Ok good to clarify that thx

docs/async-await.md
fabianfett: Added a FAQ to answer this question.

docs/async-await.md
```diff
@@ -0,0 +1,232 @@
...
+  /// The request's HTTPMethod
+  var method: HTTPMethod
+  
+  /// The request's headers
+  var headers: HTTPHeaders
+
+  /// The request's body
+  var body: Body?
+    
+  init(url: String) {
```
glbrntt, line 33: Only allowing `url` to be set in the `init` might be a little too restrictive. This pattern makes a lot of sense for large configuration objects etc. where adding a new `init` for each new piece of configuration quickly becomes a burden for maintainers. 

I don't know if that's justified here since this is likely to be used much more frequently so we should make it easy to use even if the maintenance cost is higher. Moreover I suspect there's much less scope for adding new properties to the request.

On the other hand, the "convenience" APIs might cover this just fine.

docs/async-await.md
```diff
@@ -0,0 +1,232 @@
...
+```swift
+struct HTTPClientRequest {
+  /// The requests url.
+  var url: String
+  
+  /// The request's HTTPMethod
+  var method: HTTPMethod
+  
+  /// The request's headers
+  var headers: HTTPHeaders
```
glbrntt, line 28: I was wondering whether `HTTPHeaders` is the right type here. For http/2 connections we'll need to normalise the header names to be lowercased and repackage them into `HPACKHeaders` which isn't cheap (in gRPC we used to use the http2-to-1 codecs and deal with http/1 types; we saw a pretty big increase in performance from using http2 directly, most of which came from not converting the headers). It'd be a shame to have to incur that cost here as well.

Keeping `HTTPHeaders` (over `HPACKHeaders` or a new AHC-provided headers type) is probably the preferred option at the moment since it's more or less a currency type but doesn't feel like an optimal solution.

fabianfett: @ktoso @weissi @PeterAdams-A wdyt?

PeterAdams-A: In gRPC did you not just move the overhead to HTTP/1?  Ie, made HTTP/2 the default and now have to convert when on HTTP/1?

glbrntt: Yes, exactly. It's a much clearer cut decision for gRPC since http/2 is _much_ more common; so much so that the cost of doing the extra work for http/1 doesn't really matter.

PeterAdams-A: Presumably at the point where this is currently created we don't know which HTTP version we'll be using?  Would we need to do some sort of delayed creation?  I think we need to provide and easy version to use, even if we also have a fast option.

glbrntt: That's right, we need the headers before we know what version we're using. 

One option is AHC providing its own headers type which could wrap either an `HTTPHeaders` or `HPACKHeaders`.  This has the advantage of being able to change the backing type without breaking API if we decide to change our bias from http/1 to http/2 or vice versa. Of course this potentially makes things worse by adding a third headers type... 

Another would be making interop between `HTTPHeaders` and `HPACKHeaders` cheaper -- I'm not sure how possible this is though.

PeterAdams-A: Could make the third way a protocol with default implementations for both such that if we get the right one it's close to free, and wrong one is just the conversion we'd have to do anyway.

docs/async-await.md
```diff
@@ -168,6 +168,40 @@ default:
 }
 ```
 
+Stream upload and download example using the new `FileHandle` api:
```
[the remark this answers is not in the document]
karwa: I think what this refers to is the new async APIs on Foundation's `FileHandle` type.

See: https://developer.apple.com/documentation/foundation/filehandle/3766681-bytes

docs/async-await.md
```diff
@@ -0,0 +1,168 @@
...
+}
+```
+
+A notable change from the current [`HTTPRequest`][HTTPRequest] is that the url is not of type `URL`. This makes the creation of a request non throwing. Existing issues regarding current API: 
+
+- [HTTPClient.Request.url is a let constant][issue-395]
+- [refactor to make request non-throwing](https://github.com/swift-server/async-http-client/pull/56)
+- [improve request validation](https://github.com/swift-server/async-http-client/pull/67)
+
+The url validation will become part of the normal request validation that occurs when the request is scheduled on the `HTTPClient`. If the user supplies a request with an invalid url, the http client, will reject the request.
```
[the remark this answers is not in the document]
karwa: If you want the URL validation to be part of request creation, you could create a wrapper which initializes a request from a URL object but creates the request using a String. That should give you a reasonable guarantee that HTTPClient will not reject the request due to it being an invalid URL.

It would mean parsing the URL twice, but that’s relatively cheap, and there are potentially ways to avoid it if we get shared Strings in the standard library at some point (they are already part of the ABI). If the shared String is owned by a URL object, the URL parser can assume it is already a valid URL and recover the parsed URL object from the owner without parsing the String again or allocating fresh storage.

docs/async-await.md
```diff
@@ -0,0 +1,232 @@
...
+case .ok:
+  Task {
+    var streamIterator = response.body.makeAsyncIterator()
+    writeHandle.writeabilityHandler = { handle in
+      switch try await streamIterator.next() {
+      case .some(let buffer):
+        handle.write(buffer.readData(buffer.readableBytes)!)
+      case .none:
+        handle.close()
+      }
```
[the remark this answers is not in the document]
dnadoba: In general, yes because `response.body` implements `AsyncSequence`. However, in this specific example we want to support proper backpressure meaning that we don't want to read more from the body if the writeHandler is not ready to write more data. We therefore wait until we read the next buffer from `response.body` until `writeHandle.writeabilityHandler` is called.

docs/async-await.md
```diff
@@ -0,0 +1,232 @@
...
+}
+```
+
+- **Why do we have a deadline in the function signature?** 
+    Task deadlines are not part of the Swift 5.5 release. However we think that they are an important tool to not overload the http client accidentally. For this reason we will not default them.
+- **What happened to the Logger?** We will use Task locals to propagate the logger metadata. @slashmo and @ktoso are currently working on this.
+- **How does cancellation work?** Cancellation works by cancelling the surrounding task:
+
+        ```swift
+        let task = Task {
```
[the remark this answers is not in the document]
dnadoba: Swift already has special syntax to create a child task through `async let`. What you propose would create a detached Task. I think this would probably make it too easy and users would create detached Tasks just because there is a method for it, even if they would actually want a child Task through `async let`.

PR #412, opened 2021-08-31
Add a RequestQueue

fabianfett:
### Motivation

For our Connection Pool we need to queue requests.

### Changes

- Remove previously added `Waiter` type... Not needed anymore.
- Add `Request` box
- Add `RequestQueue`

### Result

Less code in the actual state machine.


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -123,6 +123,46 @@ enum HTTPConnectionPool {
...
+
+        init(_ request: HTTPSchedulableRequest) {
+            self.req = request
+        }
+
+        var id: HTTPConnectionPool.Request.ID {
+            HTTPConnectionPool.Request.ID(self.req, eventLoopRequirement: self.requiredEventLoop?.id)
+        }
+
+        var requiredEventLoop: EventLoop? {
```
glbrntt, line 151: Perhaps I'm misremembering but I think I've seen this same `switch` in a bunch of places; would an extension on whatever the type of `eventLoopPreference` is be worth having?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -0,0 +1,115 @@
...
+
+        var count: Int {
+            self.generalPurposeQueue.count
+        }
+
+        var isEmpty: Bool {
+            self.generalPurposeQueue.isEmpty
+        }
+
+        mutating func count(for eventLoop: EventLoop?) -> Int {
```
glbrntt, line 36: I think as a user I'd be very surprised that getting the `count` is `mutating` and potentially allocates! 

Perhaps there should be a non-mutating version of `withEventLoopQueue`, something like:

```swift
func withEventLoopQueueIfAvailable<Result>(
    for eventLoopID: EventLoopID, 
    _ closure: (CircularBuffer<Request>) -> Result
) -> Result?
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -0,0 +1,115 @@
...
+            if let eventLoop = eventLoop {
+                return self.withEventLoopQueue(for: eventLoop.id) { $0.count }
+            }
+            return self.generalPurposeQueue.count
+        }
+
+        mutating func isEmpty(for eventLoop: EventLoop?) -> Bool {
```
glbrntt, line 43: Same here re: mutating

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -0,0 +1,115 @@
...
+                    // TBD: This is slow. Do we maybe want something more sophisticated here?
+                    return self.generalPurposeQueue.remove(at: index)
+                }
+                return nil
+            }
+        }
+
+        mutating func removeAll() -> [Request] {
+            var result = [Request]()
+            result = self.eventLoopQueues.reduce(into: result) { partialResult, element in
```
glbrntt, line 90: Isn't this just `eventLoopQueues.flatMap { $1 }`?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -0,0 +1,115 @@
...
+                element.value.forEach { request in
+                    partialResult.append(request)
+                }
+            }
+
+            self.generalPurposeQueue.forEach { request in
+                result.append(request)
+            }
```
glbrntt, line 98: Use `result.append(contentsOf: self.generalPurposeQueue)` to avoid intermediate reallocs of `result`

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -0,0 +1,115 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+extension HTTPConnectionPool {
+    /// A struct to store all queued requests.
+    struct RequestQueue {
+        private var generalPurposeQueue: CircularBuffer<Request>
+        private var eventLoopQueues: [EventLoopID: CircularBuffer<Request>]
```
glbrntt, line 21: Minor: this would be mildly nicer using queues of your own type rather than `CircularBuffer`

glbrntt: Although it's really only the removal code that's duplicated. Don't mind whether you change it or not.

Tests/AsyncHTTPClientTests/HTTPConnectionPool+StateTestUtils.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Dispatch
+import NIOConcurrencyHelpers
+import NIOCore
+import NIOEmbedded
+
+/// An `EventLoopGroup` of `EmbeddedEventLoop`s.
+final class EmbeddedEventLoopGroup: EventLoopGroup {
```
glbrntt, line 21: What a great idea!

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -0,0 +1,115 @@
...
+            self.eventLoopQueues = [:]
+        }
+
+        var count: Int {
+            self.generalPurposeQueue.count
+        }
+
+        var isEmpty: Bool {
+            self.generalPurposeQueue.isEmpty
+        }
```
glbrntt, line 34: This is surprising behaviour I think -- I would expect the `count` to return the total count and `isEmpty` to only return true if both queues are empty.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -123,6 +123,46 @@ enum HTTPConnectionPool {
...
+            switch self.req.eventLoopPreference.preference {
+            case .indifferent, .delegate:
+                return nil
+            case .delegateAndChannel(on: let eventLoop), .testOnly_exact(channelOn: let eventLoop, delegateOn: _):
+                return eventLoop
+            }
+        }
+
+        func __testOnly_internal_value() -> HTTPSchedulableRequest {
```
glbrntt, line 160: nit: "wrapped request" or similar would be clearer than `internal_value`

glbrntt (changes requested):
I think the queue has some surprising behaviour that should be changed or at least made much clearer.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
fabianfett: I hope to solve all those places, with this single place. It is really only important when queuing the request. So I hope this is the place!

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -0,0 +1,131 @@
...
+        private var generalPurposeQueue: CircularBuffer<Request>
+        private var eventLoopQueues: [EventLoopID: CircularBuffer<Request>]
+
+        init() {
+            self.generalPurposeQueue = CircularBuffer(initialCapacity: 32)
+            self.eventLoopQueues = [:]
+            self.count = 0
+        }
+
+        private(set) var count: Int
```
glbrntt, line 29: How often is this used externally? If it's not used frequently then dynamically computing this would be better I think...

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -0,0 +1,131 @@
...
+            }
+            return closure(&self.eventLoopQueues[eventLoopID]!)
+        }
+
+        private func withEventLoopQueueIfAvailable<Result>(
+            for eventLoopID: EventLoopID,
+            _ closure: (CircularBuffer<Request>) -> Result
+        ) -> Result? {
+            if self.eventLoopQueues[eventLoopID] != nil {
+                return closure(self.eventLoopQueues[eventLoopID]!)
```
glbrntt, line 126: nit: avoid hashing twice here (also we can avoid the bang)

glbrntt (commented):
Looks good modulo a couple of nits

glbrntt (approved):
LGTM

PR #413, opened 2021-09-02
HTTP1Connections state machine for ConnectionPool

fabianfett:
If we have HTTP/1 connections, a data structure to store them is handy.

### Changes

- Adds a `HTTPConnectionPool.HTTP1Connections` struct to store HTTP1 connections

### Result

Less code in the connection pool state machine. Shared code between HTTP/1 and HTTP/2 connection pool state machine.


Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+                preconditionFailure("Unexpected state: Did not expect to have connections with this state in the state machine: \(self.state)")
+            }
+        }
+    }
+
+    /// A structure to hold the currently active HTTP/1.1 connections.
+    ///
+    /// The general purpose connection pool (pool for requests that don't have special `EventLoop`
+    /// requirements) will grow up until `maximumConcurrentConnections`. If requests have
+    /// special `EventLoop` requirements overflow connections might be opened.
```
glbrntt, line 169: Just checking I understand: the size of the overflow isn't limited at all, we assume that the size of the overflow will be small.

Do we treat overflow connections differently? Once they are returned do we close them immediately or do we keep them around for some time?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+            case park(Connection.ID)
+            /// Close the connection and remove
+            case close(Connection)
+        }
+
+        // MARK: Connection creation
+
+        mutating func createNewConnection(on eventLoop: EventLoop) -> Connection.ID {
+            let connection = HTTP1ConnectionState(connectionID: self.generator.next(), eventLoop: eventLoop)
+            self.connections.append(connection)
```
glbrntt, line 255: What if there are overflow connections; don't we need to insert this before the overflow index?

glbrntt: Should we also be asserting that we `canGrow`? (I assume it's the callers responsibility to check this.)

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+        var availableSince: NIODeadline? {
+            switch self.state {
+            case .available(_, since: let lastReturn):
+                return lastReturn
+            case .backingOff, .starting, .leased, .closed:
+                return nil
+            }
+        }
+
+        mutating func started(_ connection: Connection) {
```
glbrntt, line 89: "started" makes me think that an attempt to establish a connection has started rather than the connection has been established and is ready to use. "connected" may be slightly clearer here.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+            return connection.connectionID
+        }
+
+        mutating func createNewOverflowConnection(on eventLoop: EventLoop) -> Connection.ID {
+            let connection = HTTP1ConnectionState(connectionID: self.generator.next(), eventLoop: eventLoop)
+            self.connections.append(connection)
+            return connection.connectionID
+        }
+
+        mutating func newHTTP1ConnectionCreated<RequestType>(
```
glbrntt, line 266: I _think_ this is called when the connection has been established, "created" doesn't give me that impression.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+        ) -> NextConnectionAction<RequestType> {
+            guard let index = self.connections.firstIndex(where: { $0.connectionID == connection.id }) else {
+                preconditionFailure("There is a new connection that we didn't request!")
+            }
+            assert(connection.eventLoop === self.connections[index].eventLoop, "Expected the new connection to be on EL")
+            self.connections[index].started(connection)
+            return self.withIdleConnection(index: index, closure: closure)
+        }
+
+        mutating func backoffNextConnectionAttempt(_ connectionID: Connection.ID) -> EventLoop {
```
glbrntt, line 278: Presumably the `EventLoop` is the loop to schedule the next attempt on?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+            }
+
+            self.connections[index].failedToStart()
+            return self.connections[index].eventLoop
+        }
+
+        // MARK: Leasing and releasing
+
+        mutating func leaseConnection(onPreferred preferredEL: EventLoop) -> Connection? {
+            guard let index = self.findAvailableConnection(onPreferred: preferredEL) else {
```
glbrntt, line 290:
```suggestion
        mutating func leaseConnection(onPreferred eventLoop: EventLoop) -> Connection? {
            guard let index = self.findAvailableConnection(onPreferred: eventLoop) else {
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+                return nil
+            }
+
+            return self.connections[index].lease()
+        }
+
+        mutating func leaseConnection(onRequired preferredEL: EventLoop) -> Connection? {
+            guard let index = self.findAvailableConnection(onRequired: preferredEL) else {
```
glbrntt, line 298:
```suggestion
        mutating func leaseConnection(onRequired eventLoop: EventLoop) -> Connection? {
            guard let index = self.findAvailableConnection(onRequired: eventLoop) else {
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+
+        struct IdleConnectionContext {
+            /// The `EventLoop` the connection runs on.
+            var eventLoop: EventLoop
+            /// The connection's use. Either general purpose or for requests with `EventLoop`
+            /// requirements.
+            var use: ConnectionUse
+        }
+
+        enum IdleConnectionAction<R> {
```
glbrntt, line 325: What's `R`, `Request`?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+            case park
+            /// Close the connection and remove
+            case close
+        }
+
+        private mutating func withIdleConnection<RequestType>(
+            index: Int,
+            closure: (IdleConnectionContext) -> (IdleConnectionAction<RequestType>)
+        ) -> NextConnectionAction<RequestType> {
+            let eventLoop = self.connections[index].eventLoop
```
glbrntt, line 338: Should we assert that we are idle?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+            }
+        }
+
+        // MARK: Connection close/removal
+
+        mutating func closeConnectionIfIdle(_ connectionID: Connection.ID) -> Connection? {
+            guard let index = self.connections.firstIndex(where: { $0.connectionID == connectionID }) else {
+                // because of a race this connection (connection close runs against trigger of timeout)
+                // was already removed from the state machine.
+                return .none
```
glbrntt, line 369: nit: `nil` (let's be consistent within a single func 🙂)

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+extension HTTPConnectionPool {
+    /// Represents the state of a single HTTP/1.1 connection
+    private struct HTTP1ConnectionState {
+        enum State {
+            /// the connection is creating a connection. Valid transitions are to: .backingOff, .available and .failed
+            case starting
+            /// the connection is waiting to retry the establishing a connection. Transition to .closed. Must be replaced for a retry.
+            case backingOff
+            /// the connection is available for a new request. Valid transitions to: .leased and .closed
+            case available(Connection, since: NIODeadline)
```
glbrntt, line 27: "idle" and "available" are both used through this PR to refer to the same thing, can we stick to one?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+        enum FailedConnectionAction {
+            /// Creates a new connection, that needs to be started
+            case replace
+            /// Removes the old connection.
+            case remove
+        }
+
+        /// Remove or replace a connection.
+        ///
+        /// Call this function after the connection start backoff is done, or a live connection is closed.
```
glbrntt, line 407: So do we expect the state to be one of `available`, `leased` or `backingOff`?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+                preconditionFailure("We tried to create a new connection that we know nothing about?")
+            }
+
+            let use: ConnectionUse
+            let eventLoop = self.connections[index].eventLoop
+            let starting: Int
+            if index < self.overflowIndex {
+                use = .generalPurpose
+                // we need to subtract 1, since we dont want this connection to count
+                starting = self.startingGeneralPurposeConnections - 1
```
glbrntt, line 428: Is this right? If we're a live connection that's been closed then we won't be in the starting state.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,602 @@
...
+            }
+        }
+
+        // MARK: Shutdown
+
+        mutating func shutdown() -> CleanupContext {
+            var cleanupContext = CleanupContext()
+            let initialOverflowIndex = self.overflowIndex
+
+            self.connections = self.connections.enumerated().compactMap { index, connectionState in
```
glbrntt, line 465: What's special about the connections that we keep track of?

glbrntt (commented):
I haven't looked at the tests -- mostly looks good, left a handful of questions and minor suggestions.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
fabianfett: > the size of the overflow isn't limited at all, we assume that the size of the overflow will be small.
yes

> Once they are returned do we close them immediately or do we keep them around for some time?
I want to close overflow connections right away, once there isn't more work for them todo.

However with this basic setup changing this to whatever should be fairly simple :)

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
fabianfett: Wow. Save of the week! This would have been really hard to debug!

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,618 @@
...
+            case .closed:
+                preconditionFailure("Invalid state: \(self.state)")
+            }
+        }
+
+        /// Cleanup the current connection for shutdown.
+        ///
+        /// This method is called, when the connections shall shutdown. Depending on the state
+        /// the connection is in, it adds itself to one of the arrays that are used to signal shutdown
+        /// intend to the underlying connections. Connections that are backing off can be easily
```
glbrntt, line 159:
```suggestion
        /// intent to the underlying connections. Connections that are backing off can be easily
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
glbrntt: It might be useful to document this to explain that while the pool can overflow we'll try to pull it back down to size when we can.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,618 @@
...
+            }
+
+            self.connections[index].release()
+            let context = self.generateIdleConnectionContextForConnection(at: index)
+            return (index, context)
+        }
+
+        // MARK: Connection close/removal
+
+        mutating func closeConnection(at index: Int) -> Connection {
```
glbrntt, line 394: might be worth mentioning that this also removes the connection

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
fabianfett: documented in the cleanup call of a single state

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
fabianfett: > It might be useful to document this to explain that while the pool can overflow we'll try to pull it back down to size when we can.

This is finally the decision of the state machine. I don't want to document here, since this might get out of sync very fast.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
glbrntt: Much clearer with the cleanup action too!

glbrntt (approved):
Looks great!

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -0,0 +1,628 @@
...
+                }
+            }
+            return connecting
+        }
+
+        func startingEventLoopConnections(on eventLoop: EventLoop) -> Int {
+            return self.connections[self.overflowIndex..<self.connections.endIndex].reduce(into: 0) { count, connection in
+                guard connection.eventLoop === eventLoop else { return }
+                if connection.isConnecting || connection.isBackingOff {
+                    count += 1
```
Lukasa, line 264: This can also be unchecked for the same reasons as above.

Lukasa (approved):
One note, otherwise LGTM.

PR #416, opened 2021-09-06
[ConnectionPool] HTTP1StateMachine

fabianfett:
Follow up to #415, #417.

For the new `HTTPConnectionPool` a new `StateMachine` is handy.


Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+StateMachine.swift
```diff
@@ -0,0 +1,287 @@
...
+            case scheduleRequestTimeout(NIODeadline?, for: Request.ID, on: EventLoop)
+            case cancelRequestTimeout(Request.ID)
+
+            case none
+        }
+
+        enum HTTPTypeStateMachine {
+            case http1(HTTP1StateMachine)
+
+            case _modifying
```
Lukasa, line 66: We almost certainly shouldn't need this case. I know I haven't provided this feedback much prior to here, so don't feel compelled to remove it now, but we should probably go through these state machines and rip out the `_modifying` states wherever we have then. Recent Swift versions do not CoW here.

Lukasa (changes requested):

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -0,0 +1,432 @@
...
+            //    cancellation.
+            return .none
+        }
+
+        mutating func shutdown() -> Action {
+            precondition(self.state == .running, "Shutdown must only be called once")
+
+            // If we have remaining request queued, we should fail all of them with a cancelled
+            // error.
+            let waitingRequests = self.queue.removeAll().map { ($0, $0.id) }
```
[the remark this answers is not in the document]
fabianfett: I create the tuple here to keep consistency between the two request actions:
```
case failRequest(Request, Error, cancelTimeout: Request.ID?)
case failRequests([(Request, cancelTimeout: Request.ID?)], Error)
```

If we fail multiple requests at the same time, we will always need to cancel their timeouts.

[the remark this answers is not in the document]
fabianfett: Chatted offline: Fixed now

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -375,6 +375,11 @@ extension HTTPConnectionPool {
             self.connections[index].lease()
         }
 
+        mutating func parkConnection(at index: Int) -> (Connection.ID, EventLoop) {
```
glbrntt, line 378: This isn't actually a `mutating` `func`

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -0,0 +1,459 @@
...
+                return .init(
+                    request: .none,
+                    connection: .createConnection(newConnectionID, on: newEventLoop)
+                )
+            }
+            self.connections.removeConnection(at: index)
+            return .none
+        }
+
+        private mutating func calculateBackoff(for attempt: Int) -> TimeAmount {
```
glbrntt, line 426: Don't think this is actually mutating?

glbrntt (commented):
(Partial review, will finish tomorrow.)

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -0,0 +1,457 @@
...
+                on: eventLoop
+            )
+
+            if !self.connections.canGrow {
+                // all connections are busy and there is no room for more connections, we need to wait!
+                return .init(request: requestAction, connection: .none)
+            }
+
+            // if we are not at max connections, we may want to create a new connection
+            if self.connections.startingGeneralPurposeConnections >= self.queue.count {
```
Lukasa, line 87: Is this counting the wrong thing? This seems to risk starting a general purpose connection because we have lots of loop specific work, but without consulting what that loop specific work is.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -0,0 +1,457 @@
...
+            case .shutDown:
+                preconditionFailure("If the pool is already shutdown, all connections must have been torn down.")
+            }
+        }
+
+        private mutating func nextActionForFailedGeneralPurposeConnection(
+            at index: Int,
+            context: HTTP1Connections.FailedConnectionContext
+        ) -> Action {
+            if context.connectionsStartingForUseCase < self.queue.count(for: nil) {
```
Lukasa, line 394: Perhaps file an issue for future enhancement (no sense gluing it into this PR), but I don't love this spelling. Probably `self.queue.generalPurposeCount` would be clearer.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+StateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+            case cancelRequestTimeout(Request.ID)
+
+            case none
+        }
+
+        enum HTTPTypeStateMachine {
+            case http1(HTTP1StateMachine)
+        }
+
+        var state: HTTPTypeStateMachine
```
glbrntt, line 67: Assuming `HTTPTypeStateMachine` will only have `http1` and `http2` cases (at least for now!) then `httpVersion`/`httpType` or similar might be a clearer name here since we're really just delegating to the state machine each case holds.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+StateMachine.swift
```diff
@@ -0,0 +1,222 @@
...
+            case .http1(var http1StateMachine):
+                let action = http1StateMachine.http1ConnectionReleased(connectionID)
+                self.state = .http1(http1StateMachine)
+                return action
+            }
+        }
+
+        mutating func shutdown() -> Action {
+            guard !self.isShuttingDown else {
+                preconditionFailure("Shutdown must only be called once")
```
glbrntt, line 180: nit: `precondition(!self.isShuttingDown, "...")`

Tests/AsyncHTTPClientTests/HTTPConnectionPool+HTTP1StateTests.swift
```diff
@@ -0,0 +1,578 @@
...
+        }
+
+        // Add 100 requests to fill request queue
+        var queuedRequestsOrder = CircularBuffer<MockRequestQueuer.RequestID>()
+        var queuer = MockRequestQueuer()
+        for _ in 0..<100 {
+            let eventLoop = elg.next()
+
+            // in 10% of the cases, we require an explicit EventLoop.
+//            let elRequired = (0..<10).randomElement().flatMap { $0 == 0 ? true : false }!
```
glbrntt, line 295: What's the deal here? (also in `testBestConnectionIsPicked`)

glbrntt (approved):
LGTM!

Lukasa (approved):
:shipit:

PR #421, opened 2021-09-10
[ConnectionPool] Preserve connection errors

fabianfett:
### Motivation

If we can't establish a connection to a remote for the user, we will retry to create the connection. Eventually the requests connection timeout will fire, and we will fail the request. Right now we fail the request always with the `HTTPClientError.getConnectionFromPoolTimeout` error. While this is very descriptive, it does not tell the user if there are to many waiting requests or if there was a connection problem.

### Changes

- If errors occur during connection creation, we will save the error in the state machine
- If a request times out, after we have seen a connection creation error, we will fail the request with the connection creation error.

### Result

If user have a typo in their url, they will get better errors. (dns failure)

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
@@ -26,6 +26,8 @@ extension HTTPConnectionPool {
 
         private var connections: HTTP1Connections
         private var failedConsecutiveConnectionAttempts: Int = 0
+        /// the error from the last connection creation
+        private var lastConnectFailure: Error?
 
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -26,6 +26,8 @@ extension HTTPConnectionPool {
 
         private var connections: HTTP1Connections
         private var failedConsecutiveConnectionAttempts: Int = 0
+        /// the error from the last connection creation
+        private var lastConnectFailure: Error?
```
Lukasa, line 30: Possible future enhancement: store a vector of errors and throw a single, composite error that encompasses them all.

Lukasa (approved):
Nice, LGTM.

PR #424, opened 2021-09-13
Improve HTTPClient tests

fabianfett:
### Motivation

We have some tests that make it hard to see the reason for a test failure. Since `HTTPClientError` is equatable, let's use `XCTAssertEqual`

### Modifications

- Use `XCTAssertEqual` to compare `HTTPClientError`s
- Don't call some server on the internet expecting it will not respond. Create a server and shut it down directly, use the now unused port.

### Result

- Better diagnostics on failing tests
- Less outside dependencies on tests

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -556,15 +556,13 @@ class HTTPClientTests: XCTestCase {
         XCTAssertEqual(0, progress.receivedBytes)
     }
 
-    func testRemoteClose() throws {
-        XCTAssertThrowsError(try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "close").wait(), "Should fail") { error in
-            guard case let error = error as? HTTPClientError, error == .remoteConnectionClosed else {
-                return XCTFail("Should fail with remoteConnectionClosed")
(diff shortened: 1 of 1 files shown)


Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -556,48 +556,46 @@ class HTTPClientTests: XCTestCase {
...
     }
 
-    func testDeadline() throws {
-        XCTAssertThrowsError(try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "wait", deadline: .now() + .milliseconds(150)).wait(), "Should fail") { error in
-            guard case let error = error as? HTTPClientError, error == .readTimeout else {
-                return XCTFail("Should fail with readTimeout")
-            }
+    func testDeadline() {
+        XCTAssertThrowsError(try self.defaultClient.get(url: self.defaultHTTPBinURLPrefix + "wait", deadline: .now() + .milliseconds(150)).wait()) {
+            XCTAssertEqual($0 as? HTTPClientError, .readTimeout)
```
fabianfett, line 598: Getting a `readTimeout` error on a deadline is weird. This will be actually fixed with the new implementation

Lukasa (approved):

PR #425, opened 2021-09-13
HTTP1Connection: Close on error while in request

fabianfett:
### Motivation

In the `HTTP1ConnectionStateMachine` we set the connection state to `.closed` after having forwarded an error to the inner RequestStateMachine. Since we set the error to close, the `state.modify` function did not create an action that would close the connection.

### Changes

- Modify the ConnectionState based on the RequestStateMachine only in the `state.modify` function.
- Add a test to verify the fixed behavior.
- Use ExpressibleByDictionaryLiteral syntax for some HTTPHeaders to increase readability.

### Result

Connections are closed, if an error happens, while we are in a request.


Lukasa (approved):
Nice catch

PR #426, opened 2021-09-14
Fix potentially flaky `HTTPClientTests.testNoBytesSentOverBodyLimit`

fabianfett:
### Motivation

If we call 
```
context.write(ByteBuffer(string: "foo"), promise: nil)
context.flush()
context.close()
```
there is no guarantee that `"foo"` will be written to the network. Currently the test `testNoBytesSentOverBodyLimit` assumes this however.

### Changes

- `testNoBytesSentOverBodyLimit` does not assume anymore, that bytes will be sent before a close of the connection.
- Use `XCTAssertEqual` to compare `Equatable` errors.

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -968,11 +968,8 @@ class HTTPClientTests: XCTestCase {
         request.body = .byteBuffer(ByteBuffer(bytes: [120, 156, 75, 76, 28, 5, 200, 0, 0, 248, 66, 103, 17]))
         request.headers.add(name: "Accept-Encoding", value: "deflate")
 
-        XCTAssertThrowsError(try localClient.execute(request: request).wait()) { error in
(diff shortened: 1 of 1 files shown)


Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -2644,19 +2641,45 @@ class HTTPClientTests: XCTestCase {
...
+            _ = try server.receiveHead()
+
+            // A request head was sent. We expect the request now to fail with a parsing error,
+            // since the client ended the connection to early (from the server's point of view.)
+            XCTAssertThrowsError(try server.readInbound()) {
+                XCTAssertEqual($0 as? HTTPParserError, HTTPParserError.invalidEOFState)
+            }
+        } catch {
+            // TBD: We sadly can't verify the error type, since it is private in `NIOTestUtils`:
+            //      NIOTestUtils.BlockingQueue<Element>.TimeoutError
```
Lukasa, line 2676: We should probably find a way to make this public so we can be confident that we're throwing the right error.

Lukasa (approved):

PR #427, opened 2021-09-14
[HTTPConnectionPool] Implementation switch

fabianfett:
This is the final pr to switch the implementation over to our new HTTPConnectionPool and Connections. There are some outstanding changes (#425, #426) that we want to land in smaller prs beforehand, but for everyone interested, this is your chance to play with the new implementation.

The failing unit tests are tests related logging, we will need to fix this before we can land this pr.


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -694,69 +704,24 @@ extension HTTPClient {
...
-                }.flatMapError { error in
-                    fatalError("Couldn't remove taskHandler: \(error)")
-                }
-            } else {
-                // TODO: This seems only reached in some internal unit test
-                // Maybe there could be a better handling in the future to make
-                // it an error outside of testing contexts
-                return self.eventLoop.makeSucceededFuture(())
-            }
+            self.promise.fail(error)
```
fabianfett, line 724: Those functions are not hit from the new implementation. However they are hit from the old implementation. Since we don't want to remove the old implementation in this pr, they should stay for now. We will be able to remove them in a follow up pr.

Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
```diff
@@ -290,7 +290,7 @@ class HTTPClientInternalTests: XCTestCase {
         let httpBin = HTTPBin()
         let httpClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup))
         defer {
-            XCTAssertNoThrow(try httpClient.syncShutdown(requiresCleanClose: true))
+            XCTAssertNoThrow(try httpClient.syncShutdown())
```
fabianfett, line 293: In the new implementation we can not make guarantees about the order of succeeding/failing the request (which will lead to a call to `syncShutdown`) and marking the connection as idle. For this reason we can not enforce a clean shutdown here.

Tests/AsyncHTTPClientTests/HTTPClientInternalTests+XCTest.swift
```diff
@@ -34,17 +34,11 @@ extension HTTPClientInternalTests {
             ("testRequestFinishesAfterRedirectIfServerRespondsBeforeClientFinishes", testRequestFinishesAfterRedirectIfServerRespondsBeforeClientFinishes),
             ("testProxyStreaming", testProxyStreaming),
             ("testProxyStreamingFailure", testProxyStreamingFailure),
-            ("testUploadStreamingBackpressure", testUploadStreamingBackpressure),
```
weissi, line 37: do we have replacements for these tests?

fabianfett: Yes. This is now tested way lower in the stack on the connection directly.

Lukasa (approved):
Nice, great work @fabianfett, this looks really good! :sparkles:

fabianfett:
Apparently we have a new flaky test (see swift-nightly). Will fix in a follow up PR. Merge this one.

PR #428, opened 2021-09-15
Add logging to the new HTTPConnectionPool

fabianfett:
### Motivation

To debug the new `HTTPConnectionPool` and `Connections`, it would be great to get a rather complete event story.

### Changes

- Added a number of trace logs
- Added a nicer debug output for `Connection.Key`

### Result

A better debuggable AHC client.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -124,7 +124,7 @@ final class ConnectionPool {
     /// A key is initialized from a `URL`, it uses the components to derive a hashed value
     /// used by the `providers` dictionary to allow retrieving and creating
     /// connection providers associated to a certain request in constant time.
-    struct Key: Hashable {
+    struct Key: Hashable, CustomStringConvertible {
         init(_ request: HTTPClient.Request) {
             switch request.scheme {
             case "http":
@@ -178,6 +178,19 @@ final class ConnectionPool {
                 config.tlsConfiguration = tlsConfiguration.base
             }
             return config
+        }
+
+        var description: String {
+            var hasher = Hasher()
+            self.tlsConfiguration?.hash(into: &hasher)
+            let hash = hasher.finalize()
+            var path = ""
+            if self.unixPath != "" {
+                path = self.unixPath
+            } else {
+                path = "\(self.host):\(self.port)"
+            }
+            return "\(self.scheme)://\(path) TLS-hash: \(hash)"
         }
     }
 }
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -130,14 +130,7 @@ final class HTTPConnectionPool {
     }
 
     private let stateLock = Lock()
-    private var _state: StateMachine {
-        didSet {
-            self.logger.trace("Connection Pool State changed", metadata: [
-                "key": "\(self.key)",
-                "state": "\(self._state)",
-            ])
-        }
```
fabianfett, line 139: Removed, since it produces way too many logs. We should solely focus on recording the events so that we can replay the finite state machine if needed.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1.1/HTTP1ClientChannelHandler.swift
```diff
@@ -142,12 +143,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
         let req = self.unwrapOutboundIn(data)
         self.request = req
 
-        self.logger.trace("New request to execute")
-
-        if let idleReadTimeout = self.request?.idleReadTimeout {
-            self.idleReadTimeoutStateMachine = .init(timeAmount: idleReadTimeout)
-        }
```
fabianfett, line 149: This could be removed, since it is already included in the request setter (`l. 143`). We have tests that ensure that idle read timeouts work as expected. Those continue to succeed.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -179,6 +179,19 @@ final class ConnectionPool {
             }
             return config
         }
+
+        var description: String {
+            var hasher = Hasher()
```
Lukasa, line 184: Are you sure this is what you want? This hash is not stable across executions.

Lukasa (approved):

Sources/AsyncHTTPClient/ConnectionPool.swift
fabianfett: I think I am... It is for aggregation within one run. I don't think I want to look at connection pool state changes between runs.

Lukasa: Great, then this all looks good.

PR #429, opened 2021-09-16
[DRAFT] HTTP2 connection pooling

dnadoba:


Lukasa:

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        var isAvailable: Bool {
+            switch self.state {
+            case .active(_, let maxStreams, let usedStreams, _):
+                return usedStreams < maxStreams
+            case .starting, .backingOff, .draining, .closed:
+                return false
+            }
+        }
+
+        /// The connection is active but there are no requests running on the connection
```
fabianfett, line 63:
```suggestion
        /// The connection is active, but there are no running requests on the connection.
        /// Every idle connection is available, but not every available connection is idle.
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        var usedAndMaxStreams: (Int, Int)? {
+            switch self.state {
+            case .starting, .backingOff, .draining, .closed:
+                return nil
+            case .active(_, let maxStreams, let usedStreams, _):
+                return (usedStreams, maxStreams)
+            }
+        }
```
fabianfett, line 89: I don't see this being used anywhere. So I guess we should be able to drop it.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        
+        var connection: Connection? {
+            switch state {
+            case .starting, .backingOff, .closed:
+                return nil
+            case .active(let connection, _, _, _),
+                    .draining(let connection, _, _):
+                return connection
+            }
+        }
```
fabianfett, line 99: I don't think, this is something we want. We only want to get access to the underlying connection, if we want to execute an action that requires this. I don't see any usage of the property, either.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        let connectionID: Connection.ID
+
+        /// should be called after the connection was successfully established
+        /// - Parameters:
+        ///   - conn: HTTP2 connection
+        ///   - maxStreams: max streams settings from the server
+        /// - Returns: number of available streams which can be leased
+        mutating func connected(_ conn: Connection, maxStreams: Int) -> Int {
+            guard case .starting = self.state else {
+                preconditionFailure("Invalid state")
```
fabianfett, line 112:
```suggestion
                preconditionFailure("Invalid state: \(self.state)")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+            return maxStreams
+        }
+        
+        /// should be called after receiving new http2 settings from the server
+        /// - Parameters:
+        ///   - maxStreams: max streams settings from the server
+        /// - Returns: number of available streams which can be leased
+        mutating func newMaxConcurrentStreams(_ maxStreams: Int) -> Int {
+            guard case let .active(conn, _, usedStreams, lastIdle) = self.state else {
+                preconditionFailure("Invalid state")
```
fabianfett, line 124:
```suggestion
                preconditionFailure("Invalid state: \(self.state)")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+            }
+            self.state = .active(conn, maxStreams: maxStreams, usedStreams: usedStreams, lastIdle: lastIdle)
+            return max(maxStreams - usedStreams, 0)
+        }
+        
+        mutating func goAwayRecieved() {
+            guard case let .active(conn, maxStreams, usedStreams, _) = self.state else {
+                preconditionFailure("Invalid state")
```
fabianfett, line 132:
```suggestion
                preconditionFailure("Invalid state: \(self.state)")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+            case .starting, .active, .backingOff, .draining:
+                self.state = .closed
+            case .closed:
+                preconditionFailure("Invalid state: \(self.state)")
+            }
+        }
+
+        mutating func lease(_ count: Int) -> Connection {
+            guard case .active(let conn, let maxStreams, var usedStreams, let lastIdle) = self.state else {
+                preconditionFailure("Invalid state")
```
fabianfett, line 158:
```suggestion
                preconditionFailure("Invalid state: \(self.state)")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        /// should be called after receiving new http2 settings from the server
+        /// - Parameters:
+        ///   - maxStreams: max streams settings from the server
+        /// - Returns: number of available streams which can be leased
+        mutating func newMaxConcurrentStreams(_ maxStreams: Int) -> Int {
+            guard case let .active(conn, _, usedStreams, lastIdle) = self.state else {
+                preconditionFailure("Invalid state")
+            }
+            self.state = .active(conn, maxStreams: maxStreams, usedStreams: usedStreams, lastIdle: lastIdle)
+            return max(maxStreams - usedStreams, 0)
```
fabianfett, line 127: I think, in the draining case, we are probably in a position where we could receive further setting updates.

```swift
            switch self.state {
            case let .active(conn, _, usedStreams, lastIdle):
                self.state = .active(conn, maxStreams: maxStreams, usedStreams: usedStreams, lastIdle: lastIdle)
                return max(maxStreams - usedStreams, 0)
            case others:
                preconditionFailure("Invalid state: \(self.state)")
            }
```

however if we are draining the available streams should always be 0.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+            guard case .active(let conn, let maxStreams, var usedStreams, var lastIdle) = self.state else {
+                preconditionFailure("Invalid state")
+            }
+            usedStreams -= 1
+            assert(usedStreams >= 0)
+            if usedStreams == 0 {
+                lastIdle = .now()
+            }
+            self.state = .active(conn, maxStreams: maxStreams, usedStreams: usedStreams, lastIdle: lastIdle)
+            return max(maxStreams - usedStreams, 0)
```
fabianfett, line 178: use switch for the state. If we are draining, we might also release connections.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        mutating func close() -> Connection {
+            guard case .active(let conn, _, 0, _) = self.state else {
+                preconditionFailure("Invalid state")
+            }
+            self.state = .closed
+            return conn
```
fabianfett, line 186: switch over state.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+                    return .removeConnection
+                } else {
+                    context.cancel.append(connection)
+                    return .keepConnection
+                }
+            case .draining(let connection, _, _):
+                context.cancel.append(connection)
+                return .keepConnection
+            case .closed:
+                preconditionFailure("Unexpected state for cleanup: Did not expect to have connections with this state in the state machine: \(self.state)")
```
fabianfett, line 227:
```suggestion
                preconditionFailure("Unexpected state for cleanup: Did not expect to have closed connections in the state machine.")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+            self.state = .starting
+        }
+    }
+    
+    struct HTTP2Connections {
+        /// A connection's use. Did it serve in the pool or was it specialized for an `EventLoop`?
+        enum ConnectionUse {
+            case generalPurpose
+            case eventLoop(EventLoop)
+        }
```
fabianfett, line 243: I don't think we need this for http2 based on our plan.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        
+        /// Information around an idle connection.
+        struct IdleConnectionContext {
+            /// The `EventLoop` the connection runs on.
+            var eventLoop: EventLoop
+            /// The connection's use. Either general purpose or for requests with `EventLoop`
+            /// requirements.
+            var use: ConnectionUse
+        }
```
fabianfett, line 252: We don't need this buddy, I think. we have the better fitting `AvailableConnectionContext`.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+            self.connections = []
+            self.connections.reserveCapacity(8) // just get's us some early breathing room
+        }
+        
+        // MARK: Connection creation
+        
+        struct AvailableConnectionContext {
+            var availableStreams: Int
+            var eventLoop: EventLoop
+        }
```
fabianfett, line 276: I think, we need another property is `idle`. Based on this we can schedule idle timeouts.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        
+        mutating func createNewConnection(on eventLoop: EventLoop) -> Connection.ID {
+            // assert no connection exists on the requested eventLoop
+            assert(self.connections.first(where: { $0.eventLoop === eventLoop }) == nil)
```
fabianfett, line 280:
```suggestion
            assert(self.connections.allSatisfy(where: { $0.eventLoop !== eventLoop }))
```

There might be another catch here. If a connection is draining, we might want to create a new one on the same EL.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        }
+        
+        /// A new HTTP/2.0 connection was established.
+        ///
+        /// This will put the connection into the idle state.
+        ///
+        /// - Parameter connection: The new established connection.
+        /// - Returns: An index and an IdleConnectionContext to determine the next action for the now idle connection.
+        ///            Call ``leaseConnection(at:)`` or ``closeConnection(at:)`` with the supplied index after
+        ///            this.
```
fabianfett, line 294:
```suggestion
        /// - Returns: An index and an AvailableConnectionContext to determine the next action for the now idle connection.
        ///            Call ``leaseStreams(at index: Int, count: Int)`` or ``closeConnection(at:)`` with the supplied index after
        ///            this.
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+                eventLoop: self.connections[index].eventLoop
+            )
+            return (index, context)
+        }
+        
+        // MARK: Leasing and releasing
+        
+        mutating func leaseStream(onPreferred eventLoop: EventLoop) -> Connection? {
+            guard let index = self.findAvailableConnection(onPreferred: eventLoop) else { return nil }
+            return leaseStreams(at: index, count: 1)
```
fabianfett, line 346:
```suggestion
            return self.leaseStreams(at: index, count: 1)
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        }
+        
+        
+        /// tries to find an available connection on the prefered `eventLoop`. If it can't find one with the given `eventLoop`, it returns the first available connection
+        private func findAvailableConnection(onPreferred eventLoop: EventLoop) -> Int? {
+            var availableConnection: Int?
+            for (index, connection) in zip(connections.indices, connections) {
```
fabianfett, line 353:
```suggestion
            for (index, connection) in connections.enumerated() {
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+                } else if availableConnection == nil {
+                    availableConnection = index
+                }
+            }
+            return availableConnection
+        }
+        
+        mutating func leaseStreams(onRequired eventLoop: EventLoop, count: Int) -> Connection? {
+            guard let index = self.findAvailableConnection(onRequired: eventLoop) else { return nil }
+            return leaseStreams(at: index, count: count)
```
fabianfett, line 366:
```suggestion
            return self.leaseStreams(at: index, count: count)
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+        }
+        
+        /// tries to find an available connection on the required `eventLoop`
+        private func findAvailableConnection(onRequired eventLoop: EventLoop) -> Int? {
+            self.connections.firstIndex(where: { $0.eventLoop === eventLoop })
```
fabianfett, line 371: We must check here, that the connection is actually available.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,446 @@
...
+            )
+
+            self.connections[index] = newConnection
+            return (newConnection.connectionID, newConnection.eventLoop)
+        }
+        
+        mutating func failConnection(_ connectionID: Connection.ID) -> (Int, FailedConnectionContext)? {
+            guard let index = self.connections.firstIndex(where: { $0.connectionID == connectionID }) else {
+                return nil
+            }
```
fabianfett, line 426: I think it would be useful, if we had a comment here, explaining why we might get this callback, and we know nothing about a connection with this id.

fabianfett (commented):
A very great first draft! 🚀

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,116 @@
...
+        typealias Action = HTTPConnectionPool.StateMachine.Action
+        
+        private(set) var estimatedStreamsPerConnection = 100
+
+        private(set) var connections: HTTP2Connections
+        
+        private var requests: RequestQueue
+        init(
+            migrationContext: HTTP2MigrationContext,
+            eventLoopGroup: EventLoopGroup
```
fabianfett, line 29: I think we should stop passing the EL group.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,116 @@
...
+        ) {
+            let eventLoopCount = eventLoopGroup.makeIterator().reduce(into: 0, { counter, _ in counter += 1 })
```
fabianfett, line 31: I think this is an optimization that is not needed.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -291,6 +291,15 @@ extension HTTPConnectionPool {
                 connection: .cleanupConnections(cleanupContext, isShutdown: isShutdown)
             )
         }
+        
+        mutating func migrateToHTTP2() -> HTTP2MigrationContext {
+            let connectionsMigrationContext = self.connections.migrateToHTTP2()
+            return HTTP2MigrationContext(
+                generator: connectionsMigrationContext.generator,
```
fabianfett, line 298: I think the generator should not be in the migration context

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -291,6 +291,15 @@ extension HTTPConnectionPool {
...
+                startingConnections: connectionsMigrationContext.starting,
```
fabianfett, line 299: We need to hand over the `HTTP1Connections` as a whole. Then we can call the migrate methods on `HTTP1Connections` from the `HTTP2StateMachine`.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,116 @@
...
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOHTTP2
+
+extension HTTPConnectionPool {
+    struct HTTP2StateMaschine {
+        typealias Action = HTTPConnectionPool.StateMachine.Action
+        
+        private(set) var estimatedStreamsPerConnection = 100
```
fabianfett, line 22: probably a `static let` is better here.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+StateMachine.swift
```diff
@@ -81,41 +98,40 @@ extension HTTPConnectionPool {
...
         mutating func executeRequest(_ request: Request) -> Action {
-            switch self.state {
-            case .http1(var http1StateMachine):
-                let action = http1StateMachine.executeRequest(request)
-                self.state = .http1(http1StateMachine)
-                return action
+            self.state.modify { http1 in
+                http1.executeRequest(request)
+            } http2: { http2 in
+                http2.executeRequest(request)
```
fabianfett, line 104: I'm afraid this won't work with Swift 5.2:
https://github.com/apple/swift-evolution/blob/main/proposals/0279-multiple-trailing-closures.md

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
dnadoba: I removed it for now because we currently do not use it

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -469,6 +469,12 @@ extension HTTPConnectionPool {
             return jitteredBackoff
         }
     }
+
+    struct HTTP2MigrationContext {
```
fabianfett, line 473: we probably should be more explicit with the name `HTTP2ToHTTP1MigrationContext`

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -469,6 +469,12 @@ extension HTTPConnectionPool {
...
+        var generator: Connection.ID.Generator
```
fabianfett, line 474: Generator should be provided by the parent state machine.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+
+        init(generator: Connection.ID.Generator) {
+            self.generator = generator
+            self.connections = []
+        }
+        
+        // MARK: Migration
+        
+        mutating func applyMigration(_ migration: HTTP1Connections.MigrationContext) {
+            for (connectionID, eventLoop) in migration.starting {
```
fabianfett, line 306: add comment here, that in this struct we can only handle starting connections. all running connections must be transitioned which is the job of the enclosing state machine.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+                backingOffConnection.failedToConnect()
+                self.connections.append(backingOffConnection)
+            }
+        }
+
+        // MARK: Connection creation
+
+        /// used in general purpose connection scenarios to check if at least one connection exist, or if should we create a new one
+        var hasStartingOrActiveConnection: Bool {
+            self.connections.contains { $0.isStarting || $0.isActive }
```
fabianfett, line 323: we should also check if we have one that is backing off, i guess.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+        }
+
+        /// used in eventLoop scenarios. does at least one connection exist for this eventLoop, or should we create a new one?
+        /// - Parameter eventLoop: connection `EventLoop` to search for
+        /// - Returns: true if at least one connection is starting or active for the given `eventLoop`
+        func startingOrActiveConnectionExists(for eventLoop: EventLoop) -> Bool {
+            self.connections.contains {
+                $0.eventLoop === eventLoop && ($0.isStarting || $0.isActive)
```
fabianfett, line 331: we should also check backingOff

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+        }
+
+        var isStarting: Bool {
+            switch self.state {
+            case .starting, .backingOff:
+                return true
+            case .active, .draining, .closed:
+                return false
+            }
+        }
```
fabianfett, line 50: It is not clearly communicated in my mind that starting is connecting _and_ backingOff.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+extension HTTPConnectionPool {
+    private struct HTTP2ConnectionState {
+        private enum State {
+            /// the pool is establishing a connection. Valid transitions are to: .idle, .available and .closed
+            case starting
```
fabianfett, line 21: i think we should rename this to `connecting`. also change in http/1 as a separate pr.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+
+        // MARK: Leasing and releasing
+
+        mutating func leaseStream(onPreferred eventLoop: EventLoop) -> Connection? {
+            guard let index = self.findAvailableConnection(onPreferred: eventLoop, minimumAvailableStreams: 1) else { return nil }
+            return self.leaseStreams(at: index, count: 1)
+        }
+
+        /// tries to find an available connection on the prefered `eventLoop`. If it can't find one with the given `eventLoop`, it returns the first available connection
+        private func findAvailableConnection(onPreferred eventLoop: EventLoop, minimumAvailableStreams: Int) -> Int? {
```
fabianfett, line 418: When would `minimumAvailableStreams` not be `1`? This function would in my mind only be called, when a new request was just created. In this case we always only require one stream.

fabianfett: Rename `findAvailableStream`

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+                if connection.eventLoop === eventLoop {
+                    return index
+                } else if availableConnection == nil {
+                    availableConnection = index
+                }
+            }
+            return availableConnection
+        }
+        
+        mutating func leaseStreams(onRequired eventLoop: EventLoop, count: Int) -> Connection? {
```
fabianfett, line 431: When would count here be not `1`? Same question as above...

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+            guard let index = self.findAvailableConnection(onRequired: eventLoop, minimumAvailableStreams: count) else { return nil }
+            return self.leaseStreams(at: index, count: count)
+        }
+
+        /// tries to find an available connection on the required `eventLoop`
+        private func findAvailableConnection(onRequired eventLoop: EventLoop, minimumAvailableStreams: Int) -> Int? {
+            self.connections.firstIndex(where: { $0.eventLoop === eventLoop && $0.canLeaseStreams(count: minimumAvailableStreams) })
+        }
+
+        /// - precondition: `index` needs to be valid
```
fabianfett, line 441: We should give a note where to get the index from. index was received from a call to this or that with a `AvailableConnectionContext`.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+        mutating func closeConnection(at index: Int) -> Connection {
+            let connection = self.connections[index].close()
+            self.connections.remove(at: index)
+            return connection
+        }
+
+        mutating func removeConnection(at index: Int) {
+            precondition(self.connections[index].isClosed, "We tried to remove a connection which is not closed")
+            self.connections.remove(at: index)
+        }
```
fabianfett, line 471: Explain where the index is comming from. Same as above.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+        mutating func createNewConnectionByReplacingClosedConnection(at index: Int) -> (Connection.ID, EventLoop) {
+            precondition(self.connections[index].isClosed)
+            let newConnection = HTTP2ConnectionState(
+                connectionID: self.generator.next(),
+                eventLoop: self.connections[index].eventLoop
+            )
+
+            self.connections[index] = newConnection
+            return (newConnection.connectionID, newConnection.eventLoop)
+        }
```
fabianfett, line 495: where does the index come from?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,227 @@
...
+                    }
+                } else {
+                    self.requests.push(request)
+                    let connectionId = self.connections.createNewConnection(on: preferredEventLoop)
+                    return .init(
+                        request: .scheduleRequestTimeout(for: request, on: preferredEventLoop),
+                        connection: .createConnection(connectionId, on: preferredEventLoop)
+                    )
+                }
+            }
```
fabianfett, line 114: split into two functions. try to exit early.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,227 @@
...
+            )
+        }
+        
+        mutating func newHTTP2MaxConcurrentStreamsReceived(_ connectionID: Connection.ID, newMaxStreams: Int) -> Action {
+            let (index, context) = self.connections.newHTTP2MaxConcurrentStreamsReceived(connectionID, newMaxStreams: newMaxStreams)
+            return nextActionForAvailableConnection(at: index, context: context)
+        }
+        
+        mutating func http2ConnectionGoAwayReceived(_ connectionID: Connection.ID) -> Action {
+            connections.goAwayReceived(connectionID)
```
fabianfett, line 150: Here we need to check, if we have open requests for the connections el. If we do, we should create a new http2 connection right away.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -0,0 +1,532 @@
...
+                preconditionFailure("We tried to create a new connection that we know nothing about?")
+            }
+
+            self.connections[index].failedToConnect()
+            return self.connections[index].eventLoop
+        }
+
+        // MARK: Connection lifecycle events
+
+        mutating func goAwayReceived(_ connectionID: Connection.ID) {
```
fabianfett, line 387: We need to communicate the connections EL here, to make a decision to maybe start a new connection on the same el.

fabianfett: Best way is probably to create a `GoAwayContext` with the EL for now.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
```diff
@@ -69,6 +69,21 @@ extension HTTPConnectionPool {
...
+        /// removes up to `max` requests from the queue for the given `eventLoop` and returns them.
+        /// - Parameters:
+        ///   - max: maximum number of requests to pop
+        ///   - eventLoop: required event loop of the request
+        /// - Returns: requests for the given `eventLoop`
+        mutating func popFirst(max: Int, for eventLoop: EventLoop? = nil) -> [Request] {
+            if let eventLoop = eventLoop {
+                return self.withEventLoopQueue(for: eventLoop.id) { queue in
+                    queue.popFirst(max: max)
+                }
```
fabianfett, line 82: if max isn't reached here, we should further pop requests from general purpose queue

fabianfett (commented):
Starting to look REALLY good

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+RequestQueue.swift
dnadoba: already handled in `HTTP2StateMaschine`

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
dnadoba: But we are actually searching for a connection and return a connection and not a stream.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
dnadoba: I removed it now completely, we no longer use it.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
dnadoba: Added a comment. I also do no longer give this function the full `MigrationContext` and instead only pass `starting` and `backingOff` connections to make it even more clear.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,302 @@
...
+            guard let connection = self.connections.leaseStream(onPreferred: eventLoop) else {
+                self.requests.push(request)
+                return .init(request: .scheduleRequestTimeout(
+                    for: request,
+                    on: eventLoop
+                ), connection: .none)
+            }
+            return .init(
+                request: .executeRequest(request, connection, cancelTimeout: false),
+                connection: .cancelTimeoutTimer(connection.id)
```
fabianfett, line 128: same as above.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,302 @@
...
+            let remainingAvailableStreams = context.availableStreams - requests.count
+            // use the remaining available streams for requests without a required event loop
+            requests += self.requests.popFirst(max: remainingAvailableStreams, for: nil)
+            let connection = self.connections.leaseStreams(at: index, count: requests.count)
+
+            let shouldScheduleIdleTimer = context.isIdle && requests.isEmpty
+            return .init(
+                request: .executeRequestsAndCancelTimeouts(requests, connection),
+                connection: shouldScheduleIdleTimer ? .scheduleTimeoutTimer(connection.id, on: context.eventLoop) : .none
+            )
```
fabianfett, line 156: I think operating with a clear `if else` would make this code clearer.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,302 @@
...
+                maxConcurrentStreams: Self.initialMaxConcurrentStreams
+            )
+            return self.nextActionForAvailableConnection(at: index, context: context)
+        }
+
+        private mutating func nextActionForAvailableConnection(
+            at index: Int,
+            context: HTTP2Connections.AvailableConnectionContext
+        ) -> Action {
+            var requests = self.requests.popFirst(max: context.availableStreams, for: context.eventLoop)
```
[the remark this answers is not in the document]
fabianfett: We should add a comment here about the potential starving issue. And i think the comment is enough for now.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        mutating func newHTTP2MaxConcurrentStreamsReceived(_ connectionID: Connection.ID, newMaxStreams: Int) -> Action {
+            let (index, context) = self.connections.newHTTP2MaxConcurrentStreamsReceived(connectionID, newMaxStreams: newMaxStreams)
+            return self.nextActionForAvailableConnection(at: index, context: context)
+        }
+
+        mutating func http2ConnectionGoAwayReceived(_ connectionID: Connection.ID) -> Action {
+            _ = self.connections.goAwayReceived(connectionID)
+            // TODO: create new connection if we have open requests on the given event loop
```
fabianfett, line 166: or no remaining connection but still requests.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,302 @@
...
+            return .init(request: .none, connection: .createConnection(newConnectionID, on: eventLoop))
+        }
+
+        mutating func timeoutRequest(_ requestID: Request.ID) -> Action {
+            // 1. check requests in queue
+            if let request = self.requests.remove(requestID) {
+                var error: Error = HTTPClientError.getConnectionFromPoolTimeout
+                if let lastError = self.lastConnectFailure {
+                    error = lastError
+                } else if !self.connections.hasConnectionThatCanOrWillBeAbleToExecuteRequests {
```
fabianfett, line 207: We can't use that here. Here we should only check for `active` connections. If we have a starting connection, but no active we need to assume `connectTimeout`.

fabianfett (commented):
Some drive by comments... Looking better and better!

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -0,0 +1,302 @@
...
+                    request: .scheduleRequestTimeout(for: request, on: eventLoop),
+                    connection: .createConnection(connectionId, on: eventLoop)
+                )
+            }
+
+            guard let connection = self.connections.leaseStream(onRequired: eventLoop) else {
+                self.requests.push(request)
+                return .init(
+                    request: .scheduleRequestTimeout(for: request, on: eventLoop),
+                    connection: .none
```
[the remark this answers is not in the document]
dnadoba: The connection cannot be idle in that case because it already leased all its streams

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -272,6 +272,22 @@ class HTTPClientTests: XCTestCase {
...
+            configuration: HTTPClient.Configuration(certificateVerification: .none),
+            backgroundActivityLogger: Logger(label: "HTTPClient", factory: StreamLogHandler.standardOutput(label:))
+        )
+        defer {
+            XCTAssertNoThrow(try localClient.syncShutdown())
+            XCTAssertNoThrow(try localHTTPBin.shutdown())
+        }
+
+        let response = try localClient.get(url: "https://localhost:\(localHTTPBin.port)/get").wait()
+        XCTAssertEqual(.ok, response.status)
```
fabianfett, line 288:
```suggestion
        var response: HTTPClient.Response?
        response = try localClient.get(url: "https://localhost:\(localHTTPBin.port)/get").wait()
        XCTAssertEqual(response?.status, .ok)
```

PR #430, opened 2021-09-20
ChannelRead because of closing connection: Remove preconditions

fabianfett:
### Motivation

NIO may send `channelRead` events without a handlers requesting more data with a `context.read()` invocation. This happens if the remote has closed the connection and NIO wants to inform the handlers as soon as possible.

### Changes

- Don't `precondition` on `channelRead` events anymore.
- Close channel if we received an http end without a `context.read()` invocation

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
@@ -523,12 +523,22 @@ struct HTTPRequestStateMachine {
             where head.status.code < 300:
 
             return self.avoidingStateMachineCoW { state -> Action in
-                let (remainingBuffer, _) = responseStreamState.end()
-                state = .running(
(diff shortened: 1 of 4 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -536,16 +536,21 @@ struct HTTPRequestStateMachine {
             assert(producerState == .paused, "Expected to have paused the request body stream, when the head was received. Invalid state: \(self.state)")
 
             return self.avoidingStateMachineCoW { state -> Action in
-                let remainingBuffer = responseStreamState.end()
+                let (remainingBuffer, _) = responseStreamState.end()
```
Lukasa, line 539: I think it would be good to explain in a comment why here and above you don't care about the second tuple element.

Lukasa (approved):
LGTM, one request for comment.

dnadoba (approved):
LGTM

Lukasa (approved):

PR #434, opened 2021-09-22
[HTTPConnectionPool] Fix timer action races

fabianfett:
### Motivation

We have a race between some actions created in the state machine under the `stateLock` and their implementation in the `HTTPConnectionPool`. 

### Changes

- `_idleTimer` and `_backoffTimer` are protected by `stateLock`
- Added a new `struct Actions` that splits up actions from the state machine into actions that need to be executed inside the `stateLock` and outside in `stateLock`

### Result

No more races in timer scheduling.


Lukasa (approved):
Great fix, nice work!

PR #435, opened 2021-09-22
[HTTP1Connection] Add download streaming backpressure test

fabianfett:
### Motivation

In #427 we want to switch the implementation over to our new `HTTPConnectionPool`. When we do this, we want to remove the possibility to access the underlying channel. For this reason we need to remove a number of tests from `HTTPClientInternalTests`. As @weissi pointed out, we should at least still integration test downstream backpressure.

This pr replicates the behavior from `HTTPClientInternalTests.testUploadStreamingBackpressure` directly on the `HTTP1Connection`.

### Changes

- Move and adjust test from `HTTPClientInternalTests.testUploadStreamingBackpressure`
- Rename the test to `testDownloadStreamingBackpressure` since we only test incoming bytes backpressure.

### Result

An integration backpressure test, that survives #427.


Lukasa (approved):
LGTM, nice patch!

fabianfett:
`swift-nightly` failed because of Sendable requirements.

PR #437, opened 2021-09-24
[HTTPClient.Configuration] Make connection pool size configurable

### LINKED ISSUE #373: make `maximumConcurrentConnections` configurable
- `maximumConcurrentConnections` is an important property for any system using AHC, it therefore should really be configurable
- 8 seems also very low

Technically, the introduction of the "max 8 connections at a time" can be seen as API breaking in AHC version 1.1.0. There are certainly programs that worked correctly with AHC 1.0.0 (without the pool) and then broke with 1.1.0 which limited the number of concurrent connections to 8.

fabianfett:
Fixes #373

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -849,11 +849,21 @@ extension HTTPClient.Configuration {
 
     /// Connection pool configuration.
     public struct ConnectionPool: Hashable {
-        // Specifies amount of time connections are kept idle in the pool.
+        /// Specifies amount of time connections are kept idle in the pool. After this time has passed without a new
+        /// request the connections are closed.
         public var idleTimeout: TimeAmount
 
+        /// The maximum number of connections that are kept alive in the connection pool per host. If requests with
+        /// an explicit eventLoopRequirement are sent, this number might be exceeded due to overflow connections.
+        public var concurrentHTTP1ConnectionsPerHostSoftLimit: Int
+
         public init(idleTimeout: TimeAmount = .seconds(60)) {
+            self.init(idleTimeout: idleTimeout, concurrentHTTP1ConnectionsPerHostSoftLimit: 8)
+        }
+
+        public init(idleTimeout: TimeAmount, concurrentHTTP1ConnectionsPerHostSoftLimit: Int) {
             self.idleTimeout = idleTimeout
+            self.concurrentHTTP1ConnectionsPerHostSoftLimit = concurrentHTTP1ConnectionsPerHostSoftLimit
         }
     }
 }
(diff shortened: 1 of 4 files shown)


fabianfett:
Currently includes #438. Added to check, if #438 actually fixes the failures.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3060,4 +3060,44 @@ class HTTPClientTests: XCTestCase {
...
+                func makeRequest() {
+                    let url = "http://127.0.0.1:\(httpBin.port)"
+                    XCTAssertNoThrow(try client.get(url: url).wait())
+                }
+                for _ in 0..<numberOfRequestsPerThread {
+                    makeRequest()
+                }
+            }
+        }
+        let timeout = DispatchTime.now() + .seconds(180)
```
Lukasa, line 3093: This timeout is _really_ long for CI. Can we lower it?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -849,11 +849,21 @@ extension HTTPClient.Configuration {
...
 
+        /// The maximum number of connections that are kept alive in the connection pool per host. If requests with
+        /// an explicit eventLoopRequirement are sent, this number might be exceeded due to overflow connections.
+        public var concurrentHTTP1ConnectionsPerHostSoftLimit: Int
+
         public init(idleTimeout: TimeAmount = .seconds(60)) {
+            self.init(idleTimeout: idleTimeout, concurrentHTTP1ConnectionsPerHostSoftLimit: 8)
+        }
+
+        public init(idleTimeout: TimeAmount = .seconds(60), concurrentHTTP1ConnectionsPerHostSoftLimit: Int = 8) {
```
Lukasa, line 864: We should avoid a new defaulted parameter here.

dnadoba: I think we can just remove the default value and get the "behaviour" of a default value because we have the init above which only takes `idleTimeout` as a parameter.

fabianfett:
Failing `swift-nightly` ci because of Sendable requirements.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
fabianfett: I'm afraid we will need to change the number of requests and threads to something smaller, if we change the timeout to 60 seconds :)

```
Test Case 'HTTPClientTests.testConnectionPoolSizeConfigValueIsRespected' started at 2021-09-24 13:48:02.845
/code/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:3101: error: HTTPClientTests.testConnectionPoolSizeConfigValueIsRespected : failed - Timed out
/code/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:3089: error: HTTPClientTests.testConnectionPoolSizeConfigValueIsRespected : XCTAssertNoThrow failed: threw error "HTTPClientError.cancelled" - 
/code/Tests/AsyncHTTPClientTests/HTTPClientTests.swift:3089: error: HTTPClientTests.testConnectionPoolSizeConfigValueIsRespected : XCTAssertNoThrow failed: threw error "HTTPClientError.alreadyShutdown" - 
```

fabianfett: Reduced the amount of requests.

Lukasa (approved):

PR #438, opened 2021-09-24
[HTTPConnectionPool] Fix request timer scheduling

fabianfett:
### Motivation

If the allocator puts a request at the same address where another request has lived before and we ran into a timer race before, we run into an assertion. This can be seen in test runs here:

https://ci.swiftserver.group/job/async-http-client-swift55-prb/137/console

### Changes

- Remove `timerLock` from `HTTPConnectionPool`

### Result

- No more race conditions. One lock is easier than two.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
@@ -28,13 +28,12 @@ final class HTTPConnectionPool {
     private var _idleTimer = [Connection.ID: Scheduled<Void>]()
     /// The connection backoff timeout timers. Protected by the stateLock
     private var _backoffTimer = [Connection.ID: Scheduled<Void>]()
+    /// The request connection timeout timers. Protected by the stateLock
+    private var _requestTimer = [Request.ID: Scheduled<Void>]()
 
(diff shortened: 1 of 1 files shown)


fabianfett:
`swift-nightly` failing because of Sendable requirements.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -282,49 +323,29 @@ final class HTTPConnectionPool {
...
-    }
-
-    private func cancelRequestTimeouts(_ requests: [Request]) {
-        let scheduled = self.timerLock.withLock {
-            requests.compactMap {
-                self._requestTimer.removeValue(forKey: $0.id)
-            }
-        }
-        scheduled.forEach { $0.cancel() }
+        cancelTimer.cancel()
```
Lukasa, line 348: Should we do these `cancel`s outside the lock?

fabianfett: Good idea!

fabianfett: Will do at a later point, tracked here: https://github.com/swift-server/async-http-client/issues/439

Lukasa (approved):

PR #442, opened 2021-09-27
[HTTP1Connection] Handle 101 Switching Protocols

fabianfett:
### Motivation

If we currently receive a `101 Switching Protocols` response, we trigger a network reachable crash.

### Changes

- Handle `101 Switching Protocols` gracefully


Lukasa (approved):
LGTM, nice patch.

Lukasa (approved):

dnadoba (approved):

PR #443, opened 2021-09-28
Remove deprecated connection pool

fabianfett:
### Motivation

After the release of 1.6.0, we can now remove the deprecated connection pool.

### Changes

- Remove now unused Connection/ConnectionPool
- Removed TaskHandler
- Remove ResponseReadBuffer
- Remove unused imports

### Result

Less code to maintain.

PR #444, opened 2021-09-28
Replace tests that depend on `TaskHandler`

fabianfett:
### Motivation

After #427, we now want to make sure, we can eventually remove now unused code. Some tests depend on `TaskHandler`, that actually test other functionality. Work that needs to be merged before #443.

### Changes

- Replicate four test that depended on `TaskHandler`

### Result

- Code can be removed, but tests continue to test.


Lukasa (approved):
LGTM, nice work!

fabianfett:
swift-nightly fails because of sendable requirements.

PR #445, opened 2021-09-29
prepare `calculateBackoff(failedAttempt:)` to be used in `HTTP2StateMachine`

dnadoba:
### Motivation
We want to use the same backoff algorithm for HTTP1 and HTTP2.
### Changes
Make `calculateBackoff(failedAttempt:)` a static method on `HTTPConnectionPool` so it can be used in the `HTTP2StateMachine` in the future.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+Backoff.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+Backoff.swift
@@ -1,0 +1,55 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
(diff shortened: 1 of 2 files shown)


fabianfett (approved):

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+Backoff.swift
```diff
@@ -0,0 +1,55 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
```
fabianfett, line 15:
```suggestion
import NIOCore
```

Lukasa (approved):

PR #448, opened 2021-10-01
[HTTPConnectionPool] StateMachine has explicit function for HTTP1 connection close

fabianfett:
Renamed generic function `connectionClosed` to specific `http1ConnectionClose`. We should have specific callbacks for http1 and http2.


dnadoba (approved):

Lukasa (approved):
Very straightforward.

PR #451, opened 2021-10-01
[HTTP2Connections] Return if connection was idle before lease

dnadoba:
### Motivation
When we lease a new stream from a connection, we only want to cancel the idle timer if the connection was previously actually idle. 

### Changes
- all lease method now return if a connection was idle before the lease.
- adjust tests to also check the wasIdle flag


fabianfett (approved):
Good catch. LGTM.

PR #452, opened 2021-10-05
[HTTP2StateMachine] test and fix HTTP2 go away

dnadoba:
### Motivation
- we should not crash if we close a stream after we have received a go away event

### Changes
- rename `AvailableConnectionContext` to `EstablishedConnectionContext` to better communicate that the connection might actually not be available and already in the draining state
- add `connectionID` property to `EstablishedConnectionContext` because we might need to schedule an idle timeout if the connection is idle and we do not execute any new requests
- only lease streams if `count` is great then or equal to one
- add variouse tests to verify correct handling of go away events


Lukasa (approved):
Nice, this is a good cleanup and tidy fix.

PR #471, opened 2021-11-10
[HTTP2] More Integration Tests

dnadoba:


Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
```diff
@@ -187,4 +201,169 @@ class HTTP2ClientTests: XCTestCase {
...
+        task = client.execute(
+            request: request,
+            delegate: delegate
+        )
+
+        XCTAssertThrowsError(try task.futureResult.timeout(after: .seconds(2)).wait(), "Should fail") { error in
+            guard case let error = error as? HTTPClientError, error == .cancelled else {
+                return XCTFail("Should fail with cancelled")
+            }
+        }
```
fabianfett, line 267:
```suggestion
        XCTAssertThrowsError(try task.futureResult.timeout(after: .seconds(2)).wait()) {
            XCTAssertEqual($0 as? HTTPClientError, . cancelled)
        }
```

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
```diff
@@ -187,4 +201,169 @@ class HTTP2ClientTests: XCTestCase {
...
+        for result in results {
+            switch result {
+            case .success:
+                XCTFail("Shouldn't succeed")
+            case .failure(let error):
+                if let clientError = error as? HTTPClientError, clientError == .cancelled {
+                    continue
+                } else {
+                    XCTFail("Unexpected error: \(error)")
+                }
```
fabianfett, line 236:
```suggestion
                XCTAssert(error as? HTTPClientError, .cancelled)
```

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
```diff
@@ -187,4 +201,169 @@ class HTTP2ClientTests: XCTestCase {
...
+            )
+            return task
+        }
+
+        for task in tasks {
+            XCTAssertThrowsError(try task.futureResult.timeout(after: .seconds(2)).wait(), "Should fail") { error in
+                guard case let error = error as? HTTPClientError, error == .cancelled else {
+                    return XCTFail("Should fail with cancelled")
+                }
+            }
```
fabianfett, line 305:
```suggestion
            XCTAssertThrowsError(try task.futureResult.timeout(after: .seconds(2)).wait()) {
                XCTAssertEqual($0 as? HTTPClientError, .cancelled)
            }
```

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -804,6 +813,10 @@ internal final class HTTPBinHandler: ChannelInboundHandler {
                 builder.add(buff)
                 self.resps.append(builder)
                 return
+            case "/sendheaderandwait":
+                // sends some headers and waits indefinitely afterwards
+                context.writeAndFlush(wrapOutboundOut(.head(HTTPResponseHead(version: HTTPVersion(major: 1, minor: 1), status: .ok))), promise: nil)
+                return
```
fabianfett, line 819: Please create a small `ChannelInboundHandler` for this in the specific test case. There is so much stuff in this single handler and all is mixed.

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -49,11 +49,16 @@ func getDefaultEventLoopGroup(numberOfThreads: Int) -> EventLoopGroup {
...
+    init(
+        backpressureEventLoop: EventLoop? = nil,
+        stateDidChangeCallback: ((State) -> Void)? = nil
+    ) {
         self.backpressureEventLoop = backpressureEventLoop
+        self.stateDidChangeCallback = stateDidChangeCallback
     }
 
     var backpressureEventLoop: EventLoop?
+    var stateDidChangeCallback: ((State) -> Void)?
```
fabianfett, line 61: this should be at least a let. however in your tests your mainly using this to get a callback once the head was received. I think, a specialized delegate only for this one job, might be a better idea.

fabianfett (changes requested):

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
```diff
@@ -67,7 +82,7 @@ class HTTP2ClientTests: XCTestCase {
     func testConcurrentRequestsFromDifferentThreads() {
         let bin = HTTPBin(.http2(compress: false))
         defer { XCTAssertNoThrow(try bin.shutdown()) }
-        let client = self.makeDefaultHTTPClient()
+        let client = self.makeClientWithActiveHTTP2Connection(to: bin)
```
fabianfett, line 85: why do we need a client with an active connection?

dnadoba: good catch, we don't.

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
```diff
@@ -187,4 +202,185 @@ class HTTP2ClientTests: XCTestCase {
         // all workers should be running, let's wait for them to finish
         allDone.wait()
     }
+
+    func testUncleanShutdownCancelsExecutingAndQueuedTasks() {
+        let bin = HTTPBin(.http2(compress: false))
+        defer { XCTAssertNoThrow(try bin.shutdown()) }
+        let clientGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer { XCTAssertNoThrow(try clientGroup.syncShutdownGracefully()) }
+        let client = self.makeClientWithActiveHTTP2Connection(to: bin, eventLoopGroupProvider: .shared(clientGroup))
```
fabianfett, line 211: Why do we need a client with an active connection here? Would love a comment...

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
```diff
@@ -187,4 +202,185 @@ class HTTP2ClientTests: XCTestCase {
...
+private final class SendHeaderAndWaitChannelHandler: ChannelInboundHandler {
+    typealias InboundIn = HTTPServerRequestPart
+    typealias OutboundOut = HTTPServerResponsePart
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        context.writeAndFlush(wrapOutboundOut(.head(HTTPResponseHead(
+            version: HTTPVersion(major: 1, minor: 1),
+            status: .ok
+        ))
+        ), promise: nil)
```
fabianfett, line 384: This will send at least two response headers for each request... you'll need to unwrap the incoming requests for `.head` and `.end`

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -49,11 +49,16 @@ func getDefaultEventLoopGroup(numberOfThreads: Int) -> EventLoopGroup {
 class TestHTTPDelegate: HTTPClientResponseDelegate {
     typealias Response = Void
 
-    init(backpressureEventLoop: EventLoop? = nil) {
+    init(
+        backpressureEventLoop: EventLoop? = nil,
+        stateDidChangeCallback: ((State) -> Void)? = nil
+    ) {
```
fabianfett, line 55: I think all changes in this file can now be reverted, can't they?

fabianfett (commented):
Three small things at the end.

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
```diff
@@ -187,4 +202,185 @@ class HTTP2ClientTests: XCTestCase {
...
+    func didReceiveHead(task: HTTPClient.Task<Void>, _ head: HTTPResponseHead) -> EventLoopFuture<Void> {
+        self.didReceiveHeadCallback(head)
+        return task.eventLoop.makeSucceededVoidFuture()
+    }
+
+    func didFinishRequest(task: HTTPClient.Task<Void>) throws {}
+}
+
+/// sends some headers and waits indefinitely afterwards
+private final class SendHeaderAndWaitChannelHandler: ChannelInboundHandler {
```
fabianfett, line 375: oh we should use this one to test idleReadTimeout as well. can be another pr.

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
dnadoba: We need an active connection to guarantee that requests are executed immediately without waiting for connection establishment. Maybe we could also write a test that observes when a connection has been established and only starts shutdown afterwards.

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
dnadoba: could we also use `channelActive` instead?

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
fabianfett: would you mind putting this into the test?

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
fabianfett: nope... you need to wait for the request.

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
dnadoba: Done

Tests/AsyncHTTPClientTests/HTTP2ClientTests.swift
dnadoba: okay thanks, we now only send it after we received `.end`.

fabianfett (approved):

Issue #231, opened 2020-05-28
deprecate `ignoreUncleanSSLShutdown: Bool` parameter 


weissi:
We should probably deprecate the `ignoreUncleanSSLShutdown` parameter. AsyncHTTPClient _knows_ when it can and when it cannot ignore an unclean SSL shutdown.

Basically if an `.end` was received and we have either a `content-length` OR a `transfer-encoding: chunked` header from the server, we can safely ignore. For HTTP/2 we can always ignore.

If we do have neither a `content-length`, nor a `transfer-encoding: chunked` header, then this should error.

weissi:
@Lukasa does this sound right?

Lukasa:
_Probably_, with the caveat that in principle we may want users to be able to tolerate even the last case if they're sure they should. But frankly I think deprecating sounds good.

weissi:
I agree with that. My thinking was that EOF framing over TLS should be super rare (am I right here?), and even rarer without CloseNotify, so I don't think that's a thing we need to support. And if we do then we can add a new one with a more scary name :). `ignoreTrulyInsecureUncleanTLSShutdowns: Bool` :)

Lukasa:
I believe it should be super rare, yes. As noted, I’m happy to deprecate.

weissi:
For more context: #238 contains the full explanation why we don't need this parameter.

PR #472, opened 2021-11-10
Handle NIOSSLError.uncleanShutdown correctly

### LINKED ISSUE #238: `uncleanShutdown` should be handled automatically (where possible)
TLS (SSL) encrypts, validates, and authenticates all the data that goes through the connection. That is a fantastic property to have and solves most issues. TLS however still runs over TCP and the control packets of TCP are not encrypted. This means that the one thing an attacker can still do on a TLS connection is to close it. The attacker could send RST or FIN packets the other peer and the receiving peer has no means to verify if this RST/FIN packet is actually coming from the other peer (as opposed to an attacker).
To fix this problem, TLS introduces a close_notify message that is send over connection as encrypted data. So if the other peer receives a close_notify it knows that it was truly sent by the other peer (and not an attacker). A well behaving peer would then reply to the `close_notify` with its own `close_notify` and after that both peers can close the TCP connection because they know that the respective other peer knows they're okay closing it.

Okay, but what's the issue with having a connection just close. Wouldn't I notice that part of the data is missing? The answer is it depends. Many protocols actually send to the other peer how much data they will send before sending the data. And they also have a well defined "end message". If you're using such a protocol, then an "unclean shutdown" (which is you have received a RST/FIN without a close_notify) is totally harmless. The higher level protocol allows you to distinguish between a truncated message (when there's still outstanding data) and a close after a completed message.

The reason SwiftNIO sends you a `NIOSSLError.uncleanShutdown` if it sees a connection closing without a prior `close_notify` is because it doesn't know what protocol you're implementing. Maybe the protocol you speak doesn't transmit length information so a truncated message cannot be told apart from a complete message.

Let's go into some example protocols and their behaviour regarding framing:
- With HTTP/2 the other peer always knows how much data to expect, so an unclean shutdown is totally harmless, the error can be ignored.
- With HTTP/1, the situation is much more complicated: HTTP/1 when using the `content-length`  header is unaffected by truncation attacks because we know the content length. So if the connection closes before we have received that many bytes, we know it was a truncated message (either by an attacker injecting a FIN/RST, or by a software/network problem somewhere along the way)
- HTTP/1 with transfer-encoding: chunked is also unaffected by truncation attacks because each chunk sends its length and there's a special "END" chunk (`0\r\n\r\n`). Unfortunately HTTP/1 can be used without either `transfer-encoding: chunked` or `content-length`. It then runs in the "EOF framing" mode which means that the message ends when we receive a connection close :cry: . Very bad, if HTTP/1 is used in "EOF framing" mode, then an unclean shutdown is actually a real error because we cannot tell a truncated message apart from a complete message.

Why should AHC do better here? Well, AHC knows if its speaking HTTP/1 or HTTP/2 and whether the server sent a `content-length` or `transfer-encoding: chunked` or not. Therefore, AHC should suppress the uncleanShutdown error in all cases apart from when HTTP/1 is run in "EOF framing mode".

The good news is that this will almost never happen. Nobody in their right mind uses EOF framing mode.

fabianfett:
### Motivation

Fixes #238 and #231.

### Changes

- Extracted the unclean shutdown test from `HTTPClientTests` into their own file `HTTPClientUncleanSSLConnectionShutdownTests`
- Copy and pasted @weissi great explanation from #238 into the test file
- Removed property `ignoreUncleanSSLShutdown` everywhere

### Result

`ignoreUncleanSSLShutdown` on `HTTPClient.Configuration` is deprecated and ignored.


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -601,7 +601,11 @@ public class HTTPClient {
         /// Enables automatic body decompression. Supported algorithms are gzip and deflate.
         public var decompression: Decompression
         /// Ignore TLS unclean shutdown error, defaults to `false`.
-        public var ignoreUncleanSSLShutdown: Bool
+        @available(*, deprecated, message: "AsyncHTTPClient now correctly supports handling unexpected SSL connection drops. This property is ignored")
```
weissi, line 604:
```suggestion
        @available(*, deprecated, message: "AsyncHTTPClient now correctly supports handling unexpected TLS connection drops. This property is ignored")
```

weissi (approved):
awesome, thank you so much!

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -201,17 +198,37 @@ struct HTTPRequestStateMachine {
...
+            // this error.
             return .wait
 
+        case .running(.streaming, .receivingBody(let responseHead, _)),
+             .running(.endSent, .receivingBody(let responseHead, _)) where error as? NIOSSLError == .uncleanShutdown:
+
+            // if we have already received the response head, the parser will ensure that we receive
+            // the complete response. we can ignore this error. we might see a HTTPParserError very
+            // soon.
+            if responseHead.headers.contains(name: "content-length") || responseHead.headers.contains(name: "transfer-encoding") {
```
Lukasa, line 214: This is not a sufficient check: there are cases where these headers are not present, but we are nonetheless confident of the length of the response. Specifically:

1. Responses to a HEAD request
2. Responses with a 204 or 304 status code

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -201,17 +198,37 @@ struct HTTPRequestStateMachine {
...
+            // this like a normal connection close. We will receive a call to channelInactive after
+            // this error.
             return .wait
 
+        case .running(.streaming, .receivingBody(let responseHead, _)),
+             .running(.endSent, .receivingBody(let responseHead, _)) where error as? NIOSSLError == .uncleanShutdown:
+
+            // if we have already received the response head, the parser will ensure that we receive
+            // the complete response. we can ignore this error. we might see a HTTPParserError very
+            // soon.
```
Lukasa, line 213: This comment seems to be out of step with the code: it implies that we can ignore the error, but then has logic that sometimes doesn't ignore the error.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
fabianfett: The `HTTPResponseDecoder` will always emit a response `.end` after having received a response head, if the request had a `HEAD` or `CONNECT` method or the response status is 204 or 304. 

https://github.com/apple/swift-nio/blob/main/Sources/NIOHTTP1/HTTPDecoder.swift#L301-L314

For this reason, we can expect that we will only be in the streaming state, if we have a "content-length" or "transfer-encoding" header or if the response is EOF terminated. I'm well aware that this is _very_ implicit. If we want to have explicit checks here, I will need to come up with a way to preserve the request method a little longer.

Lukasa: Hmm, can we get this logic into a comment then? The correctness of this code is not apparent merely from reading it.

fabianfett: Comment added.

Lukasa (approved):

Issue #28, opened 2019-05-08
should support HTTP/2


weissi:
HTTP/2 support. Partial example can be found in [swift-nio-examples/http2-client](https://github.com/apple/swift-nio-examples/blob/master/http2-client/Sources/http2-client/main.swift). This does HTTP/2 with 'prior knowledge', for this package we also need to support HTTP/2 upgrade

weissi:
apart from the obvious use-case -- to connect to HTTP/2 web servers -- this library could/should then also be used for [swift-nio-apns](https://github.com/kylebrowning/swift-nio-apns/).

weissi:
@artemredkin/@ianpartridge/@tanner0101 should we pull this into the `1.0.0` milestone? I think it would be important to verify that the API actually works for HTTP/2. I don't foresee any issues but it'd be good to validate.

ianpartridge:
I think we can ship `1.0` without HTTP/2. Adding connection pooling and HTTP/2 sounds like a great set of 2.0 features and we can make any necessary API changes then.

Lukasa:
I’m inclined to agree: it seems more important to enable users to start using the library than to throw HTTP/2 support into the client.

artemredkin:
Same here

fabianfett:
Hi,

I took some time and looked into this the last days... I haven't started programming or anything, but I think I should at least document the open questions/ideas I ran into:

I think we should go the lazy route ([like most browsers except Safari](https://caniuse.com/http2)) and first only support http/2 over TLS. The protocol negotiation should be fairly easy this way and we know before sending the first HTTP request, if we will talk HTTP1 or HTTP2 on this connection. 

I think we should be able to achieve the simpler goal with the current architecture. Is this viable?

#### Further Questions 

- How many concurrent h/2 connections do we want to support to a host? Only one? What do we want to do with `EventLoopPreference`s?
- What shall happen if an HTTP/2 connection already has the maximum number of concurrent requests running? [`SETTINGS_MAX_CONCURRENT_STREAMS`](https://tools.ietf.org/html/rfc7540#section-6.5.2) (Especially important for something like sending push notifications)
- Do we want to rename the `HTTP1ConnectionProvider` to `HTTPConnectionProvider` and save in the `ConnectionsState` whether this provider supports connections over HTTP/1.1 or HTTP/2? We should know after the first successful TLS connection.

If you have any other design thoughts in the back of your heads, I'd be happy to know them. 😉

[...]

PR #473, opened 2021-11-10
[HTTP2] Make HTTPVersion public and set to .automatic by default

dnadoba:
This PR does two things:
- makes it possible to enabling and disabling HTTP/2 by making `HTTPVersion` public
- set `HTTPVersion` to `.automatic` by default

HTTP/2 will then be used if we connect to a server with HTTPS and the server supports HTTP/2, otherwise we will continue to use HTTP/1.

A user will still be able to disable HTTP/2 if they want to by setting `HTTPVersion` to `.http1Only` on `HTTPClient.Configuration` e.g.:
```
var configuration = HTTPClient.Configuration()
configuration.httpVersion = .http1Only
let client = HTTPClient(
    eventLoopGroupProvider: .createNew,
    configuration: configuration
)
```

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -607,10 +607,8 @@ public class HTTPClient {
             set {}
         }
 
-        // TODO: make public
-        // TODO: set to automatic by default
-        /// HTTP/2 is by default disabled
-        internal var httpVersion: HTTPVersion
+        /// is set to `.automatic` by default which will use HTTP/2 if run over https and the server supports it, otherwise HTTP/1
+        public var httpVersion: HTTPVersion
 
         public init(
             tlsConfiguration: TLSConfiguration? = nil,
@@ -628,13 +626,11 @@ public class HTTPClient {
                 proxy: proxy,
                 ignoreUncleanSSLShutdown: ignoreUncleanSSLShutdown,
                 decompression: decompression,
-                // TODO: set to automatic by default
-                httpVersion: .http1Only
+                httpVersion: .automatic
             )
         }
 
-        // TODO: make public
-        internal init(
+        public init(
             tlsConfiguration: TLSConfiguration? = nil,
             redirectConfiguration: RedirectConfiguration? = nil,
             timeout: Timeout = Timeout(),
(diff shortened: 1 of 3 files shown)


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -624,13 +622,11 @@ public class HTTPClient {
...
                 decompression: decompression,
-                // TODO: set to automatic by default
-                httpVersion: .http1Only
+                httpVersion: .automatic
             )
         }
 
-        // TODO: make public
-        internal init(
+        public init(
```
fabianfett, line 629: TBH: I don't think we should add yet another initializer for HTTPClient.Configuration. I think we should deprecate all of them, except for `init()`. In `init()` we set all default values and then the user can change the config, since it is just a struct. We have precedent for this move in NIOSSL.

fabianfett:
Fixes #28.

Sources/AsyncHTTPClient/HTTPClient.swift
dnadoba: If I understand you correct, a user will then only be able to change the config by create a temporary variable, initialising it with our default config, mutating it and passing it to the `HTTPClient` initialiser, right? This would be a rather big change because we would need to update all our tests if we deprecate all other inits. I would rather do that in another PR if we would like to do that before merging this one.

fabianfett: cc @Lukasa

dnadoba: We could also think about adding an extension to `HTTPClient.Configuration` so we can change the configuration through daisy chaining like [I have added in RSocket](https://github.com/rsocket/rsocket-swift/blob/df80e64fa0d16e00b1a192c475e47b87c919c0a1/Sources/RSocketCore/Client/ClientConfiguration.swift#L107-L115):
```swift
extension HTTPClient.Configuration {
    /// Creates a copy and updates the property described by `keyPath` with the given `newValue`
    /// - Returns: updated configuration
    public func set<Value>(_ keyPath: WritableKeyPath<Self, Value>, to newValue: Value) -> Self {
        var copy = self
        copy[keyPath: keyPath] = newValue
        return copy
    }
}
```

This would easy the transition because you probably end up with the same amount of lines as a config and a user does not have to add a temporary variable.

dnadoba: When we are already on it, what about nested types `HTTPClient.Configuration` uses e.g. `HTTPClient.Configuration.ConnectionPool` and `HTTPClient.Configuration.Timeout`? They will probably suffer from the same problem, but on a much smaller scale.

Lukasa: I'm inclined to tackle this in a separate PR. For now we don't need to add a new init for this feature, we can just exclude it from the pre-existing inits.

fabianfett (approved):
LGTM! It's great that we can finally expose this!

Lukasa (approved):

PR #478, opened 2021-11-17
Only crash in debug mode, if HTTPClient was not shutdown

fabianfett:
### Motivation

Generally we want to inform users that they need to shutdown their HTTPClient. Until `1.6.0` we did this with an assert in HTTPClient's deinit. With `1.6.0` this behavior was raised to a precondition. This could lead to users to discover suddenly crashes in production, where they have seen none before.

### Changes

- This pr reverts the current behavior back to something pre `1.6.0`

### Result

- HTTPClient doesn't crash in production anymore.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -68,7 +68,7 @@ public class HTTPClient {
     let eventLoopGroupProvider: EventLoopGroupProvider
     let configuration: Configuration
     let poolManager: HTTPConnectionPool.Manager
-    var state: State
+    private var state: State
     private let stateLock = Lock()
 
     internal static let loggingDisabled = Logger(label: "AHC-do-not-log", factory: { _ in SwiftLogNoOpLogHandler() })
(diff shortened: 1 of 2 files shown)


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -118,8 +118,14 @@ public class HTTPClient {
...
 
     deinit {
-        guard case .shutDown = self.state else {
-            preconditionFailure("Client not shut down before the deinit. Please call client.syncShutdown() when no longer needed.")
+        debugOnly {
+            // We want to crash only in debug mode.
+            switch self.state {
+            case .shutDown:
+                break
+            case .shuttingDown, .upAndRunning:
```
glbrntt, line 126: Is it possible for `.shuttingDown` to ever be valid here? I.e. if using the async shutdown?

fabianfett: You are right, when we are shuttingDown, the HTTPClient is in a retain cycle, that should prevent it from ever going out of memory. Do we want to call this out explicitly?

glbrntt: Sure, no harm in it.

glbrntt (approved):

PR #479, opened 2021-11-18
Always overwrite `Transport-Encoding` and `Content-Length`

dnadoba:
### Motivation
We want to reuse request validation in the upcoming async/await implementation. During refactoring we noticed some inconsistency in the current implementation on how user defined `Transport-Encoding` and `Content-Length` headers are processed. We concluded that a user should not be allowed to set `Transport-Encoding` or `Content-Length` headers themselves. A user can only communicate its intend through the `body` property on a request. We infer the correct and most optimal `Transport-Encoding` or `Content-Length` header for them.

### Changes
- `Transport-Encoding` and `Content-Length` or no longer validated and always removed or replaced
- rename `validate` to `validateAndSetTransportFraming` because it does not only validate but also mutate the `HTTPHeaders` to include the correct headers for the given request body.
- introduce a new enum `RequestBodyLength` which can be either `.dynamic` if we do not now the size of a request in advance or `.fixed(length: Int)` if the user has specified the length of the body.
- refactor implementation to no longer expect a `HTTPClient.Request.Body?` but rather a new type `RequestBodyLength`. This allows use to reuse the function for async/await because we will not use `HTTPClient.Request.Body` but rather a new type `HTTPRequest.Body`. 
- no request body (where `HTTPClient.Request.body` is set to nil) and a request body with a length of `0` (e.g where body is `.byteBuffer(ByteBuffer())` are now processed the same way.

### Alternatives
We could also remove `RequestBodyLength` and just use and optional `Int`.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
@@ -1,0 +1,20 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+enum RequestBodyLength: Equatable {
+    /// size of the reuqest body is not known before starting the request
+    case dynamic
+    /// size of the request body is fixed and exactly `length` bytes
+    case fixed(length: Int)
+}
--- a/Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 enum RequestBodyLength: Equatable {
-    /// size of the reuqest body is not known before starting the request
+    /// size of the request body is not known before starting the request
     case dynamic
     /// size of the request body is fixed and exactly `length` bytes
     case fixed(length: Int)
--- a/Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-enum RequestBodyLength: Equatable {
+enum RequestBodyLength: Hashable {
     /// size of the request body is not known before starting the request
     case dynamic
     /// size of the request body is fixed and exactly `length` bytes
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -883,7 +883,7 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         case remoteConnectionClosed
         case cancelled
         case identityCodingIncorrectlyPresent
-        @available(*, deprecated)
+        @available(*, deprecated, message: "AsyncHTTPClient now silently corrects this invalid header.")
         case chunkedSpecifiedMultipleTimes
         case invalidProxyResponse
         case contentLengthMissing
@@ -895,7 +895,7 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         case invalidHeaderFieldNames([String])
         case bodyLengthMismatch
         case writeAfterRequestSent
-        @available(*, deprecated)
+        @available(*, deprecated, message: "AsyncHTTPClient now silently corrects invalid headers.")
         case incompatibleHeaders
         case connectTimeout
         case socksHandshakeTimeout
@@ -939,7 +939,7 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
     /// Request contains invalid identity encoding.
     public static let identityCodingIncorrectlyPresent = HTTPClientError(code: .identityCodingIncorrectlyPresent)
     /// Request contains multiple chunks definitions.
-    @available(*, deprecated)
+    @available(*, deprecated, message: "AsyncHTTPClient now silently corrects this invalid header.")
     public static let chunkedSpecifiedMultipleTimes = HTTPClientError(code: .chunkedSpecifiedMultipleTimes)
     /// Proxy response was invalid.
     public static let invalidProxyResponse = HTTPClientError(code: .invalidProxyResponse)
@@ -962,7 +962,7 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
     /// Body part was written after request was fully sent.
     public static let writeAfterRequestSent = HTTPClientError(code: .writeAfterRequestSent)
     /// Incompatible headers specified, for example `Transfer-Encoding` and `Content-Length`.
-    @available(*, deprecated)
+    @available(*, deprecated, message: "AsyncHTTPClient now silently corrects invalid headers.")
     public static let incompatibleHeaders = HTTPClientError(code: .incompatibleHeaders)
     /// Creating a new tcp connection timed out
     public static let connectTimeout = HTTPClientError(code: .connectTimeout)
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -884,6 +884,7 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         case cancelled
         case identityCodingIncorrectlyPresent
         case chunkedSpecifiedMultipleTimes
+        case transferEncodingSpecifiedButChunkedIsNotTheFinalEncoding
         case invalidProxyResponse
         case contentLengthMissing
         case proxyAuthenticationRequired
@@ -938,6 +939,8 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
     public static let identityCodingIncorrectlyPresent = HTTPClientError(code: .identityCodingIncorrectlyPresent)
     /// Request contains multiple chunks definitions.
     public static let chunkedSpecifiedMultipleTimes = HTTPClientError(code: .chunkedSpecifiedMultipleTimes)
+    /// Request specifies `Transfer-Encoding` but `chunked` is not the final encoding
+    public static let transferEncodingSpecifiedButChunkedIsNotTheFinalEncoding = HTTPClientError(code: .transferEncodingSpecifiedButChunkedIsNotTheFinalEncoding)
     /// Proxy response was invalid.
     public static let invalidProxyResponse = HTTPClientError(code: .invalidProxyResponse)
     /// Request does not contain `Content-Length` header.
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -883,8 +883,8 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         case remoteConnectionClosed
         case cancelled
         case identityCodingIncorrectlyPresent
+        @available(*, deprecated)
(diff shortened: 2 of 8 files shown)


Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -16,93 +16,79 @@ import NIOCore
...
-        if self.contains(name: "Transfer-Encoding"), self.contains(name: "Content-Length") {
+    mutating func validateAndFixTransportFraming(
+        method: HTTPMethod,
+        bodyLength: RequestBodyLength
+    ) throws -> RequestFramingMetadata {
+        let contentLength = self.first(name: "Content-Length")
+        let encodings = self[canonicalForm: "Transfer-Encoding"]
+
+        // "Transfer-Encoding" and "Content-Length" are not allowed to present at the same time (https://tools.ietf.org/html/rfc7230#section-3.3.1)
+        guard encodings.isEmpty || contentLength == nil else {
```
fabianfett, line 27: Super NIT: I think an if is easier to read here.

Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -16,93 +16,79 @@ import NIOCore
...
-            self.add(name: "Content-Length", value: String(length))
-            metadata.body = .fixedSize(length)
+            return .init(connectionClose: connectionClose, body: .stream)
         }
-
-        return metadata
     }
 
-    private func validateFieldNames() throws {
+    func validateFieldNames() throws {
```
fabianfett, line 91: why is this internal now?

Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -16,93 +16,79 @@ import NIOCore
...
+            if case .TRACE = method {
+                // A client MUST NOT send a message body in a TRACE request.
+                // https://tools.ietf.org/html/rfc7230#section-4.3.8
+                throw HTTPClientError.traceRequestWithBody
+            }
+            if encodings.isEmpty {
+                self.add(name: "Content-Length", value: String(length))
+                return .init(connectionClose: connectionClose, body: .fixedSize(length))
+            } else {
+                self.add(name: "Transfer-Encoding", value: encodings.joined(separator: ", "))
```
fabianfett, line 70: I think we should make this explicit with a comment.

> Even though a fixed length body is going to be sent, the user explicitly set an encoding. We should respect the users wish and use the specified headers.

Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -16,93 +16,79 @@ import NIOCore
...
-                transferEncoding = "chunked"
+            return .init(connectionClose: connectionClose, body: .fixedSize(0))
+        case .fixed(let length):
+            if case .TRACE = method {
+                // A client MUST NOT send a message body in a TRACE request.
+                // https://tools.ietf.org/html/rfc7230#section-4.3.8
+                throw HTTPClientError.traceRequestWithBody
+            }
+            if encodings.isEmpty {
+                self.add(name: "Content-Length", value: String(length))
```
fabianfett, line 67: There might be an interesting case here. What if the user handed us a fixed size body payload, but also set the Content-Length header to something different?

Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -16,93 +16,79 @@ import NIOCore
...
-            if !encodings.contains("chunked") {
-                guard let length = body.length else {
-                    throw HTTPClientError.contentLengthMissing
-                }
-                contentLength = length
+            if encodings.isEmpty && contentLength == nil {
+                // if a user forgot to specify a Content-Length and Transfer-Encoding, we will set it for them
+                self.add(name: "Transfer-Encoding", value: "chunked")
+            } else {
+                self.add(name: "Transfer-Encoding", value: encodings.joined(separator: ", "))
```
fabianfett, line 84: We also run in this case if the user has set a Content-Length header explicitly. I think in the `dynamic` case we should trust a user set content length header (though we should validate it is a positive int).

A case might be streaming a file somewhere.

Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -137,4 +123,33 @@ extension HTTPHeaders {
             throw HTTPClientError.invalidHeaderFieldNames(invalidFieldNames)
         }
     }
+
+    static func validateTransferEncoding<Encodings>(
+        _ encodings: Encodings
+    ) throws where Encodings: Sequence, Encodings.Element: StringProtocol {
```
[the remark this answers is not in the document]
dnadoba: We call this function during `validateAndFixTransportFraming` with `[Substring]` but it is more convenience to use `[String]` in tests.

Sources/AsyncHTTPClient/RequestValidation.swift
dnadoba: I once had validation in there but noticed that we actually have tests that check that we always overwrite the `Content-Length` header. I would also prefer to validate it but this might break user code. WDYT?

dnadoba: https://github.com/swift-server/async-http-client/blob/62ad8183765c23c48de67aa8ff08b653d679e0b4/Tests/AsyncHTTPClientTests/RequestValidationTests.swift#L43-L49

Sources/AsyncHTTPClient/RequestValidation.swift
dnadoba: I'm not really sure about this. A user can always specify the length of a stream upfront through our API and doesn't need to set the `Content-Length` header manually through the `length` parameter.
https://github.com/swift-server/async-http-client/blob/62ad8183765c23c48de67aa8ff08b653d679e0b4/Sources/AsyncHTTPClient/HTTPHandler.swift#L61-L69
This would give them a way to do the same thing in to different ways and they may conflict.
I would almost say that we should always ignore/remove "Content-Length" header.
The same is almost true for `Transport-Encoding` but I'm not fully understanding all use cases for this one.

dnadoba: If I read this code snipped correctly, `Transport-Encoding` can only be `chunked` because SwiftNIO always replace the `Transport-Encoding` with `chunked` or doesn't do chunked encoding if it is not the only encoding. 
https://github.com/apple/swift-nio/blob/addf69cfe60376c325397c8926589415576b1dd1/Sources/NIOHTTP1/HTTPEncoder.swift#L104-L129

Sources/AsyncHTTPClient/RequestValidation.swift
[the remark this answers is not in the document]
fabianfett: I think this should be private and all tests should go directly through `validateAndFixTransportFraming`.

Sources/AsyncHTTPClient/RequestValidation.swift
fabianfett: We need @Lukasa's wisdom here.

Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -16,93 +16,79 @@ import NIOCore
...
-            metadata.connectionClose = true
-        }
-
-        // validate transfer encoding and content length (https://tools.ietf.org/html/rfc7230#section-3.3.1)
-        if self.contains(name: "Transfer-Encoding"), self.contains(name: "Content-Length") {
+    mutating func validateAndFixTransportFraming(
+        method: HTTPMethod,
+        bodyLength: RequestBodyLength
+    ) throws -> RequestFramingMetadata {
+        let contentLength = self.first(name: "Content-Length")
```
Lukasa, line 23: Why are we getting the value if we don't need it?

Sources/AsyncHTTPClient/RequestValidation.swift
```diff
@@ -16,93 +16,79 @@ import NIOCore
...
+        let encodings = self[canonicalForm: "Transfer-Encoding"]
+
+        // "Transfer-Encoding" and "Content-Length" are not allowed to present at the same time (https://tools.ietf.org/html/rfc7230#section-3.3.1)
```
Lukasa, line 26: Nit:

```suggestion
        // "Transfer-Encoding" and "Content-Length" are not allowed to be present at the same time (https://tools.ietf.org/html/rfc7230#section-3.3.1)
```

Sources/AsyncHTTPClient/RequestValidation.swift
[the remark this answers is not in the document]
Lukasa: +1 to @fabianfett's note.

Sources/AsyncHTTPClient/RequestValidation.swift
fabianfett: > As to supporting Content-Length: yes, we should emit it, but we should do so when users tell us the size of their stream. We shouldn't tolerate them setting that header.

Okay, I'm happy with that!

In that case we need to fix the code comment though here:
https://github.com/swift-server/async-http-client/blob/62ad8183765c23c48de67aa8ff08b653d679e0b4/Sources/AsyncHTTPClient/HTTPHandler.swift#L61-L69

If no Content-Length header is set, we will just reach for chunked.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -937,6 +939,7 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
     /// Request contains invalid identity encoding.
     public static let identityCodingIncorrectlyPresent = HTTPClientError(code: .identityCodingIncorrectlyPresent)
     /// Request contains multiple chunks definitions.
+    @available(*, deprecated)
```
fabianfett, line 942: We should add a message here like: "AsyncHTTPClient now silently corrects invalid headers."

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -959,6 +962,7 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
     /// Body part was written after request was fully sent.
     public static let writeAfterRequestSent = HTTPClientError(code: .writeAfterRequestSent)
     /// Incompatible headers specified, for example `Transfer-Encoding` and `Content-Length`.
+    @available(*, deprecated)
```
fabianfett, line 965: See above

Tests/AsyncHTTPClientTests/RequestValidationTests.swift
```diff
@@ -336,4 +320,34 @@ class RequestValidationTests: XCTestCase {
         let req6 = try! HTTPClient.Request(url: "https://localhost/get", headers: ["host": "foo"])
         XCTAssertEqual(try req6.createRequestHead().0.headers["host"].first, "foo")
     }
+
+    func testTraceMethodIsNotAllowedToHaveAFixedLengthBody() {
+        var headers = HTTPHeaders()
+        XCTAssertThrowsError(try headers.validateAndSetTransportFraming(method: .TRACE, bodyLength: .fixed(length: 10)))
```
fabianfett, line 326: Since the errors are equatable would you mind checking, we get what we expect?

fabianfett (commented):
Wow, this looks so much better already. Tiny nits.

Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+enum RequestBodyLength: Equatable {
```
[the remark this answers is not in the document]
dnadoba: Is the binary size increase negligible for the additional synthesised conformance? AFAIK Swift doesn't remove unused synthesised conformances, even for internal types.

fabianfett (approved):
Huge improvement!

Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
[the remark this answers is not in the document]
Lukasa: Yeah, this is not the most fruitful way of shrinking binary sizes.

Lukasa (approved):

PR #480, opened 2021-11-18
Remove redundant `RequestFramingMetadata.Body.none` case

dnadoba:
### Motivation
The distinction between no request body (`.none`) and a request body of size `0` (`.fixedSize(0)`) is unnecessary and we explicitly have code to ignore this:
https://github.com/swift-server/async-http-client/blob/ec2e080d7011a81bd67f10bf41efe6104d7799d6/Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift#L668-L674
We can therefore just remove it.
### Changes
- replace `.none` with `.fixedSize(0)`
- remove `.none` from the definition of `RequestFramingMetadata.Body`


fabianfett (approved):
Great improvement

PR #481, opened 2021-11-19
Crash fix: HTTP2Connections emit events after the pool has closed them.

fabianfett:
On the swift-server slack a crash was reported:

```bash
2021-11-19T04:19:37.87233336Z 0x55a5a551c262, Backtrace.(printBacktrace in _B82A8C0ED7C904841114FDF244F9E58E)(signal: Swift.Int32) -> () at /build/.build/checkouts/swift-backtrace/Sources/Backtrace/Backtrace.swift:66
2021-11-19T04:19:37.872355121Z 0x7f702ab0d3bf
2021-11-19T04:19:37.872771454Z 0x55a5a53cdda6, Swift runtime failure: precondition failure at .build/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift:0
2021-11-19T04:19:37.872787184Z 0x55a5a53cdda6, generic specialization <Swift.Array<AsyncHTTPClient.HTTPConnectionPool.(HTTP2ConnectionState in _06AB5E1F2E49650D236C760F038E3F99)>> of (extension in Swift):Swift.Collection.firstIndex(where: (A.Element) throws -> Swift.Bool) throws -> Swift.Optional<A.Index> at .build/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift:528
2021-11-19T04:19:37.872823265Z 0x55a5a53cdda6, AsyncHTTPClient.HTTPConnectionPool.HTTP2Connections.goAwayReceived(Swift.Int) -> AsyncHTTPClient.HTTPConnectionPool.HTTP2Connections.GoAwayContext at .build/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift:527
2021-11-19T04:19:37.872830366Z 0x55a5a53cdda6, AsyncHTTPClient.HTTPConnectionPool.HTTP2StateMachine.http2ConnectionGoAwayReceived(Swift.Int) -> AsyncHTTPClient.HTTPConnectionPool.StateMachine.Action at .build/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift:301
[...]

PR #482, opened 2021-11-19
Make `http1Connections` and `http2Connections` non-optional

dnadoba:
Follow up from https://github.com/swift-server/async-http-client/pull/481#discussion_r753034593

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
@@ -537,6 +537,9 @@ extension HTTPConnectionPool {
             self.connections = self.connections.enumerated().compactMap { index, connectionState in
                 switch connectionState.migrateToHTTP2(&migrationContext) {
                 case .removeConnection:
+                    // If the connection has an index smaller than the previous overflow index,
+                    // we deal with a general purpose connection.
+                    // For this reason we need to decrement the overflow index.
                     if index < initialOverflowIndex {
                         self.overflowIndex = self.connections.index(before: self.overflowIndex)
                     }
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -532,12 +532,18 @@ extension HTTPConnectionPool {
...
-            self.connections.removeAll { connection in
-                switch connection.migrateToHTTP2(&migrationContext) {
+            let initialOverflowIndex = self.overflowIndex
+
+            self.connections = self.connections.enumerated().compactMap { index, connectionState in
+                switch connectionState.migrateToHTTP2(&migrationContext) {
                 case .removeConnection:
-                    return true
+                    if index < initialOverflowIndex {
+                        self.overflowIndex = self.connections.index(before: self.overflowIndex)
```
dnadoba, line 541: Previously we didn't update `self.overflowIndex` correctly which did hide some bugs because we rarely reused `HTTP1Connections` ever again after migration.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -41,17 +41,22 @@ extension HTTPConnectionPool {
...
+            maximumConcurrentHTTP1Connections: Int
         ) {
             self.idGenerator = idGenerator
             self.requests = RequestQueue()
 
             self.connections = HTTP2Connections(generator: idGenerator)
+            self.http1Connections = HTTP1Connections(
+                maximumConcurrentConnections: maximumConcurrentHTTP1Connections,
+                generator: idGenerator
+            )
```
fabianfett, line 54: I bet we reserve an array within this initializer. Maybe we want to add another initializer here, that really creates an empty HTTP1Connections. The struct is replaced in the migration anyway.

The constructor should set `maximumConcurrentConnections` to a default value. Then we don't need to pass it in the HTTP2 constructor.

fabianfett (commented):
Okay, I agree this makes the code nicer. Thanks for pushing for this. However we should go a little further.

dnadoba:
Delayed until https://github.com/swift-server/async-http-client/issues/487 is fixed.
Bug fix extracted into https://github.com/swift-server/async-http-client/pull/486

PR #483, opened 2021-11-19
[Tests] Add some basic IPv6 tests

karwa:
While working on the WebURL port, I noticed that there seemed to be no tests at all covering IPv6.

These are just 2 very simple ones, copies of `testGetHttpsWithIP` and `testGetHTTPSWorksOnMTELGWithIP` which use IPv6 rather than IPv4. It would be good to add more (in particular, I think it would be worth testing that the proxy's CONNECT request-target is properly formatted, but I'm not that familiar with the project so I'm unsure how to write proxy tests).

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -333,6 +333,40 @@ class HTTPClientTests: XCTestCase {
         XCTAssertEqual(.ok, response.status)
     }
 
+    func testGetHttpsWithIPv6() throws {
+        try XCTSkipUnless(canBindIPv6Loopback, "Requires IPv6")
+        let localHTTPBin = HTTPBin(.http1_1(ssl: true), bindTarget: .localhostIPv6RandomPort)
+        let localClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup),
+                                     configuration: HTTPClient.Configuration(certificateVerification: .none))
+        defer {
+            XCTAssertNoThrow(try localClient.syncShutdown())
+            XCTAssertNoThrow(try localHTTPBin.shutdown())
+        }
+        var response: HTTPClient.Response?
+        XCTAssertNoThrow(response = try localClient.get(url: "https://[::1]:\(localHTTPBin.port)/get").wait())
+        XCTAssertEqual(.ok, response?.status)
+    }
+
+    func testGetHTTPSWorksOnMTELGWithIPv6() throws {
+        try XCTSkipUnless(canBindIPv6Loopback, "Requires IPv6")
+        // Same test as above but this one will use NIO on Sockets even on Apple platforms, just to make sure
+        // this works.
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+        let localHTTPBin = HTTPBin(.http1_1(ssl: true), bindTarget: .localhostIPv6RandomPort)
+        let localClient = HTTPClient(eventLoopGroupProvider: .shared(group),
+                                     configuration: HTTPClient.Configuration(certificateVerification: .none))
+        defer {
+            XCTAssertNoThrow(try localClient.syncShutdown())
+            XCTAssertNoThrow(try localHTTPBin.shutdown())
+        }
+        var response: HTTPClient.Response?
+        XCTAssertNoThrow(response = try localClient.get(url: "https://[::1]:\(localHTTPBin.port)/get").wait())
+        XCTAssertEqual(.ok, response?.status)
+    }
+
     func testPostHttps() throws {
         let localHTTPBin = HTTPBin(.http1_1(ssl: true))
         let localClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup),
--- a/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
@@ -17,6 +17,7 @@ import Foundation
 import Logging
 import NIOConcurrencyHelpers
 import NIOCore
+import NIOEmbedded
 import NIOHPACK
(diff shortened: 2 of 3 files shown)


Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -333,6 +333,38 @@ class HTTPClientTests: XCTestCase {
...
+        let localHTTPBin = HTTPBin(.http1_1(ssl: true), bindTarget: .localhostIPv6RandomPort)
+        let localClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup),
+                                     configuration: HTTPClient.Configuration(certificateVerification: .none))
+        defer {
+            XCTAssertNoThrow(try localClient.syncShutdown())
+            XCTAssertNoThrow(try localHTTPBin.shutdown())
+        }
+
+        let response = try localClient.get(url: "https://[::1]:\(localHTTPBin.port)/get").wait()
+        XCTAssertEqual(.ok, response.status)
```
fabianfett, line 346:
```suggestion
        var response: HTTPClient.Response?
        XCTAssertNoThrow(response = try localClient.get(url: "https://[::1]:\(localHTTPBin.port)/get").wait())
        XCTAssertEqual(.ok, response?.status)
```

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -333,6 +333,38 @@ class HTTPClientTests: XCTestCase {
...
+        let localHTTPBin = HTTPBin(.http1_1(ssl: true), bindTarget: .localhostIPv6RandomPort)
+        let localClient = HTTPClient(eventLoopGroupProvider: .shared(group),
+                                     configuration: HTTPClient.Configuration(certificateVerification: .none))
+        defer {
+            XCTAssertNoThrow(try localClient.syncShutdown())
+            XCTAssertNoThrow(try localHTTPBin.shutdown())
+        }
+
+        let response = try localClient.get(url: "https://[::1]:\(localHTTPBin.port)/get").wait()
+        XCTAssertEqual(.ok, response.status)
```
fabianfett, line 365:
```suggestion
        var response: HTTPClient.Response?
        XCTAssertNoThrow(response = try localClient.get(url: "https://[::1]:\(localHTTPBin.port)/get").wait())
        XCTAssertEqual(.ok, response?.status)
```

fabianfett (commented):
Thanks @karwa! I've added two small code suggestions to the tests, that make the tests non throwing (which is how we prefer them).

karwa:
@fabianfett These tests are just copy-pasted from the 2 tests immediately above them (`testGetHttpsWithIP` and `testGetHTTPSWorksOnMTELGWithIP`). Perhaps it's worth updating them all (and similar tests in that file) in a follow-up?

fabianfett:
@karwa Yes, we are aware that not all current tests are in the form we would like them to be in. For this reason, a follow up pr would be very welcome.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -333,6 +333,38 @@ class HTTPClientTests: XCTestCase {
         XCTAssertEqual(.ok, response.status)
     }
 
+    func testGetHttpsWithIPv6() throws {
+        let localHTTPBin = HTTPBin(.http1_1(ssl: true), bindTarget: .localhostIPv6RandomPort)
```
Lukasa, line 337: These tests need to preflight that IPv6 loopback is available. Some systems do not configure the IPv6 module in Linux and so lack the ability to run these tests.

Usually the best way to do this is to just try binding and connecting to a random IPv6 port. We can hide that behind a `static let`.

karwa: I've added something - not sure if that's quite what you meant. I've added a NIOTS check to it, because the ServerBootstrap doesn't like the default ELG otherwise, and I figure we can rely on Network.framwork to support IPv6 (like... it should be an error if it doesn't work).

karwa:
I reworked the tests not to throw as @fabianfett suggested, but the preflight check for IPv6 is best modelled in XCTest by `XCSkip`, which marks a test "skipped" instead of passed/failed. That feature works by throwing, which means the test once again needs to be throwing.

Otherwise I could make the tests silently pass if we can' bind the IPv6 loopback, but I figure showing skipped adds significant value, so it's worth just accepting the `throws` here.

Lukasa (approved):

Lukasa:
A few minor formatting issues to clean up and then this is good to go:

```diff
diff --git a/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift b/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
index 0a61b10..b18c078 100644
--- a/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
@@ -51,9 +51,9 @@ let canBindIPv6Loopback: Bool = {
     let elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)
     defer { try! elg.syncShutdownGracefully() }
     let serverChannel = try? ServerBootstrap(group: elg)
-          .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
-          .bind(host: "::1", port: 0)
-          .wait()
+        .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+        .bind(host: "::1", port: 0)
+        .wait()
     let didBind = (serverChannel != nil)
     try! serverChannel?.close().wait()
     return didBind
```

karwa:
@Lukasa I've addressed the nit. For some reason, these are showing up as skipped in the bots - are they perhaps not configured with IPv6 support? The tests work (are not skipped) on my Mac.

Lukasa:
Yes, I suspect our CI is one of those environments where IPv6 doesn't work, which is probably what triggered that memory of this need.

PR #485, opened 2021-11-19
Refactor URL component extraction

karwa:
This is currently split across 4 functions (well, 3 functions and an initialiser). It's much easier to understand how AHC understands these URL formats if it happens in a single function (especially since schemes like `http+unix` are not standardised).

Additionally, this rejects empty HTTP hostnames, as they are invalid per the spec, and empty socket paths, since they are meaningless.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -108,54 +108,6 @@ extension HTTPClient {
             private static var hostRestrictedSchemes: Set = ["http", "https"]
             private static var allSupportedSchemes: Set = ["http", "https", "unix", "http+unix", "https+unix"]
 
-            init(forScheme scheme: String) throws {
-                switch scheme {
-                case "http", "https": self = .host
-                case "unix": self = .unixSocket(.baseURL)
-                case "http+unix": self = .unixSocket(.http_unix)
-                case "https+unix": self = .unixSocket(.https_unix)
-                default:
-                    throw HTTPClientError.unsupportedScheme(scheme)
-                }
-            }
-
-            func hostFromURL(_ url: URL) throws -> String {
-                switch self {
-                case .host:
-                    guard let host = url.host else {
-                        throw HTTPClientError.emptyHost
-                    }
-                    return host
-                case .unixSocket:
-                    return ""
-                }
-            }
-
-            func socketPathFromURL(_ url: URL) throws -> String {
-                switch self {
-                case .unixSocket(.baseURL):
-                    return url.baseURL?.path ?? url.path
-                case .unixSocket:
-                    guard let socketPath = url.host else {
-                        throw HTTPClientError.missingSocketPath
-                    }
-                    return socketPath
-                case .host:
-                    return ""
-                }
-            }
-
-            func uriFromURL(_ url: URL) -> String {
-                switch self {
-                case .host:
-                    return url.uri
-                case .unixSocket(.baseURL):
-                    return url.baseURL != nil ? url.uri : "/"
-                case .unixSocket:
-                    return url.uri
-                }
-            }
-
             func supportsRedirects(to scheme: String?) -> Bool {
                 guard let scheme = scheme?.lowercased() else { return false }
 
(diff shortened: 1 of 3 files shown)


dnadoba:

Lukasa (changes requested):
Yup, this makes sense, thanks. Do you mind very briefly adding tests that validate that empty socket paths are now rejected?

Lukasa:

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -267,12 +247,7 @@ extension HTTPClient {
             guard let scheme = url.scheme?.lowercased() else {
                 throw HTTPClientError.emptyScheme
             }
-
-            self.kind = try Kind(forScheme: scheme)
-            self.host = try self.kind.hostFromURL(url)
-            self.socketPath = try self.kind.socketPathFromURL(url)
-            self.uri = self.kind.uriFromURL(url)
-
+            (self.kind, self.host, self.socketPath, self.uri) = try Self.deconstructURL(url)
```
dnadoba, line 250: Great work! Thank you!

Would you mind returning the `scheme` and `port` in `deconstructURL(_:)` as well?
`scheme` is already a stored property in `HTTPClient.Request` but `port` not. 
https://github.com/swift-server/async-http-client/blob/5dfbba8bb569c0eb98b927d139e81644e344703e/Sources/AsyncHTTPClient/HTTPHandler.swift#L266-L268
However, I don't mind adding `port` as another stored property. This would allow us to better reuse this functionality in the upcoming new `HTTPClientRequest` for async/await support as described in this proposal: 
https://github.com/swift-server/async-http-client/blob/96872470475f18a5667fbf681e2e0f67ad8a0110/docs/async-await.md#new-httpclientrequest-type

dnadoba: Friendly Ping 🙂
I would like to use your changes soon. Do you have any estimate when you have time to work on this again?

dnadoba: @karwa friendly ping 🙂
I would like to use your changes soon. Do you have any estimate when you have time to work on this again?

karwa: Sure, I'll address the comments now and update the patch

karwa:
OK, I've added the test as @Lukasa requested, and `deconstructURL` now extracts the scheme and port number as @dnadoba requested.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -108,63 +108,52 @@ extension HTTPClient {
...
+                return (.unixSocket(kind), scheme, "", url.port ?? defaultPort, socketPath, url.uri)
+            case "unix":
+                let socketPath = url.baseURL?.path ?? url.path
+                let uri = url.baseURL != nil ? url.uri : "/"
+                guard !socketPath.isEmpty else {
+                    throw HTTPClientError.missingSocketPath
                 }
+                return (.unixSocket(.baseURL), scheme, "", url.port ?? 80, socketPath, uri)
+            default:
+                throw HTTPClientError.unsupportedScheme(url.scheme!)
```
dnadoba, line 156:
```suggestion
                throw HTTPClientError.unsupportedScheme(scheme)
```
We have a non-optional scheme already in scope, no need to force unwrap.

dnadoba (changes requested):
Thanks a lot! One small change and we are good to go from my side. 
CI is currently flaky because I turned parallel testing on. Fix is on its way (https://github.com/swift-server/async-http-client/pull/498) so don't worry if you see any test in `HTTPClientSOCKSTests` failing.

Sources/AsyncHTTPClient/HTTPHandler.swift
karwa: The non-optional scheme is lowercased, and I figured it would preserve existing behaviour better to retain the original scheme verbatim. But I don't mind changing it if you think that isn't worth doing.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -215,6 +215,14 @@ class HTTPClientTests: XCTestCase {
         XCTAssertNil(url10)
     }
 
+    func testBadUnixWithBaseURL() {
+        let badUnixBaseURL = URL(string: "/foo", relativeTo: URL(string: "unix:")!)!
+        XCTAssertEqual(badUnixBaseURL.baseURL?.path, "")
+        XCTAssertThrowsError(try Request(url: badUnixBaseURL), "should throw") { error in
```
fabianfett, line 221: Smallest nit ever seen: The message doesn't add any information.
```suggestion
        XCTAssertThrowsError(try Request(url: badUnixBaseURL)) { error in
```

fabianfett (approved):
This is a great change! Thanks!

Sources/AsyncHTTPClient/HTTPHandler.swift
dnadoba: Makes sense, let's keep it this way.

dnadoba (approved):

Lukasa (approved):

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
karwa: Sure, this is how the ones in `testBadRequestURI` are, but you're right that it doesn't add anything.

PR #490, opened 2021-11-24
Fix `bodyLengthMissmatch` error handling

dnadoba:
### Motivation
If we detect that a user tries to write more bytes as initially specified, we fail the request with a `HTTPClientError. bodyLengthMissmatch`. However, we forgot to update our internal state that the request is failed. One way this manifests is that subsequent errors are also propagated through the stack but should be silently ignored because the request is already failed.

In addition, two compound cases in the switch statement inside of `errorHappened(_:)` had a where clause which did only apply to the last case but should have applied to each compound cases. This affects the `bodyLengthMissmatch` error handling but also all error handling because errors were sometimes treated as `NIOSSLError.uncleanShutdown` even if they were something completely different.

### Changes
- set state to failed if we fail the request with a `.bodyLengthMissmatch` error
- duplicate where clause too each compound cases to only enter the cases body if the error is `NIOSSLError.uncleanShutdown`

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPRequestStateMachineTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPRequestStateMachineTests.swift
@@ -209,7 +209,10 @@ class HTTPRequestStateMachineTests: XCTestCase {
 
         let part1 = IOData.byteBuffer(ByteBuffer(bytes: 4...7))
         XCTAssertEqual(state.requestStreamPartReceived(part1), .sendBodyPart(part1))
-        XCTAssertEqual(state.requestStreamFinished(), .failRequest(HTTPClientError.bodyLengthMismatch, .close))
+        guard case .failRequest(let error, .close) = state.requestStreamFinished() else {
+            return XCTFail("unexpected action")
+        }
+        XCTAssertEqual(error as? HTTPClientError, .bodyLengthMismatch)
         XCTAssertEqual(state.channelInactive(), .wait)
     }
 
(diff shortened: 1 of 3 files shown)


Tests/AsyncHTTPClientTests/HTTPRequestStateMachineTests.swift
```diff
@@ -522,6 +525,27 @@ class HTTPRequestStateMachineTests: XCTestCase {
...
+    }
+
+    func testArbitraryErrorShouldBeTreatedAsARequestFailureWhileInWaitingForHeadState() {
+        struct ArbitraryError: Error {}
+        var state = HTTPRequestStateMachine(isChannelWritable: true)
+        let requestHead = HTTPRequestHead(version: .http1_1, method: .GET, uri: "/")
+        let metadata = RequestFramingMetadata(connectionClose: false, body: .fixedSize(0))
+        XCTAssertEqual(state.startRequest(head: requestHead, metadata: metadata), .sendRequestHead(requestHead, startBody: false))
+
+        XCTAssertEqual(state.errorHappened(ArbitraryError()), .failRequest(ArbitraryError(), .close))
```
fabianfett, line 545: The error is not included in the Equatable conformance. To make sure you will need to unwrap the enum with a guard and make the error equatable.

dnadoba: okay... but we have 18 places where we need to fix this.

fabianfett: doing it correctly here is a start ;)

Tests/AsyncHTTPClientTests/HTTPRequestStateMachineTests.swift
```diff
@@ -373,15 +391,21 @@ class HTTPRequestStateMachineTests: XCTestCase {
...
     func testRemoteSuddenlyClosesTheConnection() {
         var state = HTTPRequestStateMachine(isChannelWritable: true)
         let requestHead = HTTPRequestHead(version: .http1_1, method: .GET, uri: "/", headers: .init([("content-length", "4")]))
         let metadata = RequestFramingMetadata(connectionClose: false, body: .fixedSize(4))
         XCTAssertEqual(state.startRequest(head: requestHead, metadata: metadata), .sendRequestHead(requestHead, startBody: true))
-        XCTAssertEqual(state.requestCancelled(), .failRequest(HTTPClientError.remoteConnectionClosed, .close))
+        guard case .failRequest(let error, .close) = state.requestCancelled() else {
+            return XCTFail("unexpected action")
+        }
+        XCTAssertEqual(error as? HTTPClientError, .cancelled)
```
dnadoba, line 408: Previously we checked for `.remoteConnectionClosed` but that was probably wrong and I think we actually  wanted to check for `.cancel`.

fabianfett (approved):
Great work! Thanks!

PR #491, opened 2021-11-25
Update swiftformat to 0.48.8

fabianfett:
### Motivation

Our current swiftformat version does not support async/await. Since we want to add support for async/await we must update swiftformat or disable it. I tried my very best to keep the number of changes as small as possible. I assume we want to stick with the new 0.48.8 for some time.

### Changes

- Update swiftformat to 0.48.8

### Result

We can land async/await code.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,14 +1,23 @@
 # file options
 
---swiftversion 5.0
+--swiftversion 5.2
 --exclude .build
 
 # format options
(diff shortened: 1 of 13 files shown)


Lukasa (approved):
Not too bad, and we don't have a choice.

dnadoba (approved):
LGTM!

PR #494, opened 2021-11-25
SwiftFormat `--ifdef no-indent`

dnadoba:
### Motivation
We do not want code indention for compile-time conditions because will have a lot of them if we add async/await support.
Literally all new code for async/await needs to be behind `#if compiler(>=5.5)` which would indent all code by 4 spaces.
### Changes
- set `--ifdef` to `no-indent` and run SwiftFormat
- update `generate_linux_tests.rb ` to use the new indention rule as well


glbrntt (approved):
Looks good modulo sanity check failing.

dnadoba:
> Looks good modulo sanity check failing.

Ah yeah. I needed to update `generate_linux_tests.rb ` to use the new indention rule too.

PR #495, opened 2021-11-25
Run tests in parallel on CI

dnadoba:
### Motivation
I locally use parallel and randomised testing in Xcode since two months and did not yet encounter any flaky tests. I therefore think we can run tests in parallel on CI as well.

### Changes
add `--parallel` to `swift test` invocation

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -28,7 +28,7 @@ services:
 
   test:
     <<: *common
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

Lukasa:
It's good hygiene, if our tests _can_ run in parallel, for us to do it. :+1: Great work!

dnadoba:
It also cuts our CI times in half from 8-10 minutes down to 4-5 minutes 🥳

PR #496, opened 2021-11-25
[DRAFT] Use test discovery instead of generated `allTests` on linux 

dnadoba:
This is an experiment to see if all test cases are discovered on all supported swift version on CI using test discovery. This might be an option to support async/await tests if we think modifying the `generate_linux_tests.rb` is too complicated.
We may still want to use generated test cases so we can write tests for deprecated symbols without warnings in tests.


dnadoba:
All Swift versions report the same amount of tests executed and skipped before and after the change.
```
Executed 375 tests, with 2 tests skipped and 0 failures (0 unexpected)
```
I therefore assume that all tests are discovered.

fabianfett:
The Problem is if we deprecate an API and want to test the deprecated API, we need to deprecate the test as well. Deprecated test sadly emit a warning when auto discovered. Since we build with warnings as errors, we have an issue here.

fabianfett:
For reference: https://bugs.swift.org/browse/SR-12008

PR #501, opened 2021-11-26
Introduce a `ConnectionTarget` enum

karwa:
I added this as part of the WebURL port, because it can provide specific host kinds and doesn't need to re-parse the hostname, but I think it is generally a useful thing to model.

Currently, `ConnectionPool.Key` stores a `host`, `port`, and `socketPath` string, even though only one of (host + port) and (socketPath) can be populated. This lends itself well to an enum. Additionally, we can extract what kind of host we're dealing with, so we don't need to re-parse the hostname later.

One thing I'd like comments on is whether this should perhaps be moved up to `Request`. This would also allows us to stop pretending that unix domain sockets have ports. As I see it, this should really be `Request`'s business; a part of how it extracts information from the URL should be parsing the hostname string (perhaps using APIs from the URL type, if it provides that). 

The async redesign simplifies `Request` somewhat, but I couldn't see any mention of hosts or hostnames on the new request type. One idea worth exploring would be making this public API as part of that redesign.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool.swift
@@ -12,63 +12,30 @@
 //
 //===----------------------------------------------------------------------===//
 
-import enum NIOCore.SocketAddress
-
 enum ConnectionPool {
-    enum Host: Hashable {
-        // We keep the IP address serialization precisely as it is in the URL.
-        // Some platforms have quirks in their implementations of 'ntop', for example
-        // writing IPv6 addresses as having embedded IPv4 sections (e.g. [::192.168.0.1] vs [::c0a8:1]).
-        // This serialization includes square brackets, so it is safe to write next to a port number.
-        // Note: `address` must always have a non-nil port.
-        case ipAddress(serialization: String, address: SocketAddress)
-        case domain(name: String, port: Int)
-        case unixSocket(path: String)
-
-        init(remoteHost: String, port: Int) {
-            if let addr = try? SocketAddress(ipAddress: remoteHost, port: port) {
-                switch addr {
-                case .v6:
-                    self = .ipAddress(serialization: "[\(remoteHost)]", address: addr)
-                case .v4:
-                    self = .ipAddress(serialization: remoteHost, address: addr)
-                case .unixDomainSocket:
-                    fatalError("Expected a remote host")
-                }
-            } else {
-                precondition(!remoteHost.isEmpty, "HTTPClient.Request should already reject empty remote hostnames")
-                self = .domain(name: remoteHost, port: port)
-            }
-        }
-    }
-
     /// Used by the `ConnectionPool` to index its `HTTP1ConnectionProvider`s
     ///
-    /// A key is initialized from a `URL`, it uses the components to derive a hashed value
+    /// A key is initialized from a `Request`, it uses the components to derive a hashed value
     /// used by the `providers` dictionary to allow retrieving and creating
     /// connection providers associated to a certain request in constant time.
     struct Key: Hashable, CustomStringConvertible {
         var scheme: Scheme
-        var host: Host
+        var connectionTarget: ConnectionTarget
         private var tlsConfiguration: BestEffortHashableTLSConfiguration?
 
         init(_ request: HTTPClient.Request) {
+            self.connectionTarget = request.connectionTarget
             switch request.scheme {
             case "http":
                 self.scheme = .http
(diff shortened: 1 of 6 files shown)


karwa:
OK, I just did it and moved it up in to `Request`. I couldn't think of a good reason not to.

karwa:
One idea I'd like to float is potentially naming this `ConnectionTarget`, since I'm not sure unix sockets really count as "hosts" in the typical sense (`Request.host` currently returns an empty string for them, even though HTTP requests to empty hosts are technically invalid). 

Also, it would avoid overloading the term "host" too much. It already clashes with `Request.host`, meaning it requiring an underscored or other suboptimal name.

Lukasa:
`ConnectionTarget` is a good name, I'm in favour of that amendment.

karwa:
OK, I've changed it to `ConnectionTarget`, and added tests for the quirky embedded IPv4 addresses.

dnadoba:
Did a quick read through the changes and it is a nice cleanup. Thanks! Will do a proper review on monday.
I'm also refactoring some related stuff like `scheme` and moving the destructured URL into a separate type. If you are interested you can have a look at this branch https://github.com/dnadoba/async-http-client/tree/dn-scheme 
Still work in progress and naming is not final.
Nevertheless, I think we should land your changes first.

Sources/AsyncHTTPClient/ConnectionPool.swift
```diff
@@ -15,11 +15,16 @@
 enum ConnectionPool {
     /// Used by the `ConnectionPool` to index its `HTTP1ConnectionProvider`s
     ///
-    /// A key is initialized from a `URL`, it uses the components to derive a hashed value
+    /// A key is initialized from a `Request`, it uses the components to derive a hashed value
     /// used by the `providers` dictionary to allow retrieving and creating
     /// connection providers associated to a certain request in constant time.
     struct Key: Hashable, CustomStringConvertible {
+        var scheme: Scheme
+        var connectionTarget: ConnectionTarget
```
dnadoba, line 23: FYI: For now its probably okay to use `ConnectionTarget` here and it also does not change the behaviour compared to the old implementation. However, I think in the future we should probably use a different type that normalises the `ConnectionTarget`. E.g. the host should be compare case insensitive and different spellings of the same IP address should also result in the same `ConnectionPool.Key`.  Otherwise we end up with two different connections if a user wants to connect to `https://apple.com` and `https://APPLE.com` which we do not want.

dnadoba (approved):
Great patch! Looks good to me modulo one nit. We should still wait for @fabianfett review until we merge it.

w.r.t.:
> The async redesign simplifies Request somewhat, but I couldn't see any mention of hosts or hostnames on the new request type. One idea worth exploring would be making this public API as part of that redesign.

The new Request type, called `HTTPClientRequest`, does not do any parsing or validation. It also just accepts a url as a `String`. Parsing, destructering and validation of the url is all delayed until execution. The advantage is that the initialiser of `HTTPClientRequest` is non-throwing and all properties are `var`s. Therefore we will not even have `ConnectionTarget` as an internal property on `HTTPClientRequest`.

My plan is to introduce a new internal type called `PreparedRequest` which is created as part of executing the request. This type will be the new home for all properties which are computed as part of the current initialiser of `HTTPClient.Request` and the validation step which is currently also executed before sending out a `HTTPClient.Request`.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -45,6 +45,30 @@ final class HTTP1ProxyConnectHandler: ChannelDuplexHandler, RemovableChannelHand
         return self.proxyEstablishedPromise?.futureResult
     }
 
+    convenience
+    init(target: ConnectionTarget,
```
Lukasa, line 49: Minor nit: I'd like to put `convenience init` on the same line.

Lukasa (approved):
One nit, and I'd like to see @dnadoba's nits addressed, but otherwise this is great. Nice patch @karwa!

fabianfett (approved):

dnadoba:
@karwa I merge it now and fix the nits in a follow up PR quickly because I want to use them in another PR. Thanks again for this patch!

karwa:
Thanks @dnadoba! I was planning to fix in the evening yesterday, but was too tired 😴

PR #504, opened 2021-11-30
Refactor `deconstructURL ` and `scheme` parsing

dnadoba:
### Motivation
We want to reuse our URL parsing, validation and deconstruction logic for the upcoming new `HTTPClientRequest` from the async/await proposal.

### Changes
- Introduce a new `DeconstructedURL` struct which hold a `scheme`, `connectionTarget` and `uri`.
- Introduce a new `DeconstructedURL.Scheme` enum which replaced the previous `String` in `HTTPClient.Request` and `ConnectionPool.Scheme` enum
- move `deconstructURL` logic to an initialiser of  `DeconstructedURL`
- use `DeconstructedURL` instead of separate `scheme`, `connectionTarget`, and `uri` properties in `HTTPClient.Request`
- remove `HTTPClient.Request.Kind` because it is almost a replica of the `DeconstructedURL.Scheme` enum where the only difference is that `http` and `https` are represented as one case called `host`. As we now represent `scheme` as an enum instead of a `String` we can safely replace it with an exhaustive switch over `scheme` instead.
- refactor tests to use `scheme` instead of `kind`
- remove unused `config(overriding:)` from `ConnectionPool.Key`

### DIFF
--- a/Sources/AsyncHTTPClient/DeconstructedURL.swift
+++ b/Sources/AsyncHTTPClient/DeconstructedURL.swift
@@ -22,10 +22,11 @@ struct DeconstructedURL {
         case httpUnix = "http+unix"
         case httpsUnix = "https+unix"
     }
+
     let scheme: Scheme
     let connectionTarget: ConnectionTarget
     let uri: String
-    
+
     internal init(
         scheme: DeconstructedURL.Scheme,
         connectionTarget: ConnectionTarget,
@@ -45,7 +46,7 @@ extension DeconstructedURL {
         guard let scheme = Scheme(rawValue: schemeString.lowercased()) else {
             throw HTTPClientError.unsupportedScheme(schemeString)
         }
-        
+
         switch scheme {
         case .http, .https:
             guard let host = url.host, !host.isEmpty else {
@@ -56,7 +57,7 @@ extension DeconstructedURL {
                 connectionTarget: .init(remoteHost: host, port: url.port ?? scheme.defaultPort),
                 uri: url.uri
             )
-            
+
         case .httpUnix, .httpsUnix:
             guard let socketPath = url.host, !socketPath.isEmpty else {
                 throw HTTPClientError.missingSocketPath
@@ -66,7 +67,7 @@ extension DeconstructedURL {
                 connectionTarget: .unixSocket(path: socketPath),
                 uri: url.uri
             )
-            
+
         case .unix:
             let socketPath = url.baseURL?.path ?? url.path
             let uri = url.baseURL != nil ? url.uri : "/"
@@ -91,7 +92,8 @@ extension DeconstructedURL.Scheme {
             return true
         }
     }
+
     var defaultPort: Int {
-        useTLS ? 443 : 80
+        self.useTLS ? 443 : 80
     }
 }
--- a/Sources/AsyncHTTPClient/DeconstructedURL.swift
+++ b/Sources/AsyncHTTPClient/DeconstructedURL.swift
@@ -14,7 +14,7 @@
 
 import Foundation
 
-struct Endpoint {
+struct DeconstructedURL {
     enum Scheme: String {
         case http
         case https
@@ -27,7 +27,7 @@ struct Endpoint {
     let uri: String
     
     internal init(
-        scheme: Endpoint.Scheme,
+        scheme: DeconstructedURL.Scheme,
         connectionTarget: ConnectionTarget,
         uri: String
     ) {
@@ -37,7 +37,7 @@ struct Endpoint {
     }
 }
 
-extension Endpoint {
+extension DeconstructedURL {
     init(url: URL) throws {
         guard let schemeString = url.scheme else {
             throw HTTPClientError.emptyScheme
@@ -51,7 +51,6 @@ extension Endpoint {
             guard let host = url.host, !host.isEmpty else {
                 throw HTTPClientError.emptyHost
             }
-            
             self.init(
                 scheme: scheme,
                 connectionTarget: .init(remoteHost: host, port: url.port ?? scheme.defaultPort),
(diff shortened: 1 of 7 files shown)


dnadoba:

Sources/AsyncHTTPClient/DeconstructedURL.swift
```diff
@@ -0,0 +1,99 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
fabianfett, line 15: NIT: Only import what is actually needed.
```suggestion
import struct Foundation.URL
```

Sources/AsyncHTTPClient/DeconstructedURL.swift
```diff
@@ -0,0 +1,99 @@
...
+                scheme: scheme,
+                connectionTarget: .unixSocket(path: socketPath),
+                uri: uri
+            )
+        }
+    }
+}
+
+extension DeconstructedURL.Scheme {
+    var useTLS: Bool {
```
fabianfett, line 87: He, she, it das s muss mit ;)

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -277,25 +209,28 @@ extension HTTPClient {
...
+            switch self.deconstructedURL.connectionTarget {
             case .ipAddress(_, let address): return address.port!
             case .domain(_, let port): return port
-            case .unixSocket: return Request.useTLS(self.scheme) ? 443 : 80
+            case .unixSocket: return self.deconstructedURL.scheme.defaultPort
             }
         }
 
+        /// Whether request will be executed using secure socket.
+        public var useTLS: Bool { self.deconstructedURL.scheme.useTLS }
```
fabianfett, line 220: I think `uses` is better.

fabianfett: Also, why do we add this public API.

Sources/AsyncHTTPClient/DeconstructedURL.swift
```diff
@@ -0,0 +1,99 @@
...
+import Foundation
+
+struct DeconstructedURL {
+    enum Scheme: String {
+        case http
+        case https
+        case unix
+        case httpUnix = "http+unix"
+        case httpsUnix = "https+unix"
+    }
```
fabianfett, line 24: I'm not sure this should be namespaced in DeconstructedURL. Would you mind moving this out?

fabianfett (commented):
All in all looks good. Few changes wanted.

Sources/AsyncHTTPClient/HTTPHandler.swift
dnadoba: It’s already public api and we can’t change the naming

fabianfett: Oh I see. Makes sense!

Sources/AsyncHTTPClient/DeconstructedURL.swift
dnadoba: We will diverge from the public API on HTTPClient.Request but I’m okay with that

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -748,7 +748,7 @@ internal struct RedirectHandler<ResponseType> {
     }
 }
 
-extension DeconstructedURL.Scheme {
+extension Scheme {
```
fabianfett, line 751: I think this should live in the Scheme file.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -92,82 +92,19 @@ extension HTTPClient {
...
-        }
-
         /// Request HTTP method, defaults to `GET`.
         public let method: HTTPMethod
         /// Remote URL.
         public let url: URL
+
+        /// Parsed, validated and deconstructed URL.
+        internal let deconstructedURL: DeconstructedURL
+
```
fabianfett, line 102: NIT: Move down to other internal var like redirect state. Mixing internal and public makes this hard to read.

fabianfett (approved):
Very cool, except for the tiny nits.

fabianfett (approved):

fabianfett:

karwa:
One thing that might be worth considering is if the scheme can just be dropped?

- `http` == `httpUnix` == `unix`, the only difference is the connection target (host/UDS)
- `https` == `httpsUnix`, the only difference is the connection target (host/UDS)

Because it now knows which kind of connection target it has, the only thing the scheme provides is whether to use TLS.

dnadoba:
Sadly we expose the scheme as public API. Not as the `Scheme` enum but as a `String` on `HTTPClient.Request` and therefore need to preserve it for the time being.

karwa:
Technically, it could get that from the URL. There's no need for ConnectionPool or most other things to know specifically which kind of scheme was used (also, `http+unix` and `unix` schemes should resolve to the same connection in the pool if they point to the same socket).

Lukasa:
> (also, http+unix and unix schemes should resolve to the same connection in the pool if they point to the same socket).

I disagree with this: the scheme is meaningfully part of both the origin and the network key for a request, and so should use different connections.

fabianfett:
The problem here are http semantics. From an http standpoint `unix:///some/path` and `http+unix:///some/path` go to different places. While the server _might_ be the same, we don't know for sure. And since connections are trust boundaries, we need to make the differentiation based on the scheme.

karwa:
I'm not sure to what extent HTTP really intends to define the semantics of non-`http:`/`https:` URLs. It sounds more likely that the semantics of application-defined schemes are also application-defined.

Lukasa:
HTTP does not define them. I agree that in principle we can define them however we want. Given that schemes are explicitly part of the connection pooling pattern in fetch, the most prudent choice is to generalise that idea to these schemes, especially as there are very few practical problems caused by doing so.

PR #509, opened 2021-12-01
Add `HTTPClientRequest`

dnadoba:
`HTTPClientRequest` and `HTTPClientRequest.Body` as described in the [async/await proposal](https://github.com/swift-server/async-http-client/blob/96872470475f18a5667fbf681e2e0f67ad8a0110/docs/async-await.md) with some modifications:
- `.bytes(_:)` and `.stream(_:)` from `HTTPClientRequest.Body`  accept an additional optional `length` parameter to specify the `length` in advance. This will primarily be used to set a `Content-Length` header.
- new overload for  `HTTPClientRequest.Body.bytes(..)` that accepts a `RandomAccessCollection` which automatically uses the `count` property from the given collection as the body length.

Note that everything is still internal. This allows us to merge async/await components incrementally while still being able to release new versions from main.

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
@@ -1,0 +1,34 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+import NIOHTTP1
+
+@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+struct HTTPClientRequest {
+    var url: String
+    var method: HTTPMethod
+    var headers: HTTPHeaders
+
+    var body: Body?
+
+    init(url: String) {
+        self.url = url
+        self.method = .GET
+        self.headers = .init()
+        self.body = .none
+    }
+}
+
+#endif
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Body.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Body.swift
@@ -1,0 +1,110 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+import NIOCore
+
+@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+extension HTTPClientRequest {
+    struct Body {
+        internal enum Mode {
+            case asyncSequence(length: Int?, (ByteBufferAllocator) async throws -> ByteBuffer?)
+            case sequence(length: Int?, (ByteBufferAllocator) -> ByteBuffer)
+            case byteBuffer(ByteBuffer)
+        }
+
+        var mode: Mode
+
+        private init(_ mode: Mode) {
+            self.mode = mode
+        }
+    }
+}
+
+@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+extension HTTPClientRequest.Body {
+    static func byteBuffer(_ byteBuffer: ByteBuffer) -> Self {
+        self.init(.byteBuffer(byteBuffer))
+    }
+
+    static func bytes<Bytes>(
+        length: Int? = nil,
+        _ bytes: Bytes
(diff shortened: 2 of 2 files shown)


Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -31,4 +32,96 @@ struct HTTPClientRequest {
...
+    /// The whole purpose of this overload is to prevent users from providing a redundant length if `Bytes` conforms to
+    /// `RandomAccessCollection` because it already provide a property `count` to get the length in O(**1**).
+    /// - Note: `length` is ignored in favour of `bytes.count`
+    @available(*, deprecated, message: "no need to manually specify `length` because we automatically use `bytes.count` as the `length`")
+    static func bytes<Bytes>(
+        length: Int,
+        _ collection: Bytes
+    ) -> Self where Bytes: RandomAccessCollection, Bytes.Element == UInt8 {
+        return .bytes(collection)
+    }
```
fabianfett, line 95: This is an interesting idea... It pushes people into using the right thing. However I wonder if it is a good idea to introduce new API, that we deprecate from the get go. @Lukasa wdyt?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -0,0 +1,127 @@
...
+}
+
+@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+extension HTTPClientRequest.Body {
+    static func byteBuffer(_ byteBuffer: ByteBuffer) -> Self {
+        self.init(.byteBuffer(byteBuffer))
+    }
+
+    static func bytes<Bytes>(
```
Lukasa, line 58: Can we make this method `@inlinable`?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -0,0 +1,127 @@
...
+            if let buffer = bytes.withContiguousStorageIfAvailable({ allocator.buffer(bytes: $0) }) {
+                // fastpath
+                return buffer
+            }
+            // potentially really slow path
+            return allocator.buffer(bytes: bytes)
+        })
+    }
+
+    static func bytes<Bytes>(
```
Lukasa, line 72: Can we make this method `@inlinable`?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -0,0 +1,127 @@
...
+            return allocator.buffer(bytes: bytes)
+        })
+    }
+
+    /// This method should never be used and was always deprecated.
+    /// The whole purpose of this overload is to prevent users from providing a redundant length if `Bytes` conforms to
+    /// `RandomAccessCollection` because it already provide a property `count` to get the length in O(**1**).
+    /// - Note: `length` is ignored in favour of `bytes.count`
+    @available(*, deprecated, message: "no need to manually specify `length` because we automatically use `bytes.count` as the `length`")
+    static func bytes<Bytes>(
```
Lukasa, line 90: Can we make this method `@inlinable`?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -0,0 +1,127 @@
...
+        length: Int,
+        _ collection: Bytes
+    ) -> Self where Bytes: RandomAccessCollection, Bytes.Element == UInt8 {
+        return .bytes(collection)
```
Lukasa, line 94: If we're going to offer this overload, can we at least `precondition` that the length matches the count of the collection?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
Lukasa: I'd be more inclined to consider marking the method `@available(*, unavailable)` and then not giving it an implementation beyond `fatalError`

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    static func stream<SequenceOfBytes>(
```
Lukasa, line 97: Can we make this method `@inlinable`?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -0,0 +1,127 @@
...
+        _ sequenceOfBytes: SequenceOfBytes
+    ) -> Self where SequenceOfBytes: AsyncSequence, SequenceOfBytes.Element == ByteBuffer {
+        var iterator = sequenceOfBytes.makeAsyncIterator()
+        let body = self.init(.asyncSequence(length: length) { _ -> ByteBuffer? in
+            try await iterator.next()
+        })
+        return body
+    }
+
+    static func stream<Bytes>(
```
Lukasa, line 108: Can we make this method `@inlinable`?

Lukasa (commented):
I've left some notes in the diff. Are we adding tests in a separate PR?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
dnadoba: I have tried `@available(*, unavailable)` but then the swift compiler will just choose the overload for Sequence:
https://github.com/swift-server/async-http-client/blob/8e4b2b4267029978821737c42e920dfd6b2fe4d3/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift#L58-L70
Even adding `@_disfavoredOverload` to this overload doesn't help.

I think what we could also do is marking the method as deprecated and just adding a fatalError with a useful message.

dnadoba:
> Are we adding tests in a separate PR?
Yes

fabianfett (approved):
LGTM. Thanks!

Lukasa (approved):

PR #510, opened 2021-12-02
Redo HTTP cookie parsing

karwa:
This PR does a couple of things:

- Use `strptime` for time parsing.

  This is what [Apple's documentation recommends](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/DataFormatting/Articles/dfDateFormatting10_4.html#//apple_ref/doc/uid/TP40002369-SW12) for fixed, locale-independent formats. Unfortunately, Glibc needs us to define the `_GNU_SOURCE` flag, which requires the use of a C shim module.

  Besides URL (which is debatable as to how "required" it is - you can make requests without URLs), this is the only "required" functionality which depends on Foundation.

- Rewrite cookie parsing

  The previous implementation had more allocations and copies than were necessary, and some really subtle uses of Foundation APIs such as `.components(separatedBy: ";")`. I've replaced it with something that tries to keep things as Substrings as much as possible and uses only the standard library.

- Move Foundation APIs on HTTPClient.Cookie to a different file

  It turns out, there are really not many places where AsyncHTTPClient uses Foundation types in public APIs. There are a couple of initializers on Request which take a URL, Cookie has this Date, and Body has a convenience for working with Data. But that's it. By putting these APIs in a separate file, it will be easier to make them opt-out for no-foundation builds.


karwa:
Hmm, seems like some older versions of Swift won't import the function using the header shim. We'll need an actual C function to forward to `strptime`.

Package.swift
```diff
@@ -29,9 +29,11 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.0"),
     ],
     targets: [
+        .target(name: "CShims"),
```
Lukasa, line 32: Given that target names have an annoyingly profound uniqueness requirement, can we namespace this more thoroughly? `CAsyncHTTPClient` would be best.

Sources/AsyncHTTPClient/FoundationExtensions.swift
```diff
@@ -0,0 +1,49 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2018-2021 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/CShims/include/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module CShims {
```
Lukasa, line 1: Using explicit module maps breaks some older Xcode functionality in fun ways, but we don't need it. If you change the name of the `shims.h` header to be exactly the same as the module name (in this case, I've proposed `CAsyncHTTPClient`, which would give you the header file `CAsyncHTTPClient.h`, then SwiftPM will autogenerate the module map as needed.

Sources/CShims/include/shims.h
```diff
@@ -0,0 +1,33 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/CShims/shims.c
```diff
@@ -0,0 +1,31 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```

Tests/AsyncHTTPClientTests/HTTPClientCookieTests.swift
```diff
@@ -84,25 +94,49 @@ class HTTPClientCookieTests: XCTestCase {
...
+        XCTAssertEqual("/path", c?.path)
 
-        c = HTTPClient.Cookie(header: "key=\"\"", defaultDomain: "example.org")!
-        XCTAssertEqual("", c.value)
+        c = HTTPClient.Cookie(header: "key=\"\"", defaultDomain: "example.org")
+        XCTAssertEqual("", c?.value)
+    }
+}
+
+import CShims
```
Lukasa, line 120: Can we move this import to the top?

Tests/AsyncHTTPClientTests/HTTPClientCookieTests.swift
```diff
@@ -84,25 +94,49 @@ class HTTPClientCookieTests: XCTestCase {
...
+
+extension HTTPClientCookieTests {
```
Lukasa, line 122: No need for the extension, we can just write the test inline.

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -59,6 +64,19 @@ let canBindIPv6Loopback: Bool = {
     return didBind
 }()
 
+/// Runs the given block in the context of a non-English C locale (in this case, German).
+/// Throws an XCTSkip error if the locale cannot be set.
+func withCLocaleSetToGerman(_ body: () throws -> Void) throws {
+    guard let germanLocale = newlocale(LC_TIME_MASK | LC_NUMERIC_MASK, "de_DE", nil) else {
+        throw XCTSkip("Unable to create locale de_DE")
```
Lukasa, line 71: This is aggressive: I think we only want to skip the test if errno was `ENOENT`. Otherwise the error is in our code and we should know about it by `XCTFail`.

Sources/AsyncHTTPClient/FoundationExtensions.swift
```diff
@@ -0,0 +1,49 @@
...
+//===----------------------------------------------------------------------===//
+
+// Extensions which provide better ergonomics when using Foundation types,
+// or by using Foundation APIs.
+
+import Foundation
+
+extension HTTPClient.Cookie {
+    /// The cookie's expiration date.
+    public var expires: Date? {
```
Lukasa, line 22: This needs a `set` as well, to match the original API.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -42,38 +41,29 @@ extension HTTPClient {
         ///     - defaultDomain: Default domain to use if cookie was sent without one.
         /// - returns: nil if the header is invalid.
         public init?(header: String, defaultDomain: String) {
-            let components = header.components(separatedBy: ";").map {
-                $0.trimmingCharacters(in: .whitespaces)
-            }
+            let components = header.split(separator: ";", omittingEmptySubsequences: false)
```
Lukasa, line 44: This will perform _vastly_ better if we do this work on the `UTF8View`. Is there any reason we can't?

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,111 @@ extension HTTPClient {
...
+extension HTTPClient.Response {
+    /// List of HTTP cookies returned by the server.
+    public var cookies: [HTTPClient.Cookie] {
+        return self.headers["set-cookie"].compactMap { HTTPClient.Cookie(header: $0, defaultDomain: self.host) }
     }
 }
 
-extension String {
-    fileprivate func omittingQuotes() -> String {
+extension StringProtocol {
```
Lukasa, line 121: Do we really need this on `StringProtocol`? I'd much rather we extend the concrete type we care about, as it puts far less strain on the optimiser.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,111 @@ extension HTTPClient {
...
+    fileprivate func omittingQuotes() -> SubSequence {
         let dquote = "\""
-        if !hasPrefix(dquote) || !hasSuffix(dquote) {
-            return self
+        guard hasPrefix(dquote), hasSuffix(dquote) else {
```
Lukasa, line 124: Prefer `&&` to `,` when not using conditional bindings.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,111 @@ extension HTTPClient {
...
+            return self[startIndex..<endIndex]
```
Lukasa, line 125: A shorthand spelling of this is `return self[...]`

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,111 @@ extension HTTPClient {
...
         }
+        return self.dropFirst().dropLast()
+    }
 
-        let begin = index(after: startIndex)
-        let end = index(before: endIndex)
-        return String(self[begin..<end])
+    fileprivate func trimmingWhitespace() -> SubSequence {
+        guard let start = self.firstIndex(where: { !$0.isWhitespace }) else {
+            return self[self.endIndex..<self.endIndex]
```
Lukasa, line 132: Same shorthand here.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,111 @@ extension HTTPClient {
...
-extension HTTPClient.Response {
-    /// List of HTTP cookies returned by the server.
-    public var cookies: [HTTPClient.Cookie] {
-        return self.headers["set-cookie"].compactMap { HTTPClient.Cookie(header: $0, defaultDomain: self.host) }
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#endif
+import CShims
```
Lukasa, line 167: Can you move these imports to the top of the file?

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,111 @@ extension HTTPClient {
...
+    // All POSIX systems must provide a "POSIX" locale, and its date/time formats are US English.
+    // https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap07.html#tag_07_03_05
+    newlocale(LC_TIME_MASK | LC_NUMERIC_MASK, "POSIX", nil)!
+}()
+
+private func parseCookieTime(_ string: String) -> Int64? {
+    var timeComps = tm()
+    var hasExplicitTimeZone = false
+    parse: do {
+        if swiftahc_cshims_strptime_l(string, "%a, %d %b %Y %H:%M:%S %Z", &timeComps, .init(posixLocale)) != nil {
```
Lukasa, line 179: Instead of putting the `init` at every call site can we just erase the type of the `posixLocale` `private let`? We never need it as a `locale_t` anyway.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,111 @@ extension HTTPClient {
...
+            break parse
+        }
+        timeComps = tm()
+        if swiftahc_cshims_strptime_l(string, "%a, %d-%b-%y %H:%M:%S %Z", &timeComps, .init(posixLocale)) != nil {
+            hasExplicitTimeZone = true
+            break parse
+        }
+        timeComps = tm()
+        if swiftahc_cshims_strptime_l(string, "%a %b %d %H:%M:%S %Y", &timeComps, .init(posixLocale)) != nil {
+            break parse
```
Lukasa, line 190: Rather than break statements, can I ask that we extract these various calls to `strptime` into helper functions and call them in a big conditional binding block?

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,111 @@ extension HTTPClient {
...
+        }
+        return nil
+    }
+    #if canImport(Darwin)
+    // Darwin converts values with explicit time-zones to the local time-zone. Glibc appears not to.
+    // https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/strptime.3.html
+    if hasExplicitTimeZone {
+        let timestamp = Int64(mktime(&timeComps))
+        return timestamp == -1 && errno == EOVERFLOW ? nil : timestamp
```
Lukasa, line 199: We shouldn't have to do this at all: cookie expires are supposed to always be GMT, never local time. Is there a reason we think we need it instead of just only accepting GMT?

Lukasa (changes requested):
Seems like a great idea, thanks! I've left some notes in the diff.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: In practice, most implementations of `strptime` seem to only recognise GMT and sometimes also the current timezone (including Darwin). "%Z" itself isn't actually part of the POSIX spec which is why Darwin and Glibc have different behaviour with regards to converting to local time.

If we only want to support GMT, we could always just look for the literal string "GMT" and chop it off, avoiding all of this. Actually that sounds obviously worthwhile, now that I spell it out like this.

karwa:
Also, the API breakage script doesn't seem to be able to build the module (it's failing to build/find the shims library, from what I can gather).

I took a look at it, but it uses a lot of swift commands I'm not familiar with (like `dump-package` and `api-digester`), so I'm not sure how to fix it.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
Lukasa: Literal string it is!

karwa:
Just an FYI: I'm a bit swamped at the moment, but I'll push an update to address the feedback later this week

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: Actually this is between endIndex and endIndex

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: I'm not sure what you mean by this - do you mean each format should be its own function, and we do something like:

```swift
let _parsedTime = parseFormatA(...) ?? parseFormatB(...) ?? parseFormatC(...)
guard let parsedTime = _parsedTime else { return nil }
```

I'm not sure if that would be easier to read, but I also don't mind if that's what you want.

karwa:
OK apologies for taking so long, and I'm aware people will likely be on holiday so maybe not able to review it for a while.

I think I've addressed all the feedback: C library has been renamed, cookie parsing now uses the UTF-8 view, the Foundation Date's setter has been restored, etc. I've also added a lot more tests for invalid date strings.

The test failures on 5.5/nightly appear to be related to async/Sendable stuff, so nothing to do with this patch.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,116 @@ extension HTTPClient {
...
+    }
+    return timeComponents
+}
+
+private func parseCookieTime(_ timestampUTF8: String.UTF8View.SubSequence, in header: String) -> Int64? {
+    if timestampUTF8.contains(where: { $0 < 0x20 /* Control characters */ || $0 == 0x7F /* DEL */ }) {
+        return nil
+    }
+    let timestampString: String
+    if timestampUTF8.reversed().starts(with: "TMG".utf8) {
```
karwa, line 190: Yeah, there is no general "hasSuffix" function, so this is the best I could do 😅 

I thought about doing `.reversed().starts(with: "GMT".utf8.reversed())`, but the double-reversing could have a performance/size impact.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,116 @@ extension HTTPClient {
...
     }
 }
 
 extension String {
-    fileprivate func omittingQuotes() -> String {
-        let dquote = "\""
-        if !hasPrefix(dquote) || !hasSuffix(dquote) {
-            return self
+    fileprivate init(utf8Slice: String.UTF8View.SubSequence, in base: String) {
+        self.init(base[utf8Slice.startIndex..<utf8Slice.endIndex])
```
Lukasa, line 123: Why is it necessary for us to slice the base? Isn't `utf8Slice` already the subsequence we're passing here?

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
Lukasa: I think the only reason this is here is because we construct new `String`s by using the subsequence indices into the parent. If we drop that (as suggested above) we can just return an empty `Substring` here, which is zero-alloc and avoids this weird-looking construction.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -42,74 +47,39 @@ extension HTTPClient {
...
+            for component in components {
+                switch component.parseCookieComponent() {
+                case ("path", .some(let value))?:
+                    self.path = String(utf8Slice: value, in: header)
+                case ("domain", .some(let value))?:
+                    self.domain = String(utf8Slice: value, in: header)
+                case ("expires", .some(let value))?:
+                    self.expires_timestamp = parseCookieTime(value, in: header)
+                case ("max-age", .some(let value))?:
+                    self.maxAge = Int(String(utf8Slice: value, in: header))
```
Lukasa, line 79: Can we avoid reifying the `String` here? We can construct `Int` from an arbitrary `StringProtocol`, which will save us the conversion cost.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,116 @@ extension HTTPClient {
...
+
+    /// Parses this collection as either a key-value pair, or a plain key.
+    /// The returned key is trimmed of ASCII spaces and normalized to lowercase.
+    /// The returned value is trimmed of ASCII spaces.
+    fileprivate func parseCookieComponent() -> (key: String, value: SubSequence?)? {
+        let (trimmedName, trimmedValue) = self.parseKeyValuePair() ?? (self.trimmingASCIISpaces(), nil)
+        guard !trimmedName.isEmpty else {
+            return nil
+        }
+        return (String(decoding: trimmedName, as: UTF8.self).lowercased(), trimmedValue)
```
Lukasa, line 166: Do we need to do a `decoding` flow here? I think we can just construct the `String` directly from the `Substring.UTF8View`. Additionally, can we apply `lowercased` on the `Substring.UTF8View` ourselves? As we believe the text is entirely ASCII it's a simple bitwise operation, while lowercasing a Unicode string is complex.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
Lukasa: Honestly I think a one-off helper function will help with clarity here. :wink:

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -124,51 +94,116 @@ extension HTTPClient {
...
+}
+
+private func parseCookieTime(_ timestampUTF8: String.UTF8View.SubSequence, in header: String) -> Int64? {
+    if timestampUTF8.contains(where: { $0 < 0x20 /* Control characters */ || $0 == 0x7F /* DEL */ }) {
+        return nil
+    }
+    let timestampString: String
+    if timestampUTF8.reversed().starts(with: "TMG".utf8) {
+        let timezoneStart = timestampUTF8.index(timestampUTF8.endIndex, offsetBy: -3)
+        let trimmedTimestampUTF8 = timestampUTF8[..<timezoneStart].trimmingASCIISpaces()
```
Lukasa, line 192: Rather than doing `index(_:offsetBy:)` here, just do `dropLast(3)`?

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: - `String.init?(String.UTF8View.Subsequence)` can fail if the indexes are not aligned to character boundaries. We don't expect that to happen, but the SSWG guidelines also say to avoid force-unwrapping wherever possible.

- `String.init(Substring)` (used here) can not fail. If the indexes are not aligned to character boundaries, it will adjust them. It's... not unreasonable fallback behaviour, and means we can avoid a bunch of optionals that won't be ever be nil.

In terms of performance, they both allocate independent String values, so are both O(n). I would suppose the `String.init(Substring)` initializer can be slightly faster, as by widening the bounds to aligned character boundaries, it may be able to skip UTF-8 validation.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: I don't think it's that weird, but I guess it depends what you're used to. This is quite a common thing when you need to construct an empty range of a collection. As it is just making a slice, it is also zero-alloc.

I think it would be much weirder if this method returned a slice of an entirely different instance, whose indexes are not compatible with the original slice or its base collection. Are you sure it's worth doing that just to avoid writing it this way?

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: Technically, nothing checks that the string is ASCII, and that's the sort of optimisation which I would be very disappointed if the standard library didn't do by itself (especially since it has internal flags which means it already knows whether a String is ASCII).

But adding ASCII checks is only part of it - the set of allowed bytes depends on which part of the cookie string you're talking about ([RFC-6265](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1) - note that `cookie-octet` is not the same as `path-value` or `extension-av`, and is actually a restricted subset of ASCII which excludes control characters and certain delimiters).

It would be useful to add those checks at some point, and the move to parsing at the UTF-8 level makes it easier to add that as a follow-up. But those should be accompanied by much more comprehensive tests which would bloat this PR, IMO, and distract from its focus which is timestamp parsing.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: We only need the index, and this function already includes variables called `timestampUTF8`, `trimmedTimestampUTF8`, `timestampString`, so storing a single index called `timezoneStart` seems most easily understandable IMO. 

And we can't do something like:

```swift
var trimmedTimestampUTF8 = timestampUTF8.dropLast(3)
trimmedTimestampUTF8 = trimmedTimestampUTF8.trimmingASCIISpaces()
```

...because the point is we need to check that `trimmedTimestampUTF8.endIndex != timezoneStart` (i.e. there was at least one space between the timezone and rest of the timestamp).

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: I've added a doc-comment, explaining how this is different from `String.init?(String.UTF8View.Subsequence)`

karwa:
@Lukasa , I've made the following changes:
- Added a doc-comment explaining how `String(utf8Slice:in:)` is different from `String.init?(String.UTF8View.Subsequence)`
- Parse max-age directly from the substring
- Added a `hasSuffix` function, with tests
- Added `FIXME`s to `HTTPCookie.init?(header:defaultDomain)` and `HTTPCookie.init(key:value:...)` noting that they need stricter validation. In particular, the latter needs to be failable (which would be an API-breaking change).

The `hasSuffix` function is generally quite useful, so I've made it generic. It's also worth testing it for edge-cases, and generics helps with that.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
Lukasa: I think I'd rather have the optionals, tbh. The SSWG guidelines are understandable, but we're asserting that these are ASCII strings, and so there can never be a case where we have alignment issues. However, I'd like to hear from @fabianfett (SSWG member with a lot of interest in AHC) and @tomerd (SSWG member) for some more guidance here.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
Lukasa: The standard library does perform it, though it still has to alloc a new string to do it. C'est la vie. For now I'm happy to delay on this.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: Well, technically we're only asserting that our processing (trimming, etc) operates on whole scalars, which is fine since we only do ASCII-level processing.

Force-unwrapping exists for occasions where we have an optional which we can determine will never be nil, which IMO is covered by this. But yeah, the SSWG guidelines say we should avoid that, and there is a relatively easy way around it by going through `Substring`. I don't mind either way.

karwa:
OK, 2 more commits:

1. Renamed `String(utf8Slice:from:)` to `String(aligningUTF8:)` (dropping the base-string argument), implemented it by going through `Substring`, and removed it from timestamp parsing (it actually helped make the code a bit clearer).
2. While taking a look at [Section 5.2 of RFC-6265](https://datatracker.ietf.org/doc/html/rfc6265#section-5.2) (which describes the lenient parser we "MUST" use to parse Set-Cookie headers), I noticed that the way we handle parsed values and decide which ones to use was all wrong - we shouldn't be skipping `nil` values in all cases (such as the path), and we weren't handling invalid values properly for many of the other cases (we should ignore them, not set to `nil`). So I fixed it, and added a bunch of test cases for each component. I was going to leave it for a follow-up PR, but I figured that this one already rewrites basically the entire rest of the parser, and misinterpreting cookies could have privacy implications so it's important to fix it ASAP.

Sources/AsyncHTTPClient/FoundationExtensions.swift
```diff
@@ -0,0 +1,56 @@
...
+    /// - parameters:
+    ///     - name: The name of the cookie.
+    ///     - value: The cookie's string value.
+    ///     - path: The cookie's path.
+    ///     - domain: The domain of the cookie, defaults to nil.
+    ///     - expires: The cookie's expiration date, defaults to nil.
+    ///     - maxAge: The cookie's age in seconds, defaults to nil.
+    ///     - httpOnly: Whether this cookie should be used by HTTP servers only, defaults to false.
+    ///     - secure: Whether this cookie should only be sent using secure channels, defaults to false.
+    public init(name: String, value: String, path: String = "/", domain: String? = nil, expires: Date? = nil, maxAge: Int? = nil, httpOnly: Bool = false, secure: Bool = false) {
```
Lukasa, line 42: As this is public API, should we make some room for ourselves by making this failable now?

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -42,79 +47,71 @@ extension HTTPClient {
...
+            var parsedDomain: String.UTF8View.SubSequence?
 
-                    formatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss z"
-                    if let date = formatter.date(from: value) {
-                        self.expires = date
-                        continue
+            for component in components {
+                switch component.parseCookieComponent() {
+                case ("path", let value)?:
+                    // Unlike other values, unspecified and empty paths reset to the default path.
```
Lukasa, line 76: Can we cite comments like this? I love to see them and I'm really pleased this one is here, but it makes my life much easier as a reviewer (and future reader) if I can rapidly follow the citation (in this case, RFC 6265 § 5.2.4).

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -42,79 +47,71 @@ extension HTTPClient {
...
                     }
-
-                    formatter.dateFormat = "EEE, dd-MMM-yy HH:mm:ss z"
-                    if let date = formatter.date(from: value) {
-                        self.expires = date
-                        continue
+                    parsedPath = value
+                case ("domain", let value)?:
+                    guard var value = value, !value.isEmpty else {
+                        break
```
Lukasa, line 84: This is a really mild nit and you can ignore it if you feel strongly, but I kinda hate `break` in a `switch` nested in a loop. It takes me a second or two to figure out to what it applies. Can we use `continue` here instead?

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
```diff
@@ -42,79 +47,71 @@ extension HTTPClient {
...
+                    }
+                    parsedDomain = value
+                case ("expires", let value)?:
+                    guard let value = value, let timestamp = parseCookieTime(value) else {
+                        break
+                    }
+                    self.expires_timestamp = timestamp
+                case ("max-age", let value)?:
+                    guard let value = value, let age = Int(Substring(value)) else {
```
Lukasa, line 100: Fun task for the future: let's just write our own `Int.init(_ bytes: Substring.UTF8View)` initializer. It doesn't really matter, but it'd be fun and very slightly faster. Let's not do it now though, just planting seeds to keep you engaged in the future :wink:.

Sources/AsyncHTTPClient/FoundationExtensions.swift
karwa: Making it failable would be an API-breaking change (semver-major) though, right?

It would be good to do, but I'd rather land this and check that out as a follow-up.

Sources/AsyncHTTPClient/HTTPClient+HTTPCookie.swift
karwa: Yeah, it's not too difficult. I already did a similar thing for WebURL, because the standard library's version only accepts `StringProtocol` rather than arbitrary buffers of bytes: https://github.com/karwa/swift-url/blob/1f1263f94e25eb43d8688dc668de6638e092f748/Sources/WebURL/Util/ASCII.swift#L502

But yeah, it can be a follow-up. I'll have a patch up which addresses the nits in a little bit.

karwa:
@Lukasa okay, addressed 2 nits:
- Added a citation for how the "path" value is handled
- Changed `continue` to `break` in the component-parsing loop

As for making the memberwise public initialiser failable, I'm not sure how API-breaking/semver-major changes are handled (leave the PR sitting around until the next release is scheduled? conditional compilation?), but it could be done by a follow-up PR.

Sources/AsyncHTTPClient/FoundationExtensions.swift
Lukasa: Yup, I lost track of the fact that this is an existing `init`. We'd need to deprecate-and-replace it. Happy to delay that to a follow-up.

PR #511, opened 2021-12-02
Add HTTPClientReuqest.Prepared

dnadoba:
Before we can execute a `HTTPClientRequest` we need do a couple of validation and transformation steps which can fail. 
The result of the preparation is now encapsulated in a separate object called `HTTPClientRequest.Prepared`.

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Prepared.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Prepared.swift
@@ -1,0 +1,70 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
+// Licensed under Apache License v2.0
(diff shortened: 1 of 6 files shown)


fabianfett (approved):
LGTM. However we should add tests with a follow up PR.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Prepared.swift
```diff
@@ -0,0 +1,70 @@
...
+        let poolKey: ConnectionPool.Key
+        let requestFramingMetadata: RequestFramingMetadata
+        let head: HTTPRequestHead
+        let body: Body?
+    }
+}
+
+@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+extension HTTPClientRequest {
+    func prepared() throws -> Prepared {
```
Lukasa, line 30: I think I'd rather write this as an `init` on `Prepared`, if that's ok.

Lukasa (approved):
One note about API changes, otherwise this looks good.

PR #512, opened 2021-12-03
Improve test utils for running request

fabianfett:
### Motivation

To land the support for async/await, we need test utilities. We already have a `MockRequestExecutor`. Let's improve this to better handle on and off request processing.

### Changes

- Move `MockRequestExecutor` into its own file
- Add blocking APIs to `MockRequestExecutor` when called from another thread


fabianfett:
Nightly fails because of Sendable requirements.

dnadoba (approved):

PR #513, opened 2021-12-03
Test `HTTPClientRequest`

dnadoba:
Add tests for `HTTPClientRequest`, `HTTPClientRequest.Prepared` and `HTTPClientRequest.Body`.


Tests/AsyncHTTPClientTests/HTTPClientRequestTests.swift
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    func testPostWithSequenceOfUnknownLength() {
+        #if compiler(>=5.5) && canImport(_Concurrency)
+        guard #available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *) else { return }
+        XCTAsyncTest {
+            var request = Request(url: "http://example.com/post")
+            request.method = .POST
+            let sequence = AnySequence(ByteBuffer(string: "post body").readableBytesView)
+            request.body = .bytes(sequence)
```
fabianfett, line 169: Do we want to have this overload on `.bytes()`? I thought we only had `.bytes(length: Int, Sequence)` and `.bytes(Collection)`? Do we maybe have a default property on length? If so I think we should remove it.

Tests/AsyncHTTPClientTests/HTTPClientRequestTests.swift
```diff
@@ -0,0 +1,444 @@
...
+            ))
+            XCTAssertEqual(preparedRequest.requestFramingMetadata, .init(
+                connectionClose: false,
+                body: .fixedSize(9)
+            ))
+            guard let buffer = await XCTAssertNoThrowWithResult(try await preparedRequest.body.read()) else { return }
+            XCTAssertEqual(buffer, .init(string: "post body"))
+        }
+        #endif
+    }
```
fabianfett, line 351: Those are awesome! Thanks! The only other thing I can think of might, be three tests for the different unix schemes.

fabianfett: Oh and we should have an explicit test on the ConnectionPool.Key ensuring that `unix+http` != `unix`

Tests/AsyncHTTPClientTests/HTTPClientRequestTests.swift
dnadoba: We currently default to `nil` but I was thinking that we should make `length` non-optional and probably also move it the end. WDYT? However, I would then do that in a follow up PR.

dnadoba:
Nightly fails because of Sendable

fabianfett (approved):

PR #515, opened 2021-12-03
Use the current `EventLoop` as the requests preferred `EventLoop`

dnadoba:
NIO gained a new method on `EventLoopGroup` called `.any()` to get the current `EventLoop` if possible. We should use this method to selected the requests preferred `EventLoop` when the user didn't explicitly set a required `EventLoop`. This should improve performance because we increase the chance that we do not need to hop between `EventLoops` to send and receive data.


fabianfett (approved):
Great change

Lukasa (approved):

PR #516, opened 2021-12-04
Make tests faster, by reducing timeout margins

fabianfett:
### Motivation

We like fast tests. For this reason we enabled parallel testing recently. However there are a number of tests that have unnecessary high timeout margins. If we reduce those tests will be even faster.

### Changes

- Reduce connection timeout numbers from 10sec (default) down to 2sec (explicit)
- Clean up a number of tests. (Removed last instance of should fail)

### Result

Faster and cleaner tests


fabianfett:
Nightly fails because of Sendable requirements.

Tests/AsyncHTTPClientTests/HTTPConnectionPool+FactoryTests.swift
```diff
@@ -76,7 +76,10 @@ class HTTPConnectionPool_FactoryTests: XCTestCase {
         let factory = HTTPConnectionPool.ConnectionFactory(
             key: .init(request),
             tlsConfiguration: nil,
-            clientConfiguration: .init(proxy: .socksServer(host: "127.0.0.1", port: server!.localAddress!.port!)),
+            clientConfiguration: .init(
+                timeout: .init(connect: .seconds(3), read: nil),
```
dnadoba, line 80: Why do we use 3 seconds instead of 2 seconds here?

fabianfett:
> Nevertheless, I think the end goal should be to use an EmbeddedEventLoop in combination with advanceTime(to:). While I agree in general, I'm not sure this actually works for integration tests. In all our unit tests, we should already be using `EmbeddedEventLoop` or just being able to call some timeout method (in the state machine tests)

fabianfett:
@swift-server-bot test please.

dnadoba:

PR #518, opened 2021-12-06
Add async Transaction

fabianfett:
This PR is very early. But I would like to start to get some feedback on its general form.

### Motivation

I hope this is the last very large PR in a long time forward. This PR introduces an async Transaction object. The object is the translation layer between the user facing async API and the NIO channel handler. This handler is currently lacking any redirect logic.

### Changes

- Add `Transaction` and a corresponding state machine

### Result

We will be able to support an async API.


Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,189 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.5) && canImport(_Concurrency)
```
fabianfett, line 15: This file is part of #519. Please review there.

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    private func continueRequestBodyStream(
+        _ allocator: ByteBufferAllocator,
+        next: @escaping ((ByteBufferAllocator) async throws -> ByteBuffer?)
+    ) {
+        Task {
+            while true {
+                do {
+                    guard let part = try await next(allocator) else { // <---- dispatch point!
```
Lukasa, line 85: Prefer `while let`.

Lukasa (commented):
So far this pattern is looking good

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,552 @@
...
+                self.state = .finished(error: nil, streamID)
+                return .succeedContinuation(continuation, nil)
+
+            case .executing(let context, let requestState, .waitingForStream(var buffer, next: .eof)):
+                assert(!buffer.isEmpty)
+                let toReturn = buffer.removeFirst()
+                self.state = .executing(context, requestState, .buffering(streamID, buffer, next: .eof))
+                return .succeedContinuation(continuation, toReturn)
+
+            case .executing(let context, let requestState, .buffering(let streamID, var buffer, next: .askExecutorForMore)):
```
dnadoba, line 450: We should check that the `streamID` in the state is the same as the `streamID` from the parameter.

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,552 @@
...
+
+            case .executing(_, _, .buffering(let registeredStreamID, let buffer, next: .eof)) where buffer.isEmpty:
+                guard registeredStreamID == streamID else {
+                    return .failContinuation(continuation, TriedToRegisteredASecondConsumer())
+                }
+                self.state = .finished(error: nil, registeredStreamID)
+                return .succeedContinuation(continuation, nil)
+
+            case .executing(let context, let requestState, .buffering(let streamID, var buffer, next: .eof)):
+                assert(!buffer.isEmpty)
```
dnadoba, line 475: `streamID` check missing

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,552 @@
...
+                let toReturn = buffer.removeFirst()
+                self.state = .executing(context, requestState, .buffering(streamID, buffer, next: .eof))
+                return .succeedContinuation(continuation, toReturn)
+
+            case .executing(_, _, .waitingForRemote(let registeredStreamID, _)):
+                if registeredStreamID != streamID {
```
dnadoba, line 481: nit: use a guard across all streamID checks to be more consistent.

Tests/AsyncHTTPClientTests/TransactionTests.swift
```diff
@@ -0,0 +1,585 @@
...
+    private var iterator: Iterator
+
+    init(_ iterator: Iterator) {
+        self.iterator = iterator
+    }
+
+    func next() async throws -> Iterator.Element? {
+        var iter = self.iterator
+        defer { self.iterator = iter }
+        return try await iter.next()
```
dnadoba, line 552: I think this will only really work in the general case if Iterator has value semantics. Otherwise, concurrent access will still result in two next calls. AFAIK, our current test do not call `next()` concurrently but maybe still worth documenting or adding some precondition that checks we always leave `next()` before we enter `next()` it again.

dnadoba (commented):
Great work! This is the last major step towards async/await support!

Tests/AsyncHTTPClientTests/TransactionTests.swift
fabianfett: We have a precondition for this...
```
preconditionFailure("A body response continuation from this iterator already exists! Queuing calls to `next()` is not supported.")
```

dnadoba (approved):
LGTM

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+            case cancel(HTTPRequestExecutor)
+            case none
+        }
+
+        mutating func willExecuteRequest(_ executor: HTTPRequestExecutor) -> StartExecutionAction {
+            switch self.state {
+            case .initialized(let continuation), .queued(let continuation, _):
+                let context = ExecutionContext(
+                    executor: executor,
+                    allocator: .init(),
```
Lukasa, line 182: Do we have any plan to properly construct an allocator?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+            case .executing(_, .finished, _):
+                // the channels writability changed to writable after we have forwarded all the
+                // request bytes. Can be ignored.
+                return .none
+
+            case .executing(_, .initialized, .waitingForStream),
+                 .executing(_, .initialized, .buffering),
+                 .executing(_, .initialized, .waitingForRemote),
+                 .executing(_, .initialized, .finished):
+                preconditionFailure("Invalid state: Response can not be received before request. \(self.state)")
```
Lukasa, line 230: This seems odd: haven't we sent the request head now? We haven't sent the request _body_, but that's a different question. Are we confident there's no interleaving of events that allows this to be hit?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                 .executing(_, .initialized, _):
+                preconditionFailure("A request stream can only be resumed, if the request was started")
+
+            case .executing(let context, .producing, let responseSteam):
+                self.state = .executing(context, .paused, responseSteam)
+
+            case .executing(_, .paused, _),
+                 .executing(_, .finished, _),
+                 .finished:
+                // the channels writability changed to writable after we have already forwarded all
```
Lukasa, line 250: Presumably it didn't change to _writable_?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                 .queued,
+                 .executing(_, _, .initialized):
+                preconditionFailure("Invalid state: \(self.state)")
+
+            case .executing(_, _, .waitingForStream(_, next: .endOfFile)):
+                self.state = .finished(error: nil, nil)
+                return .none
+
+            case .executing(let context, _, .waitingForStream(_, next: .askExecutorForMore)):
+                self.state = .finished(error: nil, nil)
```
Lukasa, line 409: Why is this not an error?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                 .executing(_, _, .finished):
+                // user is consuming the stream with an iterator
+                return .none
+            }
+        }
+
+        mutating func cancelResponseStream(streamID: HTTPClientResponse.Body.IteratorStream.ID) -> FailAction {
+            switch self.state {
+            case .initialized, .queued, .executing(_, _, .initialized):
+                preconditionFailure("Got notice about a deinited iterator, before we even received a response. Invalid state: \(self.state)")
```
Lukasa, line 431: Weird to have `responseBodyDeinited` not complain about a deinit, but this one does.

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+
+            case .executing(_, _, .buffering(let registeredStreamID, _, next: _)),
+                 .executing(_, _, .waitingForRemote(let registeredStreamID, _)):
+                guard registeredStreamID == streamID else {
+                    // we got informed about an iterator that we don't actually serve!
```
Lukasa, line 436: Why is this not a `preconditionFailure`?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                    return .none
+                }
+                return self.fail(HTTPClientError.cancelled)
+
+            case .executing(_, _, .waitingForStream),
```
Lukasa, line 441: Uh, is `.waitingForStream` equivalent to `.finished` here? Why isn't this an error?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+            }
+        }
+
+        enum ConsumeAction {
+            case succeedContinuation(CheckedContinuation<ByteBuffer?, Error>, ByteBuffer?)
+            case failContinuation(CheckedContinuation<ByteBuffer?, Error>, Error)
+            case askExecutorForMore(HTTPRequestExecutor)
+        }
+
+        struct TriedToRegisteredASecondConsumer: Error {}
```
Lukasa, line 455:
```suggestion
        struct TriedToRegisterASecondConsumer: Error {}
```

Lukasa: Additionally, errors that are entirely internal can be enums.

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+
+            case .executing(let context, let requestState, .waitingForStream(var buffer, next: .endOfFile)):
+                assert(!buffer.isEmpty)
+                let toReturn = buffer.removeFirst()
+                self.state = .executing(context, requestState, .buffering(streamID, buffer, next: .endOfFile))
+                return .succeedContinuation(continuation, toReturn)
+
+            case .executing(let context, let requestState, .buffering(let registeredStreamID, var buffer, next: .askExecutorForMore)):
+                guard registeredStreamID == streamID else {
+                    return .failContinuation(continuation, TriedToRegisteredASecondConsumer())
```
Lukasa, line 496: Same question here, why isn't this a `preconditionFailure`?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                    return .askExecutorForMore(context.executor)
+                } else {
+                    let toReturn = buffer.removeFirst()
+                    self.state = .executing(context, requestState, .buffering(streamID, buffer, next: .askExecutorForMore))
+                    return .succeedContinuation(continuation, toReturn)
+                }
+
+            case .executing(_, _, .buffering(let registeredStreamID, _, next: .error(let error))):
+                guard registeredStreamID == streamID else {
+                    return .failContinuation(continuation, TriedToRegisteredASecondConsumer())
```
Lukasa, line 510: Same here, why isn't this a `preconditionFailure`?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                }
+                self.state = .finished(error: error, registeredStreamID)
+                return .failContinuation(continuation, error)
+
+            case .executing(_, _, .buffering(let registeredStreamID, let buffer, next: .endOfFile)) where buffer.isEmpty:
+                guard registeredStreamID == streamID else {
```
Lukasa, line 516: And here

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                    return .failContinuation(continuation, TriedToRegisteredASecondConsumer())
+                }
+                self.state = .finished(error: nil, registeredStreamID)
+                return .succeedContinuation(continuation, nil)
+
+            case .executing(let context, let requestState, .buffering(let registeredStreamID, var buffer, next: .endOfFile)):
+                guard registeredStreamID == streamID else {
```
Lukasa, line 523: Same here.

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                    return .failContinuation(continuation, TriedToRegisteredASecondConsumer())
+                }
+                assert(!buffer.isEmpty) // TODO: This looks fishy!
+                let toReturn = buffer.removeFirst()
+                self.state = .executing(context, requestState, .buffering(streamID, buffer, next: .endOfFile))
```
Lukasa, line 528: I'm confused, how do we get out of this state?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
```diff
@@ -0,0 +1,603 @@
...
+                return .succeedContinuation(continuation, toReturn)
+
+            case .executing(_, _, .waitingForRemote(let registeredStreamID, _)):
+                guard registeredStreamID == streamID else {
+                    return .failContinuation(continuation, TriedToRegisteredASecondConsumer())
+                }
+                preconditionFailure("A body response continuation from this iterator already exists! Queuing calls to `next()` is not supported.")
```
Lukasa, line 535: Is there any reason this `preconditionFailure` is less important than the check above?

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
```diff
@@ -0,0 +1,342 @@
...
+    private func writeOnceAndOneTimeOnly(byteBuffer: ByteBuffer) {
+        // This method is synchronously invoked after sending the request head. For this reason we
+        // can make a number of assumptions, how the state machine will react.
+        let writeAction = self.stateLock.withLock {
+            self.state.producedNextRequestPart(byteBuffer)
+        }
+
+        switch writeAction {
+        case .write(let part, let executor, continue: _):
+            // the continue flag can be ignored. We will maximum write the request's end.
```
Lukasa, line 66: The second sentence here is very confusing.

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
```diff
@@ -0,0 +1,342 @@
...
+            // an error/cancellation has happened. we don't need to continue here
+            return
+        }
+    }
+
+    private func continueRequestBodyStream(
+        _ allocator: ByteBufferAllocator,
+        next: @escaping ((ByteBufferAllocator) async throws -> ByteBuffer?)
+    ) {
+        Task {
```
Lukasa, line 92: Do we want to create new tasks here? Is there any reason not to use a single long async sequence and be able to translate `pause` into an `await` call?

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
fabianfett: Yes, my plan is to expose the channel allocator on `HTTPRequestExecutor` in a follow up pr.

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
fabianfett: Changed the name of `initialized` to `requestHeadSent` and on the response side I changed `initialized` to `waitingForResponseHead`. This made my error here very obvious!

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
fabianfett: It doesn't actually matter I assume. Everything will be cleaned up after this anyways, and the user isn't interested in the response body. Happy to change to `HTTPClientError.cancelled`.

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
fabianfett: We could do this. However in this case we would have at least one allocation for every write pause, as we would need to create a new continuation. 

I guess in the NIO/Async translation layer the same rules should apply to Continuations that apply for EventLoopFutures: We should try to reduce them as much as possible.

fabianfett: Just thinking about it, a new Task probably also causes a new allocation. So the issue comes down to: Do we want to allocate a continuation or a new Task for every pause.

fabianfett: I guess the big pro for creating a new task for each continue here, is that we don't have to do a fastPath/slowPath approach here to not allocate a Continuation on every body part.

Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
fabianfett: Renamed `.waitingForStream` to `.waitingForResponseIterator`. To make its meaning clearer.

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
Lukasa: Why would we have to allocate a new continuation on every body part? The same part of the code that tells us to wait can allocate the new continuation.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    init(
+        bag: Transaction,
+        version: HTTPVersion,
+        status: HTTPResponseStatus,
+        headers: HTTPHeaders
+    ) {
+        self.body = .init(bag)
```
Lukasa, line 42: Nit: I this this is less clear and less concise than

```suggestion
        self.body = Body(bag)
```

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
```diff
@@ -0,0 +1,342 @@
...
+        }
+
+        self.requestBodyStreamFinished()
+    }
+
+    private func continueRequestBodyStream(
+        _ allocator: ByteBufferAllocator,
+        next: @escaping ((ByteBufferAllocator) async throws -> ByteBuffer?)
+    ) {
+        Task.detached {
```
Lukasa, line 80: I don't think we need `Task.detached`, regular `Task` is fine.

Lukasa (approved):

PR #519, opened 2021-12-06
Add an AsyncWriter

fabianfett:
### Motivation

This PR adds an AsyncWriter that we can be used to test our new async upload streaming API.

### Changes

- Adds an AsyncWriter to test utils

### Result

We will be able to test async upload streaming API in #518.

### DIFF
--- a/Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
+++ b/Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
@@ -96,7 +96,7 @@ class AsyncSequenceWriter: AsyncSequence {
             }
 
         case .buffering(var buffer, let demandContinuation):
-            let first = buffer.popFirst()!
+            let first = buffer.removeFirst()
             if first != nil {
                 self._state = .buffering(buffer, demandContinuation)
             } else {
--- a/Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
+++ b/Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
@@ -17,16 +17,13 @@ import NIOConcurrencyHelpers
 import NIOCore
 
 @available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
-class AsyncSequenceWriter<E>: AsyncSequence {
+class AsyncSequenceWriter<Element>: AsyncSequence {
     typealias AsyncIterator = Iterator
-    typealias Element = E
 
     struct Iterator: AsyncIteratorProtocol {
-        typealias Element = E
+        private let writer: AsyncSequenceWriter<Element>
 
-        private let writer: AsyncSequenceWriter
-
-        init(_ writer: AsyncSequenceWriter) {
+        init(_ writer: AsyncSequenceWriter<Element>) {
             self.writer = writer
         }
 
@@ -40,10 +37,10 @@ class AsyncSequenceWriter<E>: AsyncSequence {
     }
 
     private enum State {
-        case buffering(CircularBuffer<Element?>, CheckedContinuation<Void, Error>?)
+        case buffering(CircularBuffer<Element?>, CheckedContinuation<Void, Never>?)
         case finished
         case waiting(CheckedContinuation<Element?, Error>)
-        case failed(Error)
+        case failed(Error, CheckedContinuation<Void, Never>?)
     }
 
     private var _state = State.buffering(.init(), nil)
@@ -60,12 +57,13 @@ class AsyncSequenceWriter<E>: AsyncSequence {
         }
     }
 
-    public func demand() async throws {
+    /// Wait until a downstream consumer has issued more demand by calling `next`.
+    public func demand() async {
         self.lock.lock()
 
         switch self._state {
         case .buffering(let buffer, .none):
-            try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
+            await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in
                 self._state = .buffering(buffer, continuation)
                 self.lock.unlock()
             }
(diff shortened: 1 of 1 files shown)


Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,189 @@
...
+            switch self._state {
+            case .failed, .finished, .buffering:
+                return false
+            case .waiting:
+                return true
+            }
+        }
+    }
+
+    public func demand() async throws {
```
dnadoba, line 63: What is this method used for? Does it wait until someone calls next but the buffer is empty?

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,189 @@
...
+
+    public func fail(_ error: Error) {
+        enum ErrorAction {
+            case failContinuation(CheckedContinuation<ByteBuffer?, Error>, Error)
+            case none
+        }
+
+        let errorAction = self.lock.withLock { () -> ErrorAction in
+            switch self._state {
+            case .buffering:
```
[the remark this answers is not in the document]
dnadoba: We should add a precondition that checks that the demand continuation is nil, otherwise we will never resume it.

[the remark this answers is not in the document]
dnadoba: Or fail the continuation.

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,189 @@
...
+            try await self.writer.next()
+        }
+    }
+
+    func makeAsyncIterator() -> Iterator {
+        return Iterator(self)
+    }
+
+    private enum State {
+        case buffering(CircularBuffer<ByteBuffer?>, CheckedContinuation<Void, Error>?)
```
dnadoba, line 43: Do we need to know that this is a ByteBuffer? Otherwise I would prefer it to be generic to reduce the accessible API surface inside the implementation of `AsyncSequenceWriter`.

dnadoba (changes requested):

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,189 @@
...
+        switch self._state {
+        case .buffering(let buffer, let demandContinuation) where buffer.isEmpty:
+            return try await withCheckedThrowingContinuation { continuation in
+                self._state = .waiting(continuation)
+                self.lock.unlock()
+                demandContinuation?.resume(returning: ())
+            }
+
+        case .buffering(var buffer, let demandContinuation):
+            let first = buffer.popFirst()!
```
glbrntt, line 99: Add a comment that we know from above that the buffer isn't empty too.

```suggestion
            let first = buffer.removeFirst()
```

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,189 @@
...
+            self.lock.unlock()
+            return first
+
+        case .failed(let error):
+            self._state = .finished
+            self.lock.unlock()
+            throw error
+
+        case .finished:
+            return nil
```
glbrntt, line 114: You're missing an unlock here

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,189 @@
...
+
+        case .waiting:
+            preconditionFailure("How can this be called twice?!")
```
glbrntt, line 117: I generally prefer precondition failure message to be clearer about what actually failed but not a big deal for test utils

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
fabianfett: Added a comment.

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,197 @@
...
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.5) && canImport(_Concurrency)
+import NIOConcurrencyHelpers
+import NIOCore
+
+@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+class AsyncSequenceWriter<E>: AsyncSequence {
+    typealias AsyncIterator = Iterator
+    typealias Element = E
```
dnadoba, line 22: We can just call the generic `Element` and then don't need this typealias.

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,197 @@
...
+    public func end() {
+        self.writeBufferOrEnd(nil)
+    }
+    
+    private enum WriteAction {
+        case succeedContinuation(CheckedContinuation<Element?, Error>, Element?)
+        case none
+    }
+
+    private func writeBufferOrEnd(_ byteBuffer: Element?) {
```
dnadoba, line 140: nit: rename `byteBuffer` to just `element` everywere

dnadoba (approved):

fabianfett:
Thanks for the thorough code review @dnadoba. Great catches. Hope all is fixed now.

dnadoba (approved):
Great! That's really convenient test utility which we will use a lot in the upcoming async/await tests.

dnadoba:
@swift-server-bot test please.

dnadoba:

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -0,0 +1,194 @@
...
+        case buffering(CircularBuffer<Element?>, CheckedContinuation<Void, Never>?)
+        case finished
+        case waiting(CheckedContinuation<Element?, Error>)
+        case failed(Error, CheckedContinuation<Void, Never>?)
+    }
+
+    private var _state = State.buffering(.init(), nil)
+    private let lock = Lock()
+
+    public var hasDemand: Bool {
```
glbrntt, line 49: nit: drop the `public` as the class is `internal` (and elsewhere)

glbrntt (approved):

PR #522, opened 2021-12-07
Refactor redirect logic to be reusable for async/await

dnadoba:
### Motivation
We want to reuse the current redirect logic, which was mainly located inside the `RedirectHandler`, for the upcoming async/await implementation.

### Changes
- move `HTTPClient.Request.RedirectState` to the top level and a separate file
- refactor redirect logic from `RedirectHandler` into methods on `HTTPHeaders` and methods on `RedirectState`
- use the refactored methods in `RedirectHandler`

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -504,11 +504,44 @@ public class HTTPClient {
     ///     - delegate: Delegate to process response parts.
     ///     - eventLoop: NIO Event Loop preference.
     ///     - deadline: Point in time by which the request must complete.
-    public func execute<Delegate: HTTPClientResponseDelegate>(request: Request,
-                                                              delegate: Delegate,
-                                                              eventLoop eventLoopPreference: EventLoopPreference,
-                                                              deadline: NIODeadline? = nil,
-                                                              logger originalLogger: Logger?) -> Task<Delegate.Response> {
+    ///     - logger: The logger to use for this request.
+    public func execute<Delegate: HTTPClientResponseDelegate>(
+        request: Request,
+        delegate: Delegate,
+        eventLoop eventLoopPreference: EventLoopPreference,
+        deadline: NIODeadline? = nil,
+        logger originalLogger: Logger?
+    ) -> Task<Delegate.Response> {
+        _execute(
+            request: request,
+            delegate: delegate,
+            eventLoop: eventLoopPreference,
+            deadline: deadline,
+            logger: originalLogger,
+            redirectState: RedirectState(
+                self.configuration.redirectConfiguration.configuration,
+                initialURL: request.url.absoluteString
+            )
+        )
+    }
+    
+    /// Execute arbitrary HTTP request and handle response processing using provided delegate.
+    ///
+    /// - parameters:
+    ///     - request: HTTP request to execute.
+    ///     - delegate: Delegate to process response parts.
+    ///     - eventLoop: NIO Event Loop preference.
+    ///     - deadline: Point in time by which the request must complete.
+    ///     - logger: The logger to use for this request.
+    func _execute<Delegate: HTTPClientResponseDelegate>(
+        request: Request,
+        delegate: Delegate,
+        eventLoop eventLoopPreference: EventLoopPreference,
+        deadline: NIODeadline? = nil,
+        logger originalLogger: Logger?,
+        redirectState: RedirectState?
+    ) -> Task<Delegate.Response> {
+        
         let logger = (originalLogger ?? HTTPClient.loggingDisabled).attachingRequestInformation(request, requestID: globalRequestID.add(1))
         let taskEL: EventLoop
         switch eventLoopPreference.preference {
@@ -544,20 +577,17 @@ public class HTTPClient {
         }
 
         let redirectHandler: RedirectHandler<Delegate.Response>? = {
-            var request = request
-            if request.redirectState == nil {
-                request.redirectState = RedirectState(
-                    self.configuration.redirectConfiguration.configuration,
-                    initialURL: request.url.absoluteString
-                )
-            }
-            guard request.redirectState != nil else { return nil }
+            guard let redirectState = redirectState else { return nil }
 
-            return .init(request: request) { newRequest in
-                self.execute(request: newRequest,
-                             delegate: delegate,
-                             eventLoop: eventLoopPreference,
-                             deadline: deadline)
+            return .init(request: request, redirectState: redirectState) { newRequest, newRedirectState in
+                self._execute(
+                    request: newRequest,
+                    delegate: delegate,
+                    eventLoop: eventLoopPreference,
+                    deadline: deadline,
+                    logger: logger,
+                    redirectState: newRedirectState
+                )
             }
         }()
 
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -519,7 +519,7 @@ public class HTTPClient {
             deadline: deadline,
             logger: originalLogger,
             redirectState: RedirectState(
-                self.configuration.redirectConfiguration.configuration,
+                self.configuration.redirectConfiguration.mode,
                 initialURL: request.url.absoluteString
             )
         )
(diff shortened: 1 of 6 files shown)


Sources/AsyncHTTPClient/RedirectState.swift
```diff
@@ -0,0 +1,134 @@
...
+
+import struct Foundation.URL
+import NIOHTTP1
+
+struct RedirectState {
+    /// number of redirects we are allowed to follow
+    private var count: Int
+
+    /// nil if we should ignore cycles, otherwise a set of all visited URLs
+    private var visited: Set<URL>?
```
fabianfett, line 23: Do we maybe want to make the ignore cycles vs. versus detect cycles more explicit?

Also I think we should record all visited urls anyways.

fabianfett: Oh do we want to use String here instead of URL?

Sources/AsyncHTTPClient/RedirectState.swift
```diff
@@ -0,0 +1,134 @@
...
+            return nil
+        case .follow(let maxRedirects, let allowCycles):
+            self.init(count: maxRedirects, visited: allowCycles ? nil : [])
+        }
+    }
+}
+
+extension RedirectState {
+    /// Call this method when you are about to do a redirect to the given `redirectURL`.
+    /// This method records that URL into `self`.
```
fabianfett, line 42: This only happens sometimes. As mentioned above I think we should always record.

Sources/AsyncHTTPClient/RedirectState.swift
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        self.count -= 1
+        self.visited?.insert(redirectURL)
+    }
+}
+
+extension HTTPHeaders {
+    /// Tries to extract a redirect URL from the `location` header if the `status` indicates it should do so.
+    /// It also validates that we can redirect to the scheme of the extracted redirect URL  from the `originalScheme`.
```
fabianfett, line 61:
```suggestion
    /// It also validates that we can redirect to the scheme of the extracted redirect URL from the `originalScheme`.
```

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -882,10 +882,41 @@ class HTTPClientTests: XCTestCase {
             XCTAssertNoThrow(try localHTTPBin.shutdown())
         }
 
-        XCTAssertThrowsError(try localClient.get(url: "https://localhost:\(localHTTPBin.port)/redirect/infinite1").wait(), "Should fail with redirect limit") { error in
+        XCTAssertThrowsError(try localClient.get(url: "https://localhost:\(localHTTPBin.port)/redirect/infinite1").timeout(after: .seconds(10)).wait(), "Should fail with redirect limit") { error in
```
fabianfett, line 885:
```suggestion
        XCTAssertThrowsError(try localClient.get(url: "https://localhost:\(localHTTPBin.port)/redirect/infinite1").timeout(after: .seconds(10)).wait()) { error in
```
add a small comment, what we need the timeout for. I suspect terminating the test in case the circle isn't detected.

Sources/AsyncHTTPClient/RedirectState.swift
```diff
@@ -0,0 +1,140 @@
...
+    body requestBody: Body?,
+    to redirectURL: URL,
+    status responseStatus: HTTPResponseStatus
+) -> (HTTPMethod, HTTPHeaders, Body?) {
+    var convertToGet = false
+    if responseStatus == .seeOther, requestMethod != .HEAD {
+        convertToGet = true
+    } else if responseStatus == .movedPermanently || responseStatus == .found, requestMethod == .POST {
+        convertToGet = true
+    }
```
fabianfett, line 120: I know that this is just the code that existed before. @Lukasa does that make any sense?

Lukasa: Yup, that's exactly right, see [step 12 in this algorithm](https://fetch.spec.whatwg.org/#http-redirect-fetch).

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -543,22 +543,23 @@ public class HTTPClient {
...
             var request = request
             if request.redirectState == nil {
-                request.redirectState = .init(count: max, visited: allowCycles ? nil : Set())
+                request.redirectState = RedirectState(
+                    self.configuration.redirectConfiguration.configuration,
+                    initialURL: request.url.absoluteString
+                )
             }
-            redirectHandler = RedirectHandler<Delegate.Response>(request: request) { newRequest in
+            guard request.redirectState != nil else { return nil }
```
Lukasa, line 554: I find this pattern pretty oblique: it's not obvious to me that having no redirect state means "don't follow redirects". Can we use a computed property on `request` to make it clearer?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -644,84 +639,34 @@ internal struct RedirectHandler<ResponseType> {
...
-        }
-
+    func redirect(
+        status: HTTPResponseStatus,
+        to redirectURL: URL,
+        promise: EventLoopPromise<ResponseType>
+    ) {
         do {
+            var redirectState = self.request.redirectState
+            try redirectState?.redirect(to: redirectURL.absoluteString)
```
Lukasa, line 657: Is it acceptable for `redirectState` to be `nil` here?

Sources/AsyncHTTPClient/RedirectState.swift
```diff
@@ -0,0 +1,140 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2018-2021 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/AsyncHTTPClient/RedirectState.swift
```diff
@@ -0,0 +1,140 @@
...
+    /// if true, `redirect(to:)` will throw an error if a cycle is detected.
+    private let allowCycles: Bool
+}
+
+extension RedirectState {
+    /// Creates a `RedirectState` from a configuration.
+    /// Returns nil if the user disallowed redirects,
+    /// otherwise an instance of `RedirectState` which respects the user defined settings.
+    init?(
+        _ configuration: HTTPClient.Configuration.RedirectConfiguration.Configuration,
```
Lukasa, line 34: This type name is _awful_. Is it possible we could give some of these types clearer names? If not, can we use a `typealias`?

Sources/AsyncHTTPClient/RedirectState.swift
```diff
@@ -0,0 +1,140 @@
...
+    mutating func redirect(to redirectURL: String) throws {
+        guard self.count > 0 else {
+            throw HTTPClientError.redirectLimitReached
+        }
+
+        if !allowCycles && self.visited.contains(redirectURL) == true {
+            throw HTTPClientError.redirectCycleDetected
+        }
+
+        self.count -= 1
```
Lukasa, line 60: Should we store `count` or just derive it from `self.visited`?

Sources/AsyncHTTPClient/RedirectState.swift
dnadoba: sadly `HTTPClient.Configuration.RedirectConfiguration` is public API but we can change the last `Configuration` to e.g. `Mode`. In addition, we can add a typealias:
```swift
internal typealias RedirectMode = HTTPClient.Configuration.RedirectConfiguration.Mode
```
WDYT?

Sources/AsyncHTTPClient/RedirectState.swift
dnadoba: `self.visited.count` counts up, `self.count` counts down until it hits zero where we know we reached the limit.
Because we need to know when we reach the limit, `self.visited` wouldn't be enough and we would still need to store the limit. Nevertheless, I think it's clearer to only count up and compare `self.visited.count` agains the constant limit to deduplicate counting.

Sources/AsyncHTTPClient/HTTPHandler.swift
dnadoba: No, I have removed `redirectState` from `Request` and moved it to `RedirectHandler` itself. We now explicitly forward the `RedirectState` and don't carry it implicitly through the Request.

Sources/AsyncHTTPClient/HTTPClient.swift
dnadoba: `redirectState` is now no longer a property on `request`. It never really belonged there in the first place and is now passed explicitly to the RedirectHandler. I also didn't like that `nil` means don't follow redirects but now that we explicitly forward it is a bit better.

If we still want to make it more explicit, I think we need to make `RedirectState` aware of the "don't follow redirects" case and ask it before we create a `RedirectHandler`.

Sources/AsyncHTTPClient/RedirectState.swift
Lukasa: I'd like both of those changes, great idea!

Lukasa (approved):

fabianfett (approved):

PR #524, opened 2021-12-13
async/await execute

dnadoba:
This is the last missing piece to actually execute a `HTTPClientRequest` and await a `HTTPClientResponse`.
It brings everything together but first preparing the `HTTPClientRequest` and then executing the `Transaction` using the `HTTPConnectionPool.Manager`. We optional follow redirects if the configurations allows it and the `HTTPClientRequest.Body` can be consumed multiple times.

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
@@ -51,6 +51,8 @@ extension HTTPClient {
     ) async throws -> HTTPClientResponse {
         var currentRequest = request
         var currentRedirectState = redirectState
+
+        // this loop is there to follow potential redirects
         while true {
             let preparedRequest = try HTTPClientRequest.Prepared(currentRequest)
             let response = try await executeCancellable(preparedRequest, deadline: deadline, logger: logger)
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
@@ -49,7 +49,7 @@ extension HTTPClient {
         redirectState: RedirectState?
     ) async throws -> HTTPClientResponse {
         let preparedRequest = try HTTPClientRequest.Prepared(request)
-        let response = try await execute(request, deadline: deadline, logger: logger)
+        let response = try await executeWithoutFollowingRedirects(preparedRequest, deadline: deadline, logger: logger)
         guard
             preparedRequest.body.canBeConsumedMultipleTimes,
             let redirectState = redirectState,
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
@@ -107,7 +107,8 @@ extension HTTPClient {
                     preferredEventLoop: self.eventLoopGroup.next(),
                     responseContinuation: continuation
                 )
-
+                
+                // `HTTPClient.Task` conflicts with Swift Concurrency Task and `Swift.Task` doesn't work
                 _Concurrency.Task {
                     await cancelHandler.registerTransaction(transaction)
                 }
@@ -115,6 +116,7 @@ extension HTTPClient {
                 self.poolManager.executeRequest(transaction)
             }
         }, onCancel: {
+            // `HTTPClient.Task` conflicts with Swift Concurrency Task and `Swift.Task` doesn't work
             _Concurrency.Task {
                 await cancelHandler.cancel()
             }
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
@@ -50,7 +50,7 @@ extension HTTPClient {
         redirectState: RedirectState?
     ) async throws -> HTTPClientResponse {
         let preparedRequest = try HTTPClientRequest.Prepared(request)
-        let response = try await executeWithoutFollowingRedirects(preparedRequest, deadline: deadline, logger: logger)
+        let response = try await executeCancellable(preparedRequest, deadline: deadline, logger: logger)
 
         if !preparedRequest.body.canBeConsumedMultipleTimes,
            let redirectState = redirectState,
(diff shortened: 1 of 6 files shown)


Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -0,0 +1,175 @@
...
+            preparedRequest.body.canBeConsumedMultipleTimes,
+            let redirectState = redirectState,
+            let redirectURL = response.headers.extractRedirectTarget(
+                status: response.status,
+                originalURL: preparedRequest.url,
+                originalScheme: preparedRequest.poolKey.scheme
+            )
+        else {
+            return response
+        }
```
fabianfett, line 63: I think this would be easier to read as an `if`.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -0,0 +1,175 @@
...
+            logger: logger,
+            redirectState: redirectState
+        )
+    }
+
+    private func executeWithoutFollowingRedirects(
+        _ request: HTTPClientRequest.Prepared,
+        deadline: NIODeadline,
+        logger: Logger
+    ) async throws -> HTTPClientResponse {
```
fabianfett, line 109: I can't find where this method is used.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -0,0 +1,175 @@
...
+            response: response,
+            deadline: deadline,
+            logger: logger
+        )
+    }
+
+    private func followRedirect(
+        redirectURL: URL,
+        redirectState: RedirectState,
+        request: HTTPClientRequest.Prepared,
```
fabianfett, line 76: the naming of request feels ambiguous here.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -0,0 +1,173 @@
...
+                case .register(let bag):
+                    self.state = .cancelled
+                    bag.cancel()
+                case .cancelled:
+                    break
+                case .initialised:
+                    self.state = .cancelled
+                }
+            }
+        }
```
fabianfett, line 144: I think we should probably pull this one out of this function... just put it on the file end with `private` access.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -0,0 +1,165 @@
...
+                let transaction = Transaction(
+                    request: request,
+                    requestOptions: .init(idleReadTimeout: nil),
+                    logger: logger,
+                    connectionDeadline: deadline,
+                    preferredEventLoop: self.eventLoopGroup.next(),
+                    responseContinuation: continuation
+                )
+
+                _Concurrency.Task {
```
fabianfett, line 111: Comment re naming conflict on `Task`.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -0,0 +1,166 @@
...
+@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+extension HTTPClient {
+    private func executeAndFollowRedirectsIfNeeded(
+        _ request: HTTPClientRequest,
+        deadline: NIODeadline,
+        logger: Logger,
+        redirectState: RedirectState?
+    ) async throws -> HTTPClientResponse {
+        let preparedRequest = try HTTPClientRequest.Prepared(request)
+        let response = try await executeWithoutFollowingRedirects(preparedRequest, deadline: deadline, logger: logger)
```
fabianfett, line 53: This method has nothing really to do with followingRedirects, but cancellation is its feature. WDYT about `executeCancellable`?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -0,0 +1,166 @@
...
+                _Concurrency.Task {
+                    await cancelHandler.registerTransaction(transaction)
+                }
+
+                self.poolManager.executeRequest(transaction)
+            }
+        }, onCancel: {
+            // `HTTPClient.Task` conflicts with Swift Concurrency Task and `Swift.Task` doesn't work
+            _Concurrency.Task {
+                await cancelHandler.cancel()
```
Lukasa, line 121: Do we need `cancel` to be awaitable? Constructing these child tasks is a bit awkward.

dnadoba: We don't need it to be awaitable but we use an actor as our synchronisation primitive so all methods are automatically async/awaitable from outside. 
I see two options:
1. We could make the current `cancel()` method private (and rename it) and add a new `nonisolated cancel()` method which would internal create the `Task` and call the `private cancel()` method.
2. We could also rewrite `TransactionCancelHandler` to use a lock instead and make it a class.

WDYT?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -0,0 +1,166 @@
...
+        let preparedRequest = try HTTPClientRequest.Prepared(request)
+        let response = try await executeCancellable(preparedRequest, deadline: deadline, logger: logger)
+
+        if !preparedRequest.body.canBeConsumedMultipleTimes,
+           let redirectState = redirectState,
+           let redirectURL = response.headers.extractRedirectTarget(
+               status: response.status,
+               originalURL: preparedRequest.url,
+               originalScheme: preparedRequest.poolKey.scheme
+           ) {
```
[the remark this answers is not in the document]
dnadoba: Good catch! We now check if the body of the new request can be send multiple times instead of looking at the original body. If we drop the body for the new request, the body will be nil and `canBeConsumedMultipleTimes` will return `true`.

dnadoba:

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
Lukasa: We can leave it as-is, I guess. I'm always a bit nervous about creating lots of tasks but we can always change it later.

fabianfett (approved):
Awesome! Thanks!

PR #526, opened 2021-12-14
Reduce unnecessary thread switches

dnadoba:
Reduce unnecessary thread switches by using the current `EventLoop` if possible.


fabianfett (approved):
awesome fix

PR #527, opened 2021-12-14
Make async/await available on older Apple Platforms

dnadoba:
### Motivation
With Xcode 13.2, and therefore Swift 5.5.2, Swift Concurrency is supported on older Apple OSs. async/await support will no longer be available on Swift before `5.5.2` but this isn't a breaking change because we have not yet made anything of it public.

### Changes
- replace all `#if compiler(>=5.5) && canImport(_Concurrency)` with `#if compiler(>=5.5.2) && canImport(_Concurrency)`
- replace all `available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)` with `available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)`

### DIFF
--- a/Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
+++ b/Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(>=5.5) && canImport(_Concurrency)
+#if compiler(>=5.5.2) && canImport(_Concurrency)
 import NIOConcurrencyHelpers
 import NIOCore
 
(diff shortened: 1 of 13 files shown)


fabianfett (approved):
LGTM. @glbrntt do you mind giving this a look?

Tests/AsyncHTTPClientTests/AsyncTestHelpers.swift
```diff
@@ -16,7 +16,7 @@
 import NIOConcurrencyHelpers
 import NIOCore
 
-@available(macOS 12.0, iOS 15.0, watchOS 8.0, tvOS 15.0, *)
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```
glbrntt, line 19: Only tests, but presumably the `#if swift` above needs to be updated too?

glbrntt (approved):
LGTM

PR #528, opened 2021-12-15
Add end-to-end tests for async/await

dnadoba:


Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -0,0 +1,441 @@
...
+            let port = server.localAddress!.port!
+
+            let request = HTTPClientRequest(url: "https://localhost:\(port)")
+
+            await XCTAssertThrowsError(try await client.execute(request, deadline: .now() + .seconds(1), logger: logger)) { error in
+                guard let httpError = error as? HTTPClientError else {
+                    return XCTFail("wrong error \(error)")
+                }
+                XCTAssertTrue(
+                    [HTTPClientError.connectTimeout, .tlsHandshakeTimeout].contains(httpError),
```
fabianfett, line 407: I don't think `tlsHandshakeTimeout` is a possible error here. We will never have an actual connection here, so tlsHandshake will never be started.

fabianfett (approved):
One small nit.

PR #529, opened 2021-12-16
Respect deadline on new HTTPClient.execute for async/await 

dnadoba:
We forgot to schedule a timeout timer for the user defined deadline and just use it as a connection timeout.


fabianfett (approved):
Great fix!

PR #530, opened 2021-12-17
Fix HTTP1 to HTTP2 migration while shutdown is in progress

dnadoba:
### Motivation
Calling `HTTPClient.shutdown()` may never return if connections are still starting and one new established connection results in a state migration (i.e. from HTTP1 to HTTP2 or vice versa). We forgot to migrate the shutdown state. This could result in a large delay until `.shutdown()` returns because we wait until connections are closed because of idle timeout. Worse, it could also never return if more requests are queued because the connections would not be idle and therefore not close itself.
### Changes
- Mirgrate shutdown state too
- add tests for this specific case

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPConnectionPool+HTTP2StateMachineTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPConnectionPool+HTTP2StateMachineTests.swift
@@ -680,40 +680,21 @@ class HTTPConnectionPool_HTTP2StateMachineTests: XCTestCase {
         var queuer = MockRequestQueuer()
         var state = HTTPConnectionPool.StateMachine(idGenerator: .init(), maximumConcurrentHTTP1Connections: 8)
 
-        /// first 8 request should create a new connection
-        var connectionIDs: [HTTPConnectionPool.Connection.ID] = []
-        for _ in 0..<8 {
-            let mockRequest = MockHTTPRequest(eventLoop: el1)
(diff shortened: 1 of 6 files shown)


Tests/AsyncHTTPClientTests/HTTPConnectionPool+HTTP2StateMachineTests.swift
```diff
@@ -670,6 +673,89 @@ class HTTPConnectionPool_HTTP2StateMachineTests: XCTestCase {
         XCTAssertTrue(queuer.isEmpty)
     }
 
+    func testMigrationFromHTTP1ToHTTP2WhileShuttingDown() {
```
fabianfett, line 676: I think this test case is a little to complex... let's just focus on the thing that we actually observed.

1. run request -> opens new connection (expect h1)
2. shutdown pool -> queued request shall be cancelled
3. connection creation is done as h2 -> connection shall be closed immediately and shutdown is done.

then we should also have this vica versa.

1. run request -> opens new connection
2. connection creation is done as h2 -> schedule request on h2 connection
3. connection is done with the single request
4. idle timeout closes h2 connection
5. new request comes in. we create a new connection (expect h2)
6. shutdown pool -> cancel queued request
7. connection creation is done as h1 -> connection shall be closed immediately and shutdown is done.

fabianfett (approved):
Looks awesome! Thanks!

PR #531, opened 2021-12-17
Prepare async/await API for public release

dnadoba:
- add nescary @inlinable/@usableFromInlinable for generic `HTTPClientReuqest.Body` builder metods.
- rename `HTTPClientReuqest.Body.byteBuffer(_:)` to `HTTPClientReuqest.Body.bytes(_:)` to be more in line with the other overloads
- add a new type called `HTTPClientReuqest.Body.Length` instead of `Int?` to specify the length of a body if we can't infer the length
- move `length` parameter to the end

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
@@ -54,38 +54,9 @@ extension HTTPClientRequest {
 
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 extension HTTPClientRequest.Body {
-    public static func byteBuffer(_ byteBuffer: ByteBuffer) -> Self {
+    
+    public static func bytes(_ byteBuffer: ByteBuffer) -> Self {
         self.init(.byteBuffer(byteBuffer))
-    }
-
-    @inlinable
-    public static func bytes<Bytes: Sequence>(
-        length: Int?,
-        _ bytes: Bytes
-    ) -> Self where Bytes.Element == UInt8 {
-        self.init(.sequence(length: length, canBeConsumedMultipleTimes: false) { allocator in
-            if let buffer = bytes.withContiguousStorageIfAvailable({ allocator.buffer(bytes: $0) }) {
-                // fastpath
-                return buffer
-            }
-            // potentially really slow path
-            return allocator.buffer(bytes: bytes)
-        })
-    }
-
-    @inlinable
-    public static func bytes<Bytes: Collection>(
-        length: Int?,
-        _ bytes: Bytes
-    ) -> Self where Bytes.Element == UInt8 {
-        self.init(.sequence(length: length, canBeConsumedMultipleTimes: true) { allocator in
-            if let buffer = bytes.withContiguousStorageIfAvailable({ allocator.buffer(bytes: $0) }) {
-                // fastpath
-                return buffer
-            }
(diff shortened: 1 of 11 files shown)


Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -48,25 +48,22 @@ struct HTTPClientResponse {
...
 
+    public struct AsyncIterator: AsyncIteratorProtocol {
         private let stream: IteratorStream
 
         fileprivate init(stream: IteratorStream) {
             self.stream = stream
         }
 
-        func next() async throws -> ByteBuffer? {
+        public func next() async throws -> ByteBuffer? {
```
glbrntt, line 60: Question rather than suggestion: should this be `mutating` to indicate that `next()` consumes from the stream?

Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
```diff
@@ -12,9 +12,16 @@
 //
 //===----------------------------------------------------------------------===//
 
-enum RequestBodyLength: Hashable {
+public struct RequestBodyLength {
```
fabianfett, line 15: We should not add this type without prefix to the top namespace.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
dnadoba: Makes sense to me. @fabianfett any objections?

Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
dnadoba: Agreed. I have reverted the change and added a new public type `HTTPClientRequest.Body.Length` which is baked by `RequestBodyLength` and used as the public type instead.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -143,4 +155,19 @@ extension Optional where Wrapped == HTTPClientRequest.Body {
     }
 }
 
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientRequest.Body {
+    public struct Length {
+        /// size of the request body is not known before starting the request
+        public static let dynamic: Self = .init(storage: .dynamic)
```
fabianfett, line 162: Is dynamic the right name here? What about `unknown`?
I like the storage workaround btw.

dnadoba: works for me

fabianfett: btw. fixed is also weird. I guess `known` would be more appropriate. Also I think this type should be `ExpressibleByIntegerLiteral`.

dnadoba:

fabianfett (approved):

glbrntt (approved):
LGTM!

Sources/AsyncHTTPClient/ConnectionPool/RequestBodyLength.swift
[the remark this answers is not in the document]
glbrntt: (Ignore these, they were from an old review.)

PR #534, opened 2021-12-19
Make Body.data generic

karwa:
I don't think there is a need for this specific API to use Foundation.Data -- we can broaden it to all RandomAccessCollections . RAC seems like the most appropriate constraint given the up-front `count` calculation.

This removes a Foundation dependency, and should be source-compatible.

### DIFF
--- a/Sources/AsyncHTTPClient/FoundationExtensions.swift
+++ b/Sources/AsyncHTTPClient/FoundationExtensions.swift
@@ -54,3 +54,13 @@ extension HTTPClient.Cookie {
         )
     }
 }
+
+extension HTTPClient.Body {
+    /// Create and stream body using `Data`.
+    ///
+    /// - parameters:
+    ///     - bytes: Body `Data` representation.
+    public static func data(_ data: Data) -> HTTPClient.Body {
+        return self.bytes(data)
+    }
+}
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -49,6 +49,12 @@ extension HTTPClient {
         /// Body chunk provider.
         public var stream: (StreamWriter) -> EventLoopFuture<Void>
 
+        @inlinable
+        init(length: Int?, stream: @escaping (StreamWriter) -> EventLoopFuture<Void>) {
+            self.length = length
+            self.stream = stream
+        }
+
         /// Create and stream body using `ByteBuffer`.
         ///
         /// - parameters:
@@ -69,13 +75,14 @@ extension HTTPClient {
             return Body(length: length, stream: stream)
         }
 
-        /// Create and stream body using `Data`.
+        /// Create and stream body using a collection of bytes.
         ///
         /// - parameters:
-        ///     - data: Body `Data` representation.
-        public static func data(_ data: Data) -> Body {
-            return Body(length: data.count) { writer in
-                writer.write(.byteBuffer(ByteBuffer(bytes: data)))
+        ///     - data: Body binary representation.
+        @inlinable
+        public static func bytes<Bytes>(_ bytes: Bytes) -> Body where Bytes: RandomAccessCollection, Bytes.Element == UInt8 {
+            return Body(length: bytes.count) { writer in
+                writer.write(.byteBuffer(ByteBuffer(bytes: bytes)))
             }
         }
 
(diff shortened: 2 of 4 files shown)


dnadoba:
Sadly this is a source breaking change because we can create an unapplied reference to this method e.g.:
```swift
let makeBodyFromData = HTTPClient.Body.data
```
After your change this no longer compiles with the error message "Generic parameter 'Bytes' could not be inferred".

What we can do instead is leaving the old method signature as is and adding your new method as an additional overload.

karwa:
@dnadoba Okay, I've added the `Data` version back as you suggested. It's a little awkward, but the idea is that it forwards to the generic version to avoid duplicating implementations.

I've also added a test to make double-sure that unapplied references to `HTTPClient.Body.data` continues to resolve to the `Data` version.

Sources/AsyncHTTPClient/FoundationExtensions.swift
```diff
@@ -54,3 +54,16 @@ extension HTTPClient.Cookie {
...
     }
 }
+
+extension HTTPClient.Body {
+    /// Create and stream body using `Data`.
+    ///
+    /// - parameters:
+    ///     - bytes: Body `Data` representation.
+    public static func data(_ bytes: Data) -> HTTPClient.Body {
+        func forwardToGenericImpl<C>(_ value: C) -> HTTPClient.Body where C: RandomAccessCollection, C.Element == UInt8 {
```
dnadoba, line 64: @Lukasa do you know if the compiler will specialise this local function?

dnadoba (commented):
Thanks a lot, great patch! @fabianfett WDYT?

Sources/AsyncHTTPClient/FoundationExtensions.swift
karwa: It appears to: https://swift.godbolt.org/z/WKeeG5r7v

Lukasa: As @karwa says, yes, the compiler should absolutely specialise this function. However, it's a pretty ugly thunk to have to have, so I'd like to ask whether we can differentiate between this method and the other method called `data` by renaming the other one. That way we don't need to write this thunk out explicitly.

karwa: Hmm, well the [async API](https://github.com/swift-server/async-http-client/blob/972bcddedc1e7a4333c77fa4445a9e2798ff6a9f/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift#L89) seems to use `bytes` for this. So we could use that.

I went with `data` to try and remove the Foundation.Data dependency whilst still supporting Data. But if we have to add the extension for source compatibility, that reasoning isn't quite as strong. Then again, there is value in having the same thing be spelled the same way, and it's the API we have, so... ⚖️ . The thunk is not going to be winning any beauty awards, for sure, but is it less ugly than having two spellings for essentially the same operation? I'm not sure.

If you really think it's worth it, I could rename it to `bytes`.

Lukasa: In general I'm disinclined to rely on type-based method overloads. We've had a bunch of problems with this in the past where having methods whose name is the same that differ only based on type leans excessively heavily on the type checker. If these differed only on return value I'd outright refuse to accept them having the same name: differing based on argument type is less bad, but in this instance I think I'd still rather use a different name. There's existing prior art in NIO for using the word `data` in method names to mean "accepts/returns `Foundation.Data`" and `bytes` to mean "accepts/returns something else", so I think we're in good company here.

karwa: Oh, I wasn't aware that pattern was used elsewhere in NIO. I've changed it to `.bytes` now.

Sources/AsyncHTTPClient/FoundationExtensions.swift
```diff
@@ -54,3 +54,13 @@ extension HTTPClient.Cookie {
...
     }
 }
+
+extension HTTPClient.Body {
+    /// Create and stream body using `Data`.
+    ///
+    /// - parameters:
+    ///     - bytes: Body `Data` representation.
+    public static func data(_ data: Data) -> HTTPClient.Body {
+        return bytes(data)
```
Lukasa, line 64: Yeah, I know this is a ridiculous nit, but:

```suggestion
        return self.bytes(data)
```

Lukasa (approved):
One tiny nit and then we're good.

Sources/AsyncHTTPClient/FoundationExtensions.swift
karwa: Done

PR #539, opened 2022-01-10
Used vendored base64 encoding implementation rather than Foundation.Data

karwa:
Uses @fabianfett's base64 implementation rather than Foundation. Apparently it's faster, and it avoids copying the string to a Foundation.Data object before encoding.

Single-file version copied from https://github.com/apple/swift-nio/blob/main/Sources/NIOWebSocket/Base64.swift

I disabled swiftformat on the file so that, other than the copyright header and swiftformat comment, it remains identical to the version in NIO. Alternatively, we could add a package dependency rather than copying the file. Up to you guys.

With this and #534, the only remaining Foundation dependencies will be `URL`, and the couple of convenience APIs in the `FoundationExtensions` file. If they're both accepted, I can go through and replace all occurrences of `import Foundation` with `import struct Foundation.URL` in a follow-up.

There are already tests for this API, so I didn't feel it necessary to add any.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -53,3 +53,10 @@ This product contains a derivation of the Tony Stone's 'process_test_files.rb'.
     * https://github.com/tonystone/build-tools/blob/master/source/xctest_tool.rb
 
 ---
+
+This product contains a derivation of Fabian Fett's 'Base64.swift'.
+
+  * LICENSE (Apache License 2.0):
+    * https://github.com/fabianfett/swift-base64-kit/blob/master/LICENSE
+  * HOMEPAGE:
+    * https://github.com/fabianfett/swift-base64-kit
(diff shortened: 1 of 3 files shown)


Lukasa (approved):
Seems reasonable to me, @fabianfett?

karwa:
Hey @fabianfett, mind taking a look at this when you get time? Thanks!

fabianfett (commented):
Sorry for taking so long to review. The code change looks good. @Lukasa how do we deal with adding a Note to the Notice file, as the original repo does not exist anymore?

Lukasa:
Copy the note from NIO.

fabianfett:
@karwa Do you mind copying the note from NIO's [NOTICE file](https://github.com/apple/swift-nio/blob/398378653f97ce695d2ca1279e7ed36a9598335d/NOTICE.txt#L75-L80)?

karwa:
@fabianfett done

fabianfett (approved):

fabianfett:
Failing nightly-main and nightly-5.6 API because of missing Sendable requirements.

PR #545, opened 2022-01-24
Print invalid state, if hitting precondition

fabianfett:
If we hit a `precondition` because of an invalid state, we should print the state to make debugging easier.


FranzBusch (approved):
LGTM

dnadoba (approved):

fabianfett:
CI failing on nightly-5.6 and nightly-main because of Sendable requirements.

PR #546, opened 2022-01-24
Fix race between connection close and scheduling new request

fabianfett:
### Motivation

We may have a race condition between scheduling a request on an http/1.1 connection and the connection being closed.

### Changes

- Fail request that is scheduled on a closed connection

### Result

We don't crash anymore.


weissi (approved):
thank you!

Lukasa (approved):

PR #549, opened 2022-02-01
Disable logging by default and add timeout API

dnadoba:
This PR makes the `logger` parameter of `execute` optional and will use a no-op logger if nil.
In addition, it also adds a new `execute` method which takes a `timeout` instead of a `deadline`.


dnadoba:

fabianfett (approved):

PR #551, opened 2022-02-04
Add `async` shutdown method to `HTTPClient`

dnadoba:
Allows to shutdown a `HTTPClient` in an async context without blocking or using the callback version of shutdown.


Lukasa (approved):
Ace, LGTM.

fabianfett (approved):

PR #553, opened 2022-02-07
Add async/await examples

dnadoba:
This PR adds examples which make use of the new async/await APIs. Sadly, async/await is not supported in top level code [(but there is a pitch)](https://forums.swift.org/t/concurrency-in-top-level-code/55001). Just spinning up a task is not enough, we also need to wait for its completion. The new `@main` attribute already supports async/await and automatically waits for completion but we need to mark the target as an `executableTarget`. `executableTargets` were introduced in Swift 5.4 and we therefore need swift-tools-version 5.4. As we still support Swift 5.2 and Swift 5.3 we sadly need to create separate Package.swift files. 

Alternatively, we could also use a semaphore to wait until completion. But as this is an example, we should be a role model.

Note: this PR requires the newest unreleased version of `swift-nio`. We can only merge it after a new release of `swift-nio`.

### DIFF
--- a/Examples/GetHTML/GetHTML.swift
+++ b/Examples/GetHTML/GetHTML.swift
@@ -1,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2022 Apple Inc. and the AsyncHTTPClient project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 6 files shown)


dnadoba:
@fabianfett suggested to create a separate Package.swift in the Examples folders which I will do shortly.

Examples/GetHTML/GetHTML.swift
```diff
@@ -0,0 +1,40 @@
...
+//===----------------------------------------------------------------------===//
+
+// TODO: remove @testable after async/await API is public
+@testable import AsyncHTTPClient
+import NIOCore
+
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+
+@main
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```
fabianfett, line 22: I think we should put those platform requirements in package itself.

fabianfett (approved):

PR #554, opened 2022-02-07
Update README.md for async/await

dnadoba:
We want to highlight the new async/await API in the README. 
- add new examples for async/await but keeps old examples for SwiftNIOs `EventLoopFuture`s
- change URL used in examples from swift.org to apple.com because swift.org doesn't respect that we don't support compression by default which results in gibberish output if examples are run by users.
- increase recommended Xcode Version to 13.2 and Swift Version to 5.5.2
- increase recommend AsyncHTTPClient version to `1.9.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 This package provides simple HTTP Client library built on top of SwiftNIO.
 
 This library provides the following:
+- First class support for Swift Concurrency (since version 1.9.0)
 - Asynchronous and non-blocking request methods
 - Simple follow-redirects (cookie headers are dropped)
 - Streaming body download
@@ -11,7 +12,7 @@ This library provides the following:
 
 ---
 
-**NOTE**: You will need [Xcode 11.4](https://apps.apple.com/gb/app/xcode/id497799835?mt=12) or [Swift 5.2](https://swift.org/download/#swift-52) to try out `AsyncHTTPClient`.
+**NOTE**: You will need [Xcode 13.2](https://apps.apple.com/gb/app/xcode/id497799835?mt=12) or [Swift 5.5.2](https://swift.org/download/#swift-552) to try out `AsyncHTTPClient`s new async/await APIs.
 
 ---
 
@@ -21,7 +22,7 @@ This library provides the following:
 Add the following entry in your <code>Package.swift</code> to start using <code>HTTPClient</code>:
 
 ```swift
-.package(url: "https://github.com/swift-server/async-http-client.git", from: "1.0.0")
+.package(url: "https://github.com/swift-server/async-http-client.git", from: "1.9.0")
 ```
 and  `AsyncHTTPClient` dependency to your target:
 ```swift
@@ -40,7 +41,21 @@ If your application does not use SwiftNIO yet, it is acceptable to use `eventLoo
 import AsyncHTTPClient
 
 let httpClient = HTTPClient(eventLoopGroupProvider: .createNew)
-httpClient.get(url: "https://swift.org").whenComplete { result in
+
+/// MARK: - Using Swift Concurrency
+let request = HTTPClientRequest(url: "https://apple.com/")
+let response = try await httpClient.execute(request, timeout: .seconds(30))
+print("HTTP head", response)
+if response.status == .ok {
+    let body = try await response.body.collect(upTo: 1024 * 1024) // 1 MB
+    // handle body
+} else {
+    // handle remote error
+}
+
+
+/// MARK: - Using SwiftNIO EventLoopFuture
+httpClient.get(url: "https://apple.com/").whenComplete { result in
     switch result {
     case .failure(let error):
         // process error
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,7 +107,7 @@ defer {
     try? httpClient.syncShutdown()
 }
 
-var request = try HTTPClient.Request(url: "https://swift.org", method: .POST)
+var request = try HTTPClient.Request(url: "https://apple.com/", method: .POST)
```
fabianfett, line 110: Why was this one changed?

README.md
```diff
@@ -105,7 +145,38 @@ httpClient.execute(request: request, deadline: .now() + .milliseconds(1))
...
+for try await buffer in response.body {
+    // For this example, we are just interested in the size of the fragment
+    receivedBytes += buffer.readableBytes
+    
+    if let expectedBytes = expectedBytes {
+        // if the body size is known, we calculate a progress indicator
+        let progress = Double(receivedBytes)/Double(expectedBytes)
+        print("progress: \(Int(progress * 100))%")
+    }
+    // in case backpressure is needed, all reads will be paused until the end of the for loop.
```
fabianfett, line 173: backpressure has nothing todo with the loop, and I don't think we should explain it here. It just works ;)

README.md
```diff
@@ -105,7 +145,38 @@ httpClient.execute(request: request, deadline: .now() + .milliseconds(1))
...
+}
+print("did receive \(receivedBytes) bytes")
```
fabianfett, line 175: close AHC?

README.md
dnadoba: I think we should mentioned this at least somewhere. AFAIK, this is not documented anywhere or am I missing something? This is a really great feature and we should definitely mention it.

README.md
dnadoba: swift.org doesn't respect that we don't support compression by default which results in gibberish responses

README.md
dnadoba: The full example has the shutdown(): https://github.com/swift-server/async-http-client/blob/3405d11bf46e0be3130aa318cceefa1941752618/Examples/StreamingByteCounter/StreamingByteCounter.swift#L25-L52
We would need to nest it in a do catch to be correct but I'm fine with it if you want to add it. I didn't add it because the old one also has no shutdown. However, I'm just discovering that the old examples doesn't create an AsyncHTTPClient, it just assumes that there is one. I see two options:
- remove HTTPClient creation
- or add shutdown
WDYT?

README.md
```diff
@@ -40,7 +41,21 @@ If your application does not use SwiftNIO yet, it is acceptable to use `eventLoo
...
+/// MARK: - Using Swift Concurrency
+let request = HTTPClientRequest(url: "https://apple.com/")
+let response = try await httpClient.execute(request, timeout: .seconds(30))
+print("HTTP head", response)
+if response.status == .ok {
+    let body = try await response.body.collect(upTo: 1024 * 1024) // 1 MB
+    // handle body
+} else {
+    // handle remote error
+}
```
Lukasa, line 54: So much nicer! :tada:

README.md
```diff
@@ -58,7 +73,32 @@ You should always shut down `HTTPClient` instances you created using `try httpCl
 
 ## Usage guide
 
-Most common HTTP methods are supported out of the box. In case you need to have more control over the method, or you want to add headers or body, use the `HTTPRequest` struct:
+The default HTTP Method is `GET`. In case you need to have more control over the method, or you want to add headers or body, use the `HTTPClientRequest` struct:
+#### Using Swift Concurrency
+```swift
```
Lukasa, line 78: As a mild nit, a bit more whitespace around this would make it read nicer when we look at it in plaintext.

README.md
Lukasa: I'm with @dnadoba that we should mention it, but I'm with @fabianfett that I don't think we need to explain it: it just works. In this case, maybe at the top of the loop I'd change that comment to something like:

```
// asynchronously iterates over all body fragments
// this loop will automatically propagate backpressure correctly
```

README.md
Lukasa: I think add the shutdown and nest in a do catch.

fabianfett (approved):
Awesome!

README.md
```diff
@@ -11,7 +12,7 @@ This library provides the following:
 
 ---
 
-**NOTE**: You will need [Xcode 11.4](https://apps.apple.com/gb/app/xcode/id497799835?mt=12) or [Swift 5.2](https://swift.org/download/#swift-52) to try out `AsyncHTTPClient`.
+**NOTE**: You will need [Xcode 13.2](https://apps.apple.com/gb/app/xcode/id497799835?mt=12) or [Swift 5.5.2](https://swift.org/download/#swift-552) to try out `AsyncHTTPClient`s new async/await APIs.
```
glbrntt, line 15: Should this note be added in addition to the previous one?

README.md
```diff
@@ -105,7 +148,43 @@ httpClient.execute(request: request, deadline: .now() + .milliseconds(1))
...
+    var receivedBytes = 0
+    // asynchronously iterates over all body fragments
+    // this loop will automatically propagate backpressure correctly
+    for try await buffer in response.body {
+        // for this example, we are just interested in the size of the fragment
+        receivedBytes += buffer.readableBytes
+        
+        if let expectedBytes = expectedBytes {
+            // if the body size is known, we calculate a progress indicator
+            let progress = Double(receivedBytes)/Double(expectedBytes)
```
glbrntt, line 174: nit, space makes it more readable:

```suggestion
            let progress = Double(receivedBytes) / Double(expectedBytes)
```

glbrntt (approved):
Looks good, I left two tiny nits though

README.md
dnadoba: I think this is now fairly old and we can remove it. I think (or at least hope) that every new user is already on an Xcode/Swift version that came out in the last 2 years.

glbrntt: True, and we're about to drop support for 5.2 and 5.3 anyway!

PR #558, opened 2022-02-10
Crash fix: HTTP2ClientRequestHandler can deal with failing writes

fabianfett:
### Motivation

AHC can currently not deal with direct failures when writing requests.

### Changes

- Make `HTTP2ClientRequestHandler` resilient to failing writes

### Result

Less crashes in AHC HTTP/2.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
@@ -139,30 +139,13 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
     // MARK: Run Actions
 
     private func run(_ action: HTTPRequestStateMachine.Action, context: ChannelHandlerContext) {
-        // NOTE: We can bang the request in the following actions, since the `HTTPRequestStateMachine`
-        //       ensures, that actions that require a request are only called, if the request is
-        //       still present. The request is only nilled as a response to a state machine action
-        //       (.failRequest or .succeedRequest).
-
         switch action {
         case .sendRequestHead(let head, let startBody):
-            if startBody {
-                context.writeAndFlush(self.wrapOutboundOut(.head(head)), promise: nil)
-                self.request!.requestHeadSent()
-                self.request!.resumeRequestBodyStream()
-            } else {
-                context.write(self.wrapOutboundOut(.head(head)), promise: nil)
-                context.write(self.wrapOutboundOut(.end(nil)), promise: nil)
-                context.flush()
-
-                self.request!.requestHeadSent()
-
-                if let timeoutAction = self.idleReadTimeoutStateMachine?.requestEndSent() {
-                    self.runTimeoutAction(timeoutAction, context: context)
-                }
-            }
+            self.sendRequestHead(head, startBody: startBody, context: context)
 
         case .pauseRequestBodyStream:
+            // We can force unwrap the request here, as we have just validated in the state machine,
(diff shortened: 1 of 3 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -204,13 +198,42 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
...
+    private func sendRequestHead(_ head: HTTPRequestHead, startBody: Bool, context: ChannelHandlerContext) {
+        if startBody {
+            context.writeAndFlush(self.wrapOutboundOut(.head(head)), promise: nil)
+
+            // The above write might trigger an error, which may lead to a call to `errorCaught`,
+            // which in turn, may fail the request and pop it from the handler. For this reason
+            // we must check if the request is still present here.
+            guard let request = self.request else { return }
+            request.requestHeadSent()
+            request.resumeRequestBodyStream()
```
glbrntt, line 219: Is it possible that as a result of writing we still have a request but are _not_ in a state where it would make sense to call either of these? `requestHeadSent()` _seems_ like it should be fine since that's just what we did, it's the `resumeRequestBodyStream()` I'm curious about.

fabianfett: Good question... I guess the big one here would be a change in writability...

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -182,18 +165,29 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
...
+            // We can force unwrap the request here, as we have just validated in the state machine,
+            // that the request is neither failed nor finished yet
             self.request!.receiveResponseHead(head)
-            if pauseRequestBodyStream {
-                self.request!.pauseRequestBodyStream()
+            if pauseRequestBodyStream, let request = self.request {
+                // The above response head forward might lead the request to mark itself as
+                // cancelled, which in turn might pop the request of the handler. For this reason we
+                // must check if the request is still present here.
+                request.pauseRequestBodyStream()
```
karwa, line 180: This all seems very fragile. Each action is very tightly coupled to the state machine to check its own internal state, and some of these methods require non-local reasoning (e.g. calling a method on a request which reaches up and nils that request variable in its owner object) are also fragile.

It's perhaps worth questioning: with the tight coupling, is it worth dispatching all of these operations with a single enum? Perhaps each case should be an individual method, invoked directly with the unwrapped request value. The cancellation stuff is also not ideal but I don't have a concrete suggestion at this time.

Certainly the crash fix is important and I'm _not_ suggesting it be delayed for refactoring, but it's maybe worth making a note to take another look at it later, and see if it can be made more robust.

fabianfett: Very much agree!

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
fabianfett: On second thought. Nope this should be fine. The first is very much just informal, the second one triggers an action... Since the first is informal, no worries! ☮️

weissi (approved):
Thank you

glbrntt (approved):

PR #561, opened 2022-02-15
SSLProtocol tls* deprecations

weissi:
Motivation:

SSLProtocol's tlsProtocol* members are deprecated but we need them in compatibility code (breaks SemVer otherwise).

Modification:

Annotate the legacy function as such.

Result:

Fewer warnings.


Lukasa:
Merging over the 5.6 and main failures, both of which are related to sendable.

PR #566, opened 2022-03-16
Call `didSendRequestPart` at the right time

FranzBusch:
# Motivation
Right now, we call `didSendRequestPart` after passing the write to the executor. However, this does not mean that the write hit the socket. To implement proper backpressure using the delegate, we should only call this method once the write was successful.

# Modification
Pass a promise to the actual channel write and only call the delegate once that promise succeeds.

# Result
The delegate method `didSendRequestPart` is only called after the write was successful.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
@@ -185,11 +185,11 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
         case .sendRequestHead(let head, startBody: let startBody):
             self.sendRequestHead(head, startBody: startBody, context: context)
 
-        case .sendBodyPart(let part):
-            context.writeAndFlush(self.wrapOutboundOut(.body(part)), promise: nil)
+        case .sendBodyPart(let part, let writePromise):
+            context.writeAndFlush(self.wrapOutboundOut(.body(part)), promise: writePromise)
 
-        case .sendRequestEnd:
-            context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
+        case .sendRequestEnd(let writePromise):
+            context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: writePromise)
 
             if let timeoutAction = self.idleReadTimeoutStateMachine?.requestEndSent() {
                 self.runTimeoutAction(timeoutAction, context: context)
@@ -260,15 +260,24 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             switch finalAction {
             case .close:
                 context.close(promise: nil)
-            case .sendRequestEnd:
-                context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
+                oldRequest.succeedRequest(buffer)
+            case .sendRequestEnd(let writePromise):
+                let writePromise = writePromise ?? context.eventLoop.makePromise(of: Void.self)
+                // We need to defer succeeding the old request to avoid ordering issues
+                writePromise.futureResult.whenComplete { result in
+                    switch result {
+                    case .success:
+                        oldRequest.succeedRequest(buffer)
+                    case .failure(let error):
+                        oldRequest.fail(error)
+                    }
+                }
+
+                context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: writePromise)
             case .informConnectionIsIdle:
                 self.connection.taskCompleted()
-            case .none:
-                break
+                oldRequest.succeedRequest(buffer)
             }
-
-            oldRequest.succeedRequest(buffer)
 
         case .failRequest(let error, let finalAction):
             // see comment in the `succeedRequest` case.
@@ -277,17 +286,26 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             self.runTimeoutAction(.clearIdleReadTimeoutTimer, context: context)
 
             switch finalAction {
-            case .close:
+            case .close(let writePromise):
                 context.close(promise: nil)
-            case .sendRequestEnd:
-                context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
-            case .informConnectionIsIdle:
+                writePromise?.fail(error)
+                oldRequest.fail(error)
+
+            case .informConnectionIsIdle(let writePromise):
                 self.connection.taskCompleted()
+                writePromise?.fail(error)
+                oldRequest.fail(error)
+
+            case .failWritePromise(let writePromise):
+                writePromise?.fail(error)
+                oldRequest.fail(error)
+
             case .none:
-                break
+                oldRequest.fail(error)
             }
 
-            oldRequest.fail(error)
+        case .failSendBodyPart(let error, let writePromise), .failSendStreamFinished(let error, let writePromise):
+            writePromise?.fail(error)
         }
     }
 
@@ -355,27 +373,28 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
 
     // MARK: Private HTTPRequestExecutor
 
-    private func writeRequestBodyPart0(_ data: IOData, request: HTTPExecutableRequest) {
+    private func writeRequestBodyPart0(_ data: IOData, request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
         guard self.request === request, let context = self.channelContext else {
             // Because the HTTPExecutableRequest may run in a different thread to our eventLoop,
             // calls from the HTTPExecutableRequest to our ChannelHandler may arrive here after
             // the request has been popped by the state machine or the ChannelHandler has been
             // removed from the Channel pipeline. This is a normal threading issue, noone has
             // screwed up.
+            promise?.fail(HTTPClientError.requestStreamCancelled)
             return
         }
 
-        let action = self.state.requestStreamPartReceived(data)
+        let action = self.state.requestStreamPartReceived(data, promise: promise)
         self.run(action, context: context)
     }
 
-    private func finishRequestBodyStream0(_ request: HTTPExecutableRequest) {
+    private func finishRequestBodyStream0(_ request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
         guard self.request === request, let context = self.channelContext else {
             // See code comment in `writeRequestBodyPart0`
             return
         }
 
-        let action = self.state.requestStreamFinished()
+        let action = self.state.requestStreamFinished(promise: promise)
         self.run(action, context: context)
     }
 
@@ -405,22 +424,22 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
 }
 
 extension HTTP1ClientChannelHandler: HTTPRequestExecutor {
-    func writeRequestBodyPart(_ data: IOData, request: HTTPExecutableRequest) {
+    func writeRequestBodyPart(_ data: IOData, request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
         if self.eventLoop.inEventLoop {
-            self.writeRequestBodyPart0(data, request: request)
+            self.writeRequestBodyPart0(data, request: request, promise: promise)
         } else {
             self.eventLoop.execute {
-                self.writeRequestBodyPart0(data, request: request)
+                self.writeRequestBodyPart0(data, request: request, promise: promise)
             }
         }
     }
 
-    func finishRequestBodyStream(_ request: HTTPExecutableRequest) {
+    func finishRequestBodyStream(_ request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
         if self.eventLoop.inEventLoop {
-            self.finishRequestBodyStream0(request)
+            self.finishRequestBodyStream0(request, promise: promise)
         } else {
             self.eventLoop.execute {
-                self.finishRequestBodyStream0(request)
+                self.finishRequestBodyStream0(request, promise: promise)
             }
         }
     }
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
@@ -291,9 +291,8 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
                 writePromise?.fail(error)
                 oldRequest.fail(error)
 
-            case .informConnectionIsIdle(let writePromise):
+            case .informConnectionIsIdle:
                 self.connection.taskCompleted()
-                writePromise?.fail(error)
                 oldRequest.fail(error)
 
             case .failWritePromise(let writePromise):
@@ -391,6 +390,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
     private func finishRequestBodyStream0(_ request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
         guard self.request === request, let context = self.channelContext else {
             // See code comment in `writeRequestBodyPart0`
+            promise?.fail(HTTPClientError.requestStreamCancelled)
             return
         }
 
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ConnectionStateMachine.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ConnectionStateMachine.swift
@@ -47,9 +47,7 @@ struct HTTP1ConnectionStateMachine {
             /// The promise is an optional write promise.
             case close(EventLoopPromise<Void>?)
             /// Inform an observer that the connection has become idle
-            ///
-            /// The promise is an optional write promise.
-            case informConnectionIsIdle(EventLoopPromise<Void>?)
+            case informConnectionIsIdle
             /// Fail the write promise
             case failWritePromise(EventLoopPromise<Void>?)
             /// Do nothing.
@@ -191,7 +189,7 @@ struct HTTP1ConnectionStateMachine {
             // as closed.
             //
             // TODO: AHC should support a fast rescheduling mechanism here.
-            return .failRequest(HTTPClientError.remoteConnectionClosed, .failWritePromise(nil))
+            return .failRequest(HTTPClientError.remoteConnectionClosed, .none)
 
         case .idle:
             var requestStateMachine = HTTPRequestStateMachine(isChannelWritable: self.isChannelWritable)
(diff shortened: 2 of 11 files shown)


dnadoba:
@swift-nio-bot add to allowlist

Lukasa (changes requested):
There needs to be much clearer auditing about who is responsible for succeeding this promise. A bunch of these calls return without either storing the promise or passing it on, and in those cases we risk leaking the promise without succeeding it. That all needs to be updated.

FranzBusch:
@Lukasa Thanks for your review. You absolutely write I just dropped the promise in a couple of places. I audited the code paths again and made sure to fail the promise in the other branches and added some tests.

The only thing that I wasn't really sure about is what errors to return in some places. Looking forward to suggestions!

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -281,17 +281,18 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
...
 
-    private func writeRequestBodyPart0(_ data: IOData, request: HTTPExecutableRequest) {
+    private func writeRequestBodyPart0(_ data: IOData, request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
         guard self.request === request, let context = self.channelContext else {
             // Because the HTTPExecutableRequest may run in a different thread to our eventLoop,
             // calls from the HTTPExecutableRequest to our ChannelHandler may arrive here after
             // the request has been popped by the state machine or the ChannelHandler has been
             // removed from the Channel pipeline. This is a normal threading issue, noone has
             // screwed up.
+            promise?.fail(ChannelError.eof)
```
Lukasa, line 291: Where you've used `ChannelError.eof`, I don't think that's right: I think we want `ioOnClosedChannel`.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -965,6 +966,8 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
     public static let readTimeout = HTTPClientError(code: .readTimeout)
     /// Remote connection was closed unexpectedly.
     public static let remoteConnectionClosed = HTTPClientError(code: .remoteConnectionClosed)
+    /// Request failed.
+    public static let failed = HTTPClientError(code: .failed)
```
Lukasa, line 970: Can we call this `requestFailed` instead of just `failed`?

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -154,8 +154,11 @@ final class RequestBag<Delegate: HTTPClientResponseDelegate> {
             return self.task.eventLoop.makeFailedFuture(error)
 
         case .write(let part, let writer, let future):
-            writer.writeRequestBodyPart(part, request: self)
-            self.delegate.didSendRequestPart(task: self.task, part)
+            let promise = self.task.eventLoop.makePromise(of: Void.self)
+            promise.futureResult.whenSuccess {
+                self.delegate.didSendRequestPart(task: self.task, part)
```
Lukasa, line 159: Should we report an error to the delegate in the failure case?

Sources/AsyncHTTPClient/HTTPClient.swift
FranzBusch: I was aligning it with the naming of `cancelled` which is documented as a cancelled request. What do you prefer?

Sources/AsyncHTTPClient/RequestBag.swift
FranzBusch: Was thinking about that as well. Open to input here. @weissi For your backpressure implementation would this be valuable?

fabianfett: We should make sure the error delegate is only called once.

Lukasa (approved):
LGTM but I'd like to see @fabianfett or @dnadoba review.

FranzBusch:
The API breakage is failing because of the two new protocol requirements. However, I added default implementations in an extension so this is not a breaking change.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -290,6 +291,7 @@ struct HTTPRequestStateMachine {
             if let expected = expectedBodyLength, sentBodyBytes + part.readableBytes > expected {
                 let error = HTTPClientError.bodyLengthMismatch
                 self.state = .failed(error)
+                promise?.fail(error)
```
fabianfett, line 294: See above. Fail promise outside of state machine lock.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -303,9 +305,10 @@ struct HTTPRequestStateMachine {
 
             self.state = .running(requestState, responseState)
 
-            return .sendBodyPart(part)
+            return .sendBodyPart(part, promise)
 
         case .failed:
+            promise?.fail(HTTPClientError.failed)
```
fabianfett, line 311: See above. Fail promise outside of state machine lock.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -336,35 +340,40 @@ struct HTTPRequestStateMachine {
             if let expected = expectedBodyLength, expected != sentBodyBytes {
                 let error = HTTPClientError.bodyLengthMismatch
                 self.state = .failed(error)
+                promise?.fail(error)
```
fabianfett, line 343: See above. Fail promise outside of state machine lock.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -336,35 +340,40 @@ struct HTTPRequestStateMachine {
...
-            return .sendRequestEnd
+            return .sendRequestEnd(promise)
 
         case .running(.streaming(let expectedBodyLength, let sentBodyBytes, _), .receivingBody(let head, let streamState)):
             assert(head.status.code < 300)
 
             if let expected = expectedBodyLength, expected != sentBodyBytes {
                 let error = HTTPClientError.bodyLengthMismatch
                 self.state = .failed(error)
+                promise?.fail(error)
```
fabianfett, line 356: See above. Fail promise outside of state machine lock.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -336,35 +340,40 @@ struct HTTPRequestStateMachine {
...
 
             self.state = .running(.endSent, .receivingBody(head, streamState))
-            return .sendRequestEnd
+            return .sendRequestEnd(promise)
 
         case .running(.streaming(let expectedBodyLength, let sentBodyBytes, _), .endReceived):
             if let expected = expectedBodyLength, expected != sentBodyBytes {
                 let error = HTTPClientError.bodyLengthMismatch
                 self.state = .failed(error)
+                promise?.fail(error)
```
fabianfett, line 367: See above. Fail promise outside of state machine lock.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -336,35 +340,40 @@ struct HTTPRequestStateMachine {
...
                 return .failRequest(error, .close)
             }
 
             self.state = .finished
+            promise?.succeed(())
```
fabianfett, line 372: See above. Succeed promise outside of state machine lock.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -336,35 +340,40 @@ struct HTTPRequestStateMachine {
...
             return .succeedRequest(.sendRequestEnd, .init())
 
         case .failed:
+            promise?.fail(HTTPClientError.failed)
```
fabianfett, line 376: See above. Fail promise outside of state machine lock.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -377,6 +386,7 @@ struct HTTPRequestStateMachine {
 
             // We may still receive something, here because of potential race conditions with the
             // producing thread.
+            promise?.fail(ChannelError.eof)
```
fabianfett, line 389: See above. Fail promise outside of state machine lock.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -431,12 +431,27 @@ public protocol HTTPClientResponseDelegate: AnyObject {
...
     func didSendRequestPart(task: HTTPClient.Task<Response>, _ part: IOData)
 
+    /// Called when a part of the request body failed to sent. Could be called zero or more times.
+    ///
+    /// - parameters:
+    ///     - task: Current request context.
+    ///     - part: Request body `Part`.
+    ///     - error: Error that occured during sending the request part.
+    func didFailSendingRequestPart(task: HTTPClient.Task<Response>, _ part: IOData, error: Error)
+
```
fabianfett, line 441: Why do we need to extend the public API here?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -431,12 +431,27 @@ public protocol HTTPClientResponseDelegate: AnyObject {
...
     ///     - task: Current request context.
     func didSendRequest(task: HTTPClient.Task<Response>)
 
+    /// Called when the request failed to sent. Will be called once.
+    ///
+    /// - parameters:
+    ///     - task: Current request context.
+    ///     - error: Error that occured during sending the request.
+    func didFailSendingRequest(task: HTTPClient.Task<Response>, error: Error)
+
```
fabianfett, line 454: Why do we need to extend the public API here?

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -154,8 +154,16 @@ final class RequestBag<Delegate: HTTPClientResponseDelegate> {
...
         case .write(let part, let writer, let future):
-            writer.writeRequestBodyPart(part, request: self)
-            self.delegate.didSendRequestPart(task: self.task, part)
+            let promise = self.task.eventLoop.makePromise(of: Void.self)
+            promise.futureResult.whenComplete { result in
+                switch result {
+                case .success:
+                    self.delegate.didSendRequestPart(task: self.task, part)
+                case .failure(let error):
+                    self.delegate.didFailSendingRequestPart(task: self.task, part, error: error)
```
fabianfett, line 163: I think, we should not change the old API at the moment and should just ignore this case. The error will be reported to the user anyway.

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -168,11 +176,18 @@ final class RequestBag<Delegate: HTTPClientResponseDelegate> {
...
-
-            self.delegate.didSendRequest(task: self.task)
+        case .forwardStreamFinished(let writer, let writerPromise):
+            let promise = self.task.eventLoop.makePromise(of: Void.self)
+            promise.futureResult.whenComplete { result in
+                switch result {
+                case .success:
+                    self.delegate.didSendRequest(task: self.task)
+                case .failure(let error):
+                    self.delegate.didFailSendingRequest(task: self.task, error: error)
```
fabianfett, line 186: Same as above. Don't extend public API here.

fabianfett (commented):
Thanks so much for driving this. Left a bunch of comments.

Sources/AsyncHTTPClient/ConnectionPool/HTTPRequestStateMachine.swift
```diff
@@ -274,6 +274,7 @@ struct HTTPRequestStateMachine {
             // won't be interested. We expect that the producer has been informed to pause
             // producing.
             assert(producerState == .paused)
+            promise?.fail(HTTPClientError.requestStreamCancelled)
```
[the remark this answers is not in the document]
weissi: +1, yeah we can't do that. That'll "invite" arbitrary code right into the state machine's lock and makes the state machine have implicit side effects.

fabianfett (commented):
Looks much better now. However we should change one more critical piece.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -937,6 +937,8 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         case getConnectionFromPoolTimeout
         case deadlineExceeded
         case httpEndReceivedAfterHeadWith1xx
+        case requestBodyStreamCancelled
```
fabianfett, line 940: There already is `requestStreamCancelled`, maybe we can reuse that?

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -269,8 +269,9 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             }
 
             oldRequest.succeedRequest(buffer)
+            promise?.succeed(())
```
fabianfett, line 272: this promise should be passed to:

```swift
            case .sendRequestEnd:
                context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
```

this will be more obvious once we attach the promise to `sendRequestEnd` and not to `succeedRequest`. In this case we should also defer the sending of `oldRequest.succeedRequest(buffer)` until the promise has fired, since we may run into ordering issues otherwise.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -269,8 +269,9 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
...
 
-        case .failRequest(let error, let finalAction):
+        case .failRequest(let error, let finalAction, let promise):
```
fabianfett, line 274:
```suggestion
        case .failRequest(let error, let finalAction, let writePromise):
```

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -288,6 +289,10 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             }
 
             oldRequest.fail(error)
+            promise?.fail(error)
```
fabianfett, line 292:
```suggestion
            writePromise?.fail(error)
```

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -355,27 +360,28 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
...
 
-    private func writeRequestBodyPart0(_ data: IOData, request: HTTPExecutableRequest) {
+    private func writeRequestBodyPart0(_ data: IOData, request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
         guard self.request === request, let context = self.channelContext else {
             // Because the HTTPExecutableRequest may run in a different thread to our eventLoop,
             // calls from the HTTPExecutableRequest to our ChannelHandler may arrive here after
             // the request has been popped by the state machine or the ChannelHandler has been
             // removed from the Channel pipeline. This is a normal threading issue, noone has
             // screwed up.
+            promise?.fail(ChannelError.eof)
```
fabianfett, line 370:
```suggestion
            promise?.fail(HTTPClientError.requestStreamCancelled)
```

Sources/AsyncHTTPClient/RequestBag.swift
```diff
@@ -168,11 +171,13 @@ final class RequestBag<Delegate: HTTPClientResponseDelegate> {
...
-            promise?.succeed(())
-
-            self.delegate.didSendRequest(task: self.task)
+        case .forwardStreamFinished(let writer, let writerPromise):
+            let promise = self.task.eventLoop.makePromise(of: Void.self)
+            promise.futureResult.whenSuccess {
+                self.delegate.didSendRequest(task: self.task)
+            }
+            writer.finishRequestBodyStream(self, promise: promise)
+            writerPromise?.succeed(())
```
fabianfett, line 180: Why do we create a new promise? Why is the writePromise just succeeded? This looks odd. This needs at least a comment.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -258,36 +258,52 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
...
-                context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
-            case .informConnectionIsIdle:
+            case .close(let writePromise):
+                let writePromise = writePromise ?? context.eventLoop.makePromise(of: Void.self)
+                // We need to defer succeeding the old request to avoid ordering issues
+                writePromise.futureResult.whenComplete { _ in
+                    oldRequest.succeedRequest(buffer)
+                }
+
+                context.close(promise: writePromise)
```
FranzBusch, line 268: @fabianfett Is it okay to pass the `writePromise` to `close` here?

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -258,36 +258,52 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
...
             self.runTimeoutAction(.clearIdleReadTimeoutTimer, context: context)
 
             switch finalAction {
-            case .close:
-                context.close(promise: nil)
-            case .sendRequestEnd:
-                context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
-            case .informConnectionIsIdle:
+            case .close(let writePromise):
+                context.close(promise: writePromise)
```
FranzBusch, line 299: Same question here @fabianfett ?

Sources/AsyncHTTPClient/RequestBag.swift
FranzBusch: I thought that we need a new promise here but I just tested it again and with using the same promise and it worked.

fabianfett (commented):
And another round!

fabianfett (commented):
Nearly done!

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ConnectionStateMachine.swift
```diff
@@ -390,41 +408,57 @@ extension HTTP1ConnectionStateMachine.State {
...
+                return .failRequest(error, .close(nil))
+
+            case (.inRequest(_, _), .close(let writePromise)):
+                self = .closing
+                return .failRequest(error, .close(writePromise))
+
+            // otherwise we fail with .informConnectionIsIdle
+            case (.inRequest(_, _), .none):
+                self = .idle
+                return .failRequest(error, .informConnectionIsIdle(nil))
```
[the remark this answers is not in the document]
FranzBusch: Changed it back

fabianfett (approved):
Thank you! This has been a ton of work!

PR #580, opened 2022-04-11
[Redirect] Allow redirect response to have body

### LINKED ISSUE #574: GET to a URL with a redirection never completes nor times out
Hi there, 
I realized some requests in our backend are getting stuck, they don't complete or finish unless I set a deadline.
Upon further investigation, the requests that are getting stuck are ones with a redirect. Doing a `HEAD` instead of a `GET` does result in a response (with a 301 status when disabling redirects in the client).

I'm trying to use the `FileDownloadDelegate` and I also noticed that nothing is getting called (no calls to reportHead or reportProgress). But as I cannot issue a simple `GET` I guess that the issue is not in the `FileDownloadDelegate`.

This is some sample code that reproduces the issue:
```swift
import AsyncHTTPClient

let url = "https://ipfs.infura.io/ipfs/QmVES23Brg2HP2Cn5AEzkiToYpAPxvYfXrHbrg91ScLUES/image.jpeg"

let response = try httpClient.execute(request: .init(url: url, method: .GET)).wait()

print(response)
```
When running that on my Mac (Xcode 13.3 Apple Swift version 5.6 (swiftlang-5.6.0.323.62 clang-1316.0.20.8)), the request never completes.

Changing the `.GET` to a `.HEAD` does make the request complete successfully.

---

To workaround this issue, I'm creating an HTTPClient that does not follow redirects, to be able to inspect the `location` header and then use that URL on the client that downloads files.
I also noticed that if you do a `HEAD` request to a URL that will perform a redirect, there seems to be no way of obtaining the redirected URL (hence having to disable redirects on the workaround client). That would also be a nice to have.

Might be related to #538

fabianfett:
### Motivation

Currently, we don’t consume the response body of redirect responses. Because of this requests, that receive a redirect response with response body, may hang indefinitely.

### Changes

- Consume redirect response body if less than 3kb
- Cancel redirect response if larger than 3kb

### Result

Redirect responses are consumed. Fixes #574

### DIFF
--- a/Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
+++ b/Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
@@ -16,6 +16,16 @@ import struct Foundation.URL
 import NIOCore
 import NIOHTTP1
 
+extension HTTPClient {
+    /// The maximum body size allowed, before a redirect response is cancelled. 3KB.
+    ///
+    /// Why 3KB? We feel like this is a good compromise between potentially reusing the
+    /// connection in HTTP/1.1 mode (if we load all data from the redirect response we can
+    /// reuse the connection) and not being to wasteful in the amount of data that is thrown
+    /// away being transferred.
+    fileprivate static let maxBodySizeRedirectResponse = 1024 * 3
+}
+
 extension RequestBag {
     struct StateMachine {
         fileprivate enum State {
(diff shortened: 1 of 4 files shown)


fabianfett:
@swift-server-bot test this please!

Lukasa (approved):

PR #582, opened 2022-04-21
Fix flaky `TransactionTests.testResponseStreamFails()` test

dnadoba:
# Motivation 
`EmbeddedEventLoop` event loop is not thread-safe but we create a `EventLoopFuture` that is required to be thread-safe from an `EmbeddedEventLoop`. This leads to undefined behaviour and therefore a flaky test.

# Modification
This PR replaces the `EventLoopFuture` with a simple `Promise` implementation that uses Swift Concurrency to protect against concurrent access.

# Result
Before this patch, I could reliably fail the test after ~5K repetitions. 
With this patch, I have run the test 100K times without a single failure.

Note: this only effects code in the test target


Lukasa (approved):
Nice, this LGTM.

dnadoba:
5.7 & nightly fail because of sendable related errors.

PR #583, opened 2022-04-21
Add HTTPClientError shortDescription property

cpriebe:
This adds a shortDescription property to HTTPClientError which provides a short
description of the error without associated values.

This is useful for e.g. metric labels which should not contain dynamic, potentially unbounded, associated values.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -949,6 +949,76 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         return "HTTPClientError.\(String(describing: self.code))"
     }
 
+    /// Short description of the error that can be used in case a bounded set of error descriptions is expected, e.g. to
+    /// include in metric labels. For this reason the description must not contain associated values.
+    public var shortDescription: String {
+        // When adding new cases here, do *not* include dynamic (associated) values in the description.
+        switch self.code {
+        case .invalidURL:
+            return "Invalid URL"
+        case .emptyHost:
+            return "Empty host"
+        case .missingSocketPath:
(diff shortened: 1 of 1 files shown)


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -949,6 +949,73 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         return "HTTPClientError.\(String(describing: self.code))"
     }
 
+    public var shortDescription: String {
```
fabianfett, line 952: Would you mind adding a comment that describes, when users should use `shortDescription` and what they need to take into consideration when adding another case?

cpriebe: Added now. Happy to adjust the comments of course!

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -949,6 +949,76 @@ public struct HTTPClientError: Error, Equatable, CustomStringConvertible {
         return "HTTPClientError.\(String(describing: self.code))"
     }
 
+    /// Short description of the error that can be used in case a bounded set of error descriptions is expected, e.g. to
+    /// include in metric labels. The description does not contain associated values.
```
fabianfett, line 953:
```suggestion
    /// include in metric labels. For this reason the description must not contain associated values.
```

fabianfett (commented):
One small change :) Since this is a public API change, let's get @Lukasa on here as well.

Lukasa:

fabianfett (approved):
LGTM. Thanks! Let's wait for @Lukasa thumb though.

Lukasa (approved):

PR #584, opened 2022-04-21
Fix flaky `HTTPClientTests.testResponseDelayGet()` test

dnadoba:
# Motivation 
We observed a test failure on our CI:
```
/async-http-client/Tests/AsyncHTTPClientTests/HTTPClientTests.swift: error: HTTPClientTests.testResponseDelayGet : XCTAssertGreaterThan failed: ("1.9851820468902588") is not greater than ("2.0") -
Test Case 'HTTPClientTests.testResponseDelayGet' failed (1.989 seconds)
```

# Modification
I could not reproduce the failure myself but giving this a bit more leeway does not hurt. To measure the time difference, we now use `NIODeadline` which internally uses nanoseconds since boot to measure the time instead of `Date`. `Date` uses the system time which may change during execution due to time sync with a time server and is therefore not best suited to measure elapsed time.

# Result
Immune to system time sync and slightly higher tolerance agains inaccurate timer implementations.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -19,5 +19,6 @@
 --disable redundantReturn
 --disable preferKeyPath
 --disable sortedSwitchCases
+--disable numberFormatting
 
 # rules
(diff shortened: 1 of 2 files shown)


fabianfett:
It seems that we need a code format run

Lukasa (approved):

fabianfett (approved):
LGTM. Thanks.

dnadoba:
5.7 & nightly fail because of sendable related errors.

PR #588, opened 2022-05-31
Improve errors and testing using NIOTS

Lukasa:
Motivation

Currently error reporting with NIO Transport Services is often sub-par.
This occurs because the Network.framework connections may enter the
waiting state until the network connectivity state changes. We were not
watching for the user event that contains the error in that state, so if
we timed out in that state we'd just give a generic timeout error,
instead of telling the user anything more detailed.

Additionally, several of our tests assume that failure will be fast, but
in NIO Transport Services we will enter that .waiting state. This is
reasonable, as changed network connections may make a connection that
was not succeeding suddenly viable. However, it's inconvenient for
testing, where we're mostly interested in confirming that the error path
works as expected.

Modifications

- Add an observer of the WaitingForConnectivity event that records it
  into our state machine for later reporting.
- Add support for disabling waiting for connectivity for testing
  purposes.
- Add annotations to several tests to stop them waiting for
  connectivity.

Results

Faster tests, better coverage, better errors for our users.

### DIFF
--- a/Sources/AsyncHTTPClient/NIOTransportServices/NWWaitingHandler.swift
+++ b/Sources/AsyncHTTPClient/NIOTransportServices/NWWaitingHandler.swift
@@ -14,11 +14,11 @@
 
 #if canImport(Network)
 import Network
-#endif
 import NIOCore
 import NIOHTTP1
 import NIOTransportServices
 
+@available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
 final class NWWaitingHandler<Requester: HTTPConnectionRequester>: ChannelInboundHandler {
     typealias InboundIn = Any
     typealias InboundOut = Any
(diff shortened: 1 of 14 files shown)


dnadoba (approved):
Great patch, only two minor suggestions

Sources/AsyncHTTPClient/NIOTransportServices/NWWaitingHandler.swift
```diff
@@ -0,0 +1,49 @@
...
+    private let connectionID: HTTPConnectionPool.Connection.ID
+
+    init(requester: Requester, connectionID: HTTPConnectionPool.Connection.ID) {
+        self.requester = requester
+        self.connectionID = connectionID
+    }
+
+    func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        #if canImport(Network)
+        if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *) {
```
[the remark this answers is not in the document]
Lukasa: Sure, why not.

Sources/AsyncHTTPClient/NIOTransportServices/NWWaitingHandler.swift
```diff
@@ -0,0 +1,46 @@
...
+    func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        if let waitingEvent = event as? NIOTSNetworkEvents.WaitingForConnectivity, let requester = self.requester {
+            requester.waitingForConnectivity(self.connectionID, error: HTTPClient.NWErrorHandler.translateError(waitingEvent.transientError))
+            self.requester = nil
+        }
+        context.fireUserInboundEventTriggered(event)
+    }
+
+    func handlerRemoved(context: ChannelHandlerContext) {
+        self.requester = nil
```
fabianfett, line 43: just out of interest, why do we release the requester here explicitly? Could there be a race that we get an inbound event after we have been removed from the pipeline?

fabianfett (approved):
This looks great. And I'm sure that the new code path is hit plenty in more complex scenarios. However I would love to see two more test cases:

1. Unit style: the new `NWWaitingHandler` in an `EmbeddedChannel`. Is the callout triggered.
2. An explicit test: NWError is reported if `networkFrameworkWaitForConnectivity` is turned on. Currently this is implicit in `testAvoidLeakingTLSHandshakeCompletionPromise`.

Doing those in a follow up is totally fine with me.

Sources/AsyncHTTPClient/NIOTransportServices/NWWaitingHandler.swift
Lukasa: No, this was initially defensive to prevent requesters living an excessively long time. However, now that I think about it I suspect this is unnecessarily defensive, as nothing else is holding onto this handler.

Lukasa:
I've addressed (2), I'm going to tackle (1) in the future because it requires me to make an enhancement to NIOTS.

PR #592, opened 2022-06-09
Use a local TCP server that doesn’t accept connections on macOS for `testConnectTimeout()`

dnadoba:
We have observed some test failures on our CI where `198.51.100.254` accepts connection but it shouldn't.
Luckily, this happens only on macOS and we can workaround this by creating a local TCP server with a backlog of `1`.
This server will only send a `SYN/ACK` back to the first connection as it doesn't actually accept the connection. Further connection attempts will not see a `SYN/ACK`. We can use this behaviour to test a connection timeout on macOS.

The backlog behaviour on Linux can not be used to simulate a non-reachable server. Linux sends a `SYN/ACK` back even if the `backlog` queue is full as it has two queues. The second queue is not limit by `ChannelOptions.backlog` but by `/proc/sys/net/ipv4/tcp_max_syn_backlog`. We will therefore fallback to `198.51.100.254` on Linux.


dnadoba:
Swift 5.7 & nightly fails because of `Sendable` warnings.

Lukasa (approved):

PR #594, opened 2022-06-15
Add `testSelfSignedCertificateIsRejectedWithCorrectError`

dnadoba:
We had bug report that an error during TLS handshake does not get correctly reported and instead a deadlineExceeded error is thrown. We have already fixed the issue in #588 and this PR just adds a test to verify the behavior.

### Modifications:
- add test `testSelfSignedCertificateIsRejectedWithCorrectError` which uses a self signed certificate to fail the TLS handshake
- improve error description of `NWTLSError`. Previously a TLS error had a `description` that looked like this:
```
The operation couldn’t be completed. (Network.NWError error 2.)
```
which contains not enough for further investigation. It now contains the OSStatus code too which can be used to find out what actually went wrong:
```
-9808: Optional(bad certificate format)
```
- make internal class `NWErrorHandler` final
- fix error message of `XCTAssertEqualTypeAndValue`. It had the left hand side type and right hand side swapped.

### Result
Improved test coverage and better error messages


dnadoba:
5.7 & nightly fail because of `Sendable` warnings

PR #599, opened 2022-06-17
Correctly close the connection if sendEnd fails

Lukasa:
Motivation

If we receive an early HTTP response, the last action on a HTTP/1.1
connection is to send the .end message. While we had an error handling
path in the code, it wasn't tested, and when executed it would end up
leaking the connection by failing to close it _or_ return it to the
pool.

This patch fixes the issue by appropriately terminating the connection
and adding a test.

Modifications

Add a test
Terminate the connection if sendEnd fails

Result

Fewer connection leaks


fabianfett (approved):
Great fix. Thanks @Lukasa 🙏

PR #601, opened 2022-06-24
Report last connection error if request deadline is exceeded

dnadoba:
If the request deadline is exceeded, we should provide as much information as possible to let the user know the underlying reason why the deadline was exceeded. We should therefore report the last connection error instead of `HTTPClientError.deadlineExceeded` if the request was never executed because of a connection error.

### Modification
Do not fail the request in the `RequestBag` if  canceled. Instead, let the connection pool state machine fail the request as it has access to the last connection error.

### Result
Instead of `HTTPClientError.deadlineExceeded` we now get a connection error if the request couldn't be scheduled because of a connection error.

If a user manually cancels a request, we currently still just report `HTTPClientError.cancelled`. However, we could now easily change the behavior and also report the last connection error if we decided to do so.

async/await does not use `RequestBag` but `Transaction` and still has the old behaviour. I will apply similar changes to `Transaction` in a follow up PR.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
@@ -323,9 +323,10 @@ extension HTTPConnectionPool {
 
         mutating func cancelRequest(_ requestID: Request.ID) -> Action {
             // 1. check requests in queue
-            if self.requests.remove(requestID) != nil {
+            if let request = self.requests.remove(requestID) {
+                let error = lastConnectFailure ?? HTTPClientError.cancelled
                 return .init(
-                    request: .cancelRequestTimeout(requestID),
+                    request: .failRequest(request, error, cancelTimeout: true),
                     connection: .none
                 )
             }
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
@@ -324,6 +324,7 @@ extension HTTPConnectionPool {
         mutating func cancelRequest(_ requestID: Request.ID) -> Action {
             // 1. check requests in queue
             if let request = self.requests.remove(requestID) {
+                // Use the last connection error to let the user know why the request was never scheduled
                 let error = self.lastConnectFailure ?? HTTPClientError.cancelled
                 return .init(
                     request: .failRequest(request, error, cancelTimeout: true),
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
@@ -324,7 +324,7 @@ extension HTTPConnectionPool {
         mutating func cancelRequest(_ requestID: Request.ID) -> Action {
             // 1. check requests in queue
             if let request = self.requests.remove(requestID) {
-                let error = lastConnectFailure ?? HTTPClientError.cancelled
+                let error = self.lastConnectFailure ?? HTTPClientError.cancelled
                 return .init(
                     request: .failRequest(request, error, cancelTimeout: true),
                     connection: .none
--- a/Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
+++ b/Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
@@ -31,7 +31,9 @@ extension RequestBag {
         fileprivate enum State {
             case initialized
             case queued(HTTPRequestScheduler)
-            /// if the deadline was exceeded while in the `.queued(_:)` state, we wait until the request pool fails the request with a potential more descriptive error message if a connection failure has occured while the request was queued.
+            /// if the deadline was exceeded while in the `.queued(_:)` state,
+            /// we wait until the request pool fails the request with a potential more descriptive error message,
+            /// if a connection failure has occured while the request was queued.
             case deadlineExceededWhileQueued
             case executing(HTTPRequestExecutor, RequestStreamState, ResponseStreamState)
             case finished(error: Error?)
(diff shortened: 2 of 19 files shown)


Lukasa:
Your PR comment implies that if a user manually cancels we still report `.cancelled`, but the code seems to have functionality that diverges from that. Can you clarify what's happening there?

dnadoba:
@Lukasa I have removed `CancelationReason` and instead added a new method/action `deadlineExceeded()` to make it more explicit that we only get the new behaviour for deadlineExceeded and not manual user cancelation.

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -532,8 +534,23 @@ extension RequestBag.StateMachine {
...
+        case fail(FailAction)
+    }
+
+    mutating func deadlineExceeded() -> DeadlineExceededAction {
+        switch self.state {
+        case .queued(let queuer):
+            self.state = .deadlineExceededWhileQueued
+            return .cancelScheduler(queuer)
+        default:
+            return .fail(self.fail(HTTPClientError.deadlineExceeded))
```
Lukasa, line 548: Is it correct that we don't change state here?

Lukasa: It doesn't look right: historically we transitioned into a terminal state in `fail`, now we don't.

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -542,31 +559,39 @@ extension RequestBag.StateMachine {
...
         case .finished(.none):
             // An error occurred after the request has finished. Ignore...
             return .none
+        case .deadlineExceededWhileQueued:
+            // if we just get a `HTTPClientError.cancelled` we can use the orignal cancelation reason
+            // to give a more descriptive error to the user.
+            if (error as? HTTPClientError) == .cancelled {
+                return .failTask(HTTPClientError.deadlineExceeded, nil, nil)
+            }
+            // otherwise we already had an intermidate connection error which we should present to the user instead
```
Lukasa, line 593:
```suggestion
            // otherwise we already had an intermediate connection error which we should present to the user instead
```

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -542,31 +559,39 @@ extension RequestBag.StateMachine {
...
+            return .failTask(error, nil, executor)
         case .redirected:
             self.state = .finished(error: error)
-            return .failTask(nil, nil)
+            return .failTask(error, nil, nil)
         case .finished(.none):
             // An error occurred after the request has finished. Ignore...
             return .none
+        case .deadlineExceededWhileQueued:
+            // if we just get a `HTTPClientError.cancelled` we can use the orignal cancelation reason
```
Lukasa, line 588:
```suggestion
            // if we just get a `HTTPClientError.cancelled` we can use the original cancellation reason
```

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
dnadoba: We call down to `self.fail(_:)` which does the state transition for us.
 https://github.com/swift-server/async-http-client/blob/5e575855576050753c764145ed6e0068222fea59/Sources/AsyncHTTPClient/RequestBag%2BStateMachine.swift#L558-L601

Lukasa: Ah, ok, can we add a comment to that effect? It was not entirely obvious to me.

dnadoba: Done in https://github.com/swift-server/async-http-client/pull/601/commits/c7640da4227db0da213c7d88709b13b4e390e715

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -95,7 +97,7 @@ extension RequestBag.StateMachine {
         case .initialized, .queued:
             self.state = .executing(executor, .initialized, .initialized)
             return true
-        case .finished(error: .some):
+        case .finished(error: .some), .deadlineExceededWhileQueued:
             return false
```
fabianfett, line 101: We'll need to fail the request here explicitly...

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -532,8 +534,25 @@ extension RequestBag.StateMachine {
...
+        case cancelScheduler(HTTPRequestScheduler?)
+        case fail(FailAction)
+    }
+
+    mutating func deadlineExceeded() -> DeadlineExceededAction {
+        switch self.state {
+        case .queued(let queuer):
+            self.state = .deadlineExceededWhileQueued
+            return .cancelScheduler(queuer)
+        default:
```
fabianfett, line 547: Can we make all the other cases explicit here?

Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
```diff
@@ -532,8 +534,25 @@ extension RequestBag.StateMachine {
...
 
+    enum DeadlineExceededAction {
+        case cancelScheduler(HTTPRequestScheduler?)
+        case fail(FailAction)
+    }
+
+    mutating func deadlineExceeded() -> DeadlineExceededAction {
+        switch self.state {
+        case .queued(let queuer):
+            self.state = .deadlineExceededWhileQueued
```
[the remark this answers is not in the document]
fabianfett: For this we def. need a test case as well.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -323,9 +323,10 @@ extension HTTPConnectionPool {
 
         mutating func cancelRequest(_ requestID: Request.ID) -> Action {
             // 1. check requests in queue
-            if self.requests.remove(requestID) != nil {
+            if let request = self.requests.remove(requestID) {
+                let error = self.lastConnectFailure ?? HTTPClientError.cancelled
```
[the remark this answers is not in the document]
fabianfett: In theory we could check here if we are at max connections... In those cases we could issue a more explicit: `deadlineExceededBecauseToQueueDepthToLong` error.

fabianfett (commented):
Overall looks great. We should remove the `.cancelRequestTimeout` Action and make sure we handle the race condition correctly.

fabianfett (changes requested):
Hupps. Wrong state selected before.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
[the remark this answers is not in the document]
dnadoba: Not sure how useful this kind of error message would be because `deadlineExceededBecauseToQueueDepthToLong` may only be a symptom as other requests in front could not be scheduled because there was a connection failure.

fabianfett (approved):
Looks awesome! Thanks!

PR #607, opened 2022-08-03
Remove the last remaining NIOAtomic

### LINKED ISSUE #606: HTTPClient is still using NIOAtomic
Given the other NIOAtomics were replaced with ManagedAtomic I assume the HTTPClient one will want to be replaced as well

https://github.com/swift-server/async-http-client/blob/2adca4b0038a7ef81ea0838c5feeda9439c52b64/Sources/AsyncHTTPClient/HTTPClient.swift#L39

Lukasa:
Motivation

Warnings aren't great, and NIOAtomic is deprecated.

Modifications

Replace the last use of NIOAtomic with ManagedAtomic.

Result

Fewer warnings
Fixes #606

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -12,8 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-import Foundation
 import Atomics
+import Foundation
 import Logging
 import NIOConcurrencyHelpers
 import NIOCore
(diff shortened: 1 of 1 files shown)


dnadoba (approved):

adam-fowler:
One minor thing, maybe the version constraints in Package.swift should require versions that don't use NIOAtomics

Lukasa:
Do you mean versions of swift-nio? I don't think that's necessary: there are no relevant correctness issues with old copies of NIO, and we don't require the new ones, so there's no particular issue with letting users resolve older versions if that's necessary to get a buildable tree.

Issue #537, opened 2021-12-29
HTTP/2 decompression support missing


ve6yeq:
AsyncHTTPClient commit hash: 19e83a35df172e9cc02a1caa7bb615aeab11e587

Context:
It is possible to enable transparent decompression of compressed responses using the `Configuration.decompression` property. This works fine for HTTP/1.1 connections with the response body being "transparently" decompressed but does not appear to be implemented for HTTP/2 connections where the response body is still compressed. 

Any idea on where this transparent decompression should be connected into with HTTP/2?

Lukasa:
I think this is for @fabianfett and @dnadoba: we may need to bring `NIOHTTPResponseDecompressor` into our HTTP/2 pipelines.

rnantes:
This caused me issues. I ended up having to use  `config.httpVersion = .http1Only` Or  `headers.add(name: "accept-Encoding", value: "identity")`

PR #610, opened 2022-08-05
Support transparent decompression with HTTP/2

dnadoba:
Fixes https://github.com/swift-server/async-http-client/issues/537


Lukasa (approved):
Nice one, good patch

PR #611, opened 2022-08-05
Limit max recursion depth delivering body parts

Lukasa:
Motivation

When receiving certain patterns of response body parts, we can end up
recursing almost indefinitely to deliver them to the application. This
can lead to crashes, so we might politely describe it as "sub-optimal".

Modifications

Keep track of our stack depth and avoid creating too many stack frames.
Added some unit tests.

Result

We no longer explode when handling bodies with lots of tiny parts.


dnadoba (approved):
Great patch, thanks!

Issue #364, opened 2021-05-11
fix FileDownloadDelegate


weissi:
`FileDownloadDelegate` calls `pool.start()` on a `NIOThreadPool` that is either provided by the user or by AHC itself through

```swift
public init(
        path: String,
        pool: NIOThreadPool = NIOThreadPool(numberOfThreads: 1),
        reportHead: ((HTTPResponseHead) -> Void)? = nil,
        reportProgress: ((Progress) -> Void)? = nil
    )
```

AHC however never calls `.shutdownGracefully`/`.syncShutdownGracefully` on that pool. That means the user _has_ to provide a `NIOThreadPool`, or else we'll just leak it (because AHC never calls stop).

Ie. in the default use case we
- spawn a fresh thread _per download_
- and then leak the thread

😭

PR #614, opened 2022-08-12
Fix thread leak in `FileDownloadDelegate`

dnadoba:
Fixes https://github.com/swift-server/async-http-client/issues/364

### Modification
- remember if we own the thread pool or not.
- shutdown the thread pool if we own it after we are done with it

### Result
we no longer leak one thread per download

### DIFF
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -30,17 +30,9 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
     public typealias Response = Progress
 
     private let filePath: String
-    private let io: NonBlockingFileIO
+    private var io: NonBlockingFileIO?
     private let reportHead: ((HTTPResponseHead) -> Void)?
     private let reportProgress: ((Progress) -> Void)?
-
-    private enum ThreadPool {
-        case unowned
-        // if we own the thread pool we also need to shut it down
-        case owned(NIOThreadPool)
-    }
-
-    private let threadPool: ThreadPool
 
     private var fileHandleFuture: EventLoopFuture<NIOFileHandle>?
     private var writeFuture: EventLoopFuture<Void>?
@@ -61,7 +53,7 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
         reportHead: ((HTTPResponseHead) -> Void)? = nil,
         reportProgress: ((Progress) -> Void)? = nil
     ) throws {
-        try self.init(path: path, sharedThreadPool: pool, reportHead: reportHead, reportProgress: reportProgress)
+        try self.init(path: path, pool: .some(pool), reportHead: reportHead, reportProgress: reportProgress)
     }
 
     /// Initializes a new file download delegate and spawns a new thread for file I/O.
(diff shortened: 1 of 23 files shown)


dnadoba:
API breakage failure is a false positive
```
16:14:03 1 breaking change detected in AsyncHTTPClient:
16:14:03   💔 API breakage: constructor FileDownloadDelegate.init(path:pool:reportHead:reportProgress:) has removed default argument from parameter 1
```
The change in question can be simplified to:
```swift
struct Foo {
    // old init
    init(bar: Int = 0) {}
    // new inits
    init() {}
    init(bar: Int) {}
}
```
replacing a default argument with two inits, one with the argument and one without, is an ABI breaking change but not an API breaking change.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -128,4 +177,14 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
...
         return self.progress
     }
+
+    deinit {
+        switch threadPool {
+        case .unowned:
+            break
+        case .owned(let pool):
+            // if the delegate is unused we still need to shutdown the thread pool
+            try! pool.syncShutdownGracefully()
```
weissi, line 187: we shouldn't destroy this in `deinit`, you don't even have a real guarantee that `deinit` isn't running in a thread owned by the pool (which would make it deadlock).

I'd suggest to add a lifecycle hook (shutdown or so) to the delegate and have AHC invoke that. That new shutdown method can have an empty default implementation to not break the API.

Separately, I think AHC should _provide_ the thread pool or else we spin up/down a thread per download and that kinda makes it pointless to do asynchronous I/O if we need a new thread each time.

weissi: actually you could add two delegate methods (with default empty impls):

- `provideThreadPool(ioPool)` where AHC provides a running thread pool to the delegate
- `shutdown()` where AHC tells the delegate that it its job is done

After that, we can change the default here and have the delegate not even have the option to accept an io pool from the user.

dnadoba: If AHC provides the default shared thread pool we don't need the `shutdown()` lifecycle hook because the delegate doesn't need to shutdown the thread pool anymore but the AHC client.

weissi: that's true

dnadoba: I have added a shared file IO thread pool to the `HTTPClient` which is passed to the delegate through a call to `provideSharedThreadPool(fileIOPool:)`
https://github.com/swift-server/async-http-client/pull/614/commits/1afc151d644c489a233916d26b99b966d6154dee

weissi (approved):
awesome, thank you!

dnadoba:
I have changed the implementation because we would otherwise now only ever use one thread for all file IO. This is not good because file IO can block the thread and there is now easy way of making it non-blocking. We therefore initialize the thread pool with `ProcessInfo.processInfo.processorCount` threads. As this is potentially much more expensive operation then just creating one thread we also delay initialization until the first actual write.

Lukasa (changes requested):
Generally looks good, thanks, but we'll want some tests of the functionality I think.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -622,15 +623,26 @@ extension HTTPClient {
         private var _isCancelled: Bool = false
         private var _taskDelegate: HTTPClientTaskDelegate?
         private let lock = Lock()
+        private let makeOrGetFileIOThreadPool: () -> NIOThreadPool
 
-        init(eventLoop: EventLoop, logger: Logger) {
+        public var fileIOThreadPool: NIOThreadPool {
```
dnadoba, line 629: I just realized that we actually don't need to make this public. @Lukasa any objections to me making it internal?

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -47,20 +47,60 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
...
+            self.io = nil
+        }
+
         self.filePath = path
 
         self.reportHead = reportHead
         self.reportProgress = reportProgress
     }
 
+    public func provideSharedThreadPool(fileIOPool: NIOThreadPool) {
```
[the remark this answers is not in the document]
dnadoba: Not used any more. Removed.

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa: None at all.

dnadoba: Done in https://github.com/swift-server/async-http-client/pull/614/commits/c3c68b2a27cc55c7597ddd83025b71fec7404ffc

dnadoba:
@Lukasa test added in https://github.com/swift-server/async-http-client/pull/614/commits/9ee5c82cfbf732d7cf2aab533c1f35392b8d8680

Lukasa (approved):

PR #616, opened 2022-08-16
Accurately apply the connect timeout in async code

Lukasa:
Motivation

We should apply the connect timeout to the complete set of connection
attempts, rather than the request deadline. This allows users
fine-grained control over how long we attempt to connect for. This is
also the behaviour of our old-school interface.

Modifications

- Changed the connect deadline calculation for async/await to match that
  of the future-based code.
- Added a connect timeout test.

Result

Connect timeouts are properly handled


artemredkin (approved):

dnadoba (approved):
Great patch! Thanks for fixing this!

Lukasa:
Nightly fails due to missing Sendable checks.

PR #618, opened 2022-08-20
Use NIOCore.System.coreCount for the fileIO thread pool

karwa:
`ProcessInfo` is a Foundation API, so this broke the Foundationless build


Lukasa (approved):
Good catch, thanks!

PR #619, opened 2022-08-20
NIOFoundationCompat does not appear to be used

karwa:


Lukasa (approved):
Yeah, I think this is vestigial from a time when we used URL or Data.

PR #625, opened 2022-09-06
Allow immediate request failure on connection error

dnadoba:
### Motivation
In integration tests we often want to inspect the connection error thrown during connection establishment for a request. Currently the only way to get at the connection error is to wait until the request timeout is expired. This is because of the connection retry behaviour backed into AHC which cannot be turned off. AHC will try to create new connections on connection failure with an exponential backoff. It stores the last connection and will fail the request if the request times out. 

### Modification
Adds a new **internal** flag called `retryConnectionEstablishment` to `HTTPClient.Configuration.ConnectionPool`.
If false, requests will fail immediately after a connection could not be established. Otherwise the old behavior is used which is still the default.

Set `retryConnectionEstablishment` to false to speedup test from ~40s to ~1s in `HTTPClientNIOTSTests.swift`. More tests will follow.

### Result
Allows fast and reliable test execution.


Tests/AsyncHTTPClientTests/HTTPConnectionPool+HTTP1StateTests.swift
```diff
@@ -102,13 +103,86 @@ class HTTPConnectionPool_HTTP1StateMachineTests: XCTestCase {
...
+            guard case .none = action.connection else {
+                return XCTFail("Unexpected connection action")
+            }
+            XCTAssertEqual(.scheduleRequestTimeout(for: request, on: mockRequest.eventLoop), action.request)
+            XCTAssertNoThrow(try queuer.queue(mockRequest, id: request.id))
+        }
+
+        // timeout all queued requests except for two
+
+        // fail all connection attempts
```
fabianfett, line 151: I guess those can be removed.

dnadoba: Good catch! Removed in https://github.com/swift-server/async-http-client/pull/625/commits/2cff156b95db93d9bc9bedb1b0c7890a827ad138

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -29,16 +30,19 @@ extension HTTPConnectionPool {
 
         private(set) var requests: RequestQueue
         private(set) var lifecycleState: StateMachine.LifecycleState
+        private let retryConnectionEstablishment: Bool
```
fabianfett, line 33: Can we add a comment here, why we have this property. We should note wherever we can, that this is a dull setting. :P

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -219,6 +223,17 @@ extension HTTPConnectionPool {
 
             switch self.lifecycleState {
             case .running:
+                guard self.retryConnectionEstablishment else {
+                    guard let (index, _) = self.connections.failConnection(connectionID) else {
+                        preconditionFailure("Failed to create a connection that is unknown to us?")
```
fabianfett, line 228: I often write the precondition messages like this... But I actually consider this bad style. This might be better:
```suggestion
                        preconditionFailure("A connection attempt failed, that the state machine knows nothing about. Somewhere state was lost.")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -244,6 +259,13 @@ extension HTTPConnectionPool {
         mutating func waitingForConnectivity(_ error: Error, connectionID: Connection.ID) -> Action {
             self.lastConnectFailure = error
 
+            guard self.retryConnectionEstablishment else {
+                return .init(
+                    request: self.failAllRequests(reason: error),
+                    connection: .none
+                )
+            }
+
```
fabianfett, line 268: This is weird! If we always fail all requests, when we a connection attempt failed, we should never reschedule a connection attempt. And this means we should never reach this point.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -33,16 +34,19 @@ extension HTTPConnectionPool {
         private let idGenerator: Connection.ID.Generator
 
         private(set) var lifecycleState: StateMachine.LifecycleState
+        private let retryConnectionEstablishment: Bool
```
fabianfett, line 37: Can we add a comment here, why we have this property. We should note wherever we can, that this is a dull setting. :P

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -401,13 +405,38 @@ extension HTTPConnectionPool {
...
+                    preconditionFailure("Failed to create a connection that is unknown to us?")
+                }
+                _ = self.connections.closeConnection(at: index)
+
+                return .init(
+                    request: self.failAllRequests(reason: error),
+                    connection: .none
+                )
+            }
+
```
fabianfett, line 439: As stated before, we should never reach this point, as we never scheduled a backoff timer.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -401,13 +405,38 @@ extension HTTPConnectionPool {
             self.failedConsecutiveConnectionAttempts += 1
             self.lastConnectFailure = error
 
+            guard self.retryConnectionEstablishment else {
+                guard let (index, _) = self.connections.failConnection(connectionID) else {
+                    preconditionFailure("Failed to create a connection that is unknown to us?")
```
fabianfett, line 410: Please rephrase failure mode.
```
preconditionFailure("A connection attempt failed, that the state machine knows nothing about. Somewhere state was lost.")
```

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -219,6 +223,17 @@ extension HTTPConnectionPool {
 
             switch self.lifecycleState {
             case .running:
+                guard self.retryConnectionEstablishment else {
```
fabianfett, line 226: Instead of using the guard here it might be nicer to go with where clause in the switch...

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
```diff
@@ -401,13 +405,38 @@ extension HTTPConnectionPool {
             self.failedConsecutiveConnectionAttempts += 1
             self.lastConnectFailure = error
 
+            guard self.retryConnectionEstablishment else {
```
fabianfett, line 408: Why don't we check the current running state here? Like in the http1 case? This seems weird. I think we should add this here. If the pool is closing, we should not backoff anymore.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -975,13 +975,20 @@ extension HTTPClient.Configuration {
         /// an explicit eventLoopRequirement are sent, this number might be exceeded due to overflow connections.
         public var concurrentHTTP1ConnectionsPerHostSoftLimit: Int
 
+        /// If true, ``HTTPClient`` will try to create new connections on connection failure with an exponential backoff.
+        /// Requests will only fail after the ``HTTPClient/Configuration/Timeout-swift.struct/connect`` timeout exceeded.
+        /// If false, requests will fail immediately after a connection could not be established.
+        /// Defaults to `true`
```
fabianfett, line 981:
```suggestion
        /// If false, all requests that have no assigned connection will fail immediately after a connection could not be established.
        /// Defaults to `true`
```

fabianfett (changes requested):
And a bit more...

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
dnadoba: `waitingForConnectivity` is called if a `NWConnection` transitions into the waiting state which has nothing to do with our own connection retry logic. This means that it will be called if a NWConnection "fails" instead of `failedToCreateNewConnection` because the connection will do its own retry thing.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
dnadoba: Yeah agree this is weird but has nothing to do with this change as we don't check the state at all. We should tackle this in a separate PR as we likely want to add test.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
dnadoba: A where constraint only applies to the last case if multiple cases are defined e.g.
```swift
case .running, .shuttingDown where self.retryConnectionEstablishment == true:
```
the where constrain only applies to `.shuttingDown` but not `.running`. We had an issue in AHC which we debugged together and decided we should stop using where clauses in switch cases altogether because it is not obvious.

fabianfett: true, however here we only have one case here. Anyway, not a hill I want to die on. If we want to stick with an inline if/guard, I would prefer this to be an `if self.retryConnectionEstablishment {}` and then move the retry logic into the if.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
fabianfett: Can we please add a code comment, that this is only triggered in the `Network.framework` flow?

If we don't want to retry connection establishment, NW.framework should not go into the waiting state, should it?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
fabianfett: I think, if we don't want to wait for connection establishment, we should never wait for connectivity in NW. I think I would prefer a precondition with a comment here.

fabianfett (commented):
We need TODOs for the incorrect http2 shutdown behavior, and we should change handling in waiting for NW.framework and/or document it better.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift
dnadoba: https://github.com/swift-server/async-http-client/issues/638

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
dnadoba: What's the benefit of the `if`? Guard at least guarantees that we return, if can fall through which we don't want.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
dnadoba: I have removed the code again. We have a different flag (`networkFrameworkWaitForConnectivity`) for disabling Network.frameworks waiting behaviour.

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
fabianfett: I consider it nicer to read. But that's personal...

fabianfett (approved):
Thanks for pushing through!

PR #626, opened 2022-09-06
add a future-returning shutdown method

weissi:


Lukasa (changes requested):
Thanks for this! I'm happy with this patch, but can we place the code in `HTTPClient.swift` instead of in the AsyncAwait directory?

weissi:
whoops, didn't see that. good catch, fixed

Lukasa (approved):

PR #630, opened 2022-09-28
Raise minimum supported Swift version from 5.4 to 5.5

glbrntt:
Motivation:

SwiftNIO periodically drops support for older Swift versions. Now that 5.7 has been released, 5.4 will be dropped.

Modifications:

- Remove 5.4 specific Package.swift and docker-compose
- Update the 5.7 docker-compose to use the released 5.7 and move from focal (2004) to jammy (2204)
- Update tools version in Package@swift-5.5.swift to 5.5 (from 5.4)
- Add supported versions section to README

Results:

Minimum Swift version is 5.5


Lukasa:

Lukasa (approved):
I have a sneaking suspicion that the 5.5 build is going to fail.

docker/docker-compose.2204.57.yaml
```diff
@@ -0,0 +1,17 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: async-http-client:22.04-5.7
+    build:
+      args:
+        base_image: "swift:5.7-jammy"
```
Lukasa, line 9: Can we use the same split here as elsewhere?

Lukasa (changes requested):

Lukasa (approved):

fabianfett (approved):

PR #631, opened 2022-10-04
Handle ResponseAccumulator not being able to buffer large response in memory

gjcairo:
## Motivation

If the response being buffered by the `ResponseAccumulator` is too large (larger than `UInt32.max`), a crash will occur.

## Modifications

* Check that the number of bytes being written won't cause the buffer to go over the limit. If it does, return a failed future from `didReceiveBodyPart(task:_:)`
* Created a new `ResponseTooBigError` to be used with the failed future.

## Result

Will now fail graciously if too big a response is buffered in memory via the `ResponseAccumulator`.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -367,8 +367,8 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
         case error(Error)
     }
 
-    struct ResponseTooBigError: Error, CustomStringConvertible {
-        var description: String {
+    public struct ResponseTooBigError: Error, CustomStringConvertible {
+        public var description: String {
             return "ResponseTooBigError: writing response part would exceed internal storage capacity."
         }
     }
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -367,6 +367,12 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
         case error(Error)
     }
 
+    struct ResponseTooBigError: Error, CustomStringConvertible {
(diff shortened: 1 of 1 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -367,6 +367,12 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
         case error(Error)
     }
 
+    struct ResponseTooBigError: Error, CustomStringConvertible {
```
Lukasa, line 370: We'll need this to be `public`.

gjcairo: Ah, so that it can be handled by the (failed) future's consumer?

Lukasa: Yeah, in general it's a good idea to make errors public if they're likely to be thrown.

Lukasa:

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -397,6 +403,10 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
         case .head(let head):
             self.state = .body(head, part)
         case .body(let head, var body):
+            if body.writerIndex + part.readableBytes > UInt32.max {
```
dnadoba, line 406: `UInt32.max` is a bit magic. I think the best solution would be to add a static property in `swift-nio` to `ByteBuffer` with the max size it can handle. But we can also just give this constant a descriptive name here and document it.

dnadoba (changes requested):

dnadoba:
closing in favor of https://github.com/swift-server/async-http-client/pull/637

PR #632, opened 2022-10-04
Make `HTTPClientResponse.init` public

dnadoba:
### Motivation
We want to allow user to initialise `HTTPClientResponse` for mocking purposes in tests.

### Modifications

- add `AnyAsyncSequence<Element>` to erase the type of the user provided `AsyncSequence`s. We need this because we do not want to make `HTTPClientResponse.Body` generic over the specific `AsyncSequence`.
- rename the old implementation of `HTTPClientResponse.Body` to `TransactionBody` and make it internal.
- remove no longer necessary type hiding in `TransactionBody` as it is internal. 
- add `Either<A, B>` and conditionally conform it to `AsyncSequence` if `A` and `B` are are `AsyncSequence`s with the same `Element` type. We use it to still allow the compiler to specialize the `HTTPClientResponse.Body` default case where the `AsyncSequence` is of type `TransactionBody`
- use `Either<TransactionBody, AnyAsyncSeqence<ByteBuffer>` as the new storage type for `HTTPClientReponse.Body`. We still hide the concrete storage type behind a struct.
- move `AsyncSequenceFromSyncSequence` from `AsyncHTTPClientTests` to `AsyncHTTPClient` and make it inlinable. (we could replace the implementation with [`AsyncLazySequence` from `swift-async-algorithms`](https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncAlgorithms/AsyncAlgorithms.docc/Guides/Lazy.md))
- Add there new public `HTTPClientResponse.Body` inits:
```swift
extension HTTPClientResponse.Body { 
    init() { ... }
    init(_ byteBuffer: ByteBuffer) { ... }
    init<SequenceOfBytes>(
        _ sequenceOfBytes: SequenceOfBytes
    ) where SequenceOfBytes: AsyncSequence & Sendable, SequenceOfBytes.Element == ByteBuffer { ... }
}
```
- Finally, add a new public init to `HTTPClientResponse`:
```swift
extension HTTPClientReponse {
    init(){
        self.version = .http1_1
        self.status = .ok
        self.headers = [:]
        self.body = Body()
    }
}
```

### Result
Users can now freely initialise and modify `HTTPClientReponse`s:
```swift
var response = HTTPClientResponse()
response.headers = [
    "Content-Type": "text/html"
]
response.body = .init(ByteBuffer(string: "<html> .... </html>"))
```


Sources/AsyncHTTPClient/AsyncAwait/AnyAsyncSequence.swift
```diff
@@ -0,0 +1,48 @@
...
+        }
+        
+        @inlinable mutating func next() async throws -> Element? {
+            try await nextCallback()
+        }
+    }
+    
+    @usableFromInline var makeAsyncIteratorCallback: @Sendable () -> AsyncIteratorNextCallback
+    
+    @inlinable public init<SequenceOfBytes>(
```
Lukasa, line 34:
```suggestion
    @inlinable init<SequenceOfBytes>(
```

Sources/AsyncHTTPClient/AsyncAwait/AsyncSequenceFromSyncSequence.swift
```diff
@@ -0,0 +1,46 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2022 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/AsyncHTTPClient/AsyncAwait/AsyncSequenceFromSyncSequence.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+    
+    @inlinable func makeAsyncIterator() -> AsyncIterator {
+        .init(iterator: self.wrapped.makeIterator())
+    }
+}
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension Sequence where Self: Sendable {
+    /// Turns `self` into an `AsyncSequence` by wending each element of `self` asynchronously.
```
Lukasa, line 42:
```suggestion
    /// Turns `self` into an `AsyncSequence` by vending each element of `self` asynchronously.
```

Sources/AsyncHTTPClient/AsyncAwait/AsyncSequenceFromSyncSequence.swift
```diff
@@ -0,0 +1,46 @@
...
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+@usableFromInline
+struct AsyncSequenceFromSyncSequence<Wrapped: Sequence & Sendable>: AsyncSequence, Sendable {
+    @usableFromInline typealias Element = Wrapped.Element
```
Lukasa, line 18: `Element` also needs to conform to `Sendable`, I believe.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -33,98 +33,74 @@ public struct HTTPClientResponse: Sendable {
...
-
-        public mutating func next() async throws -> ByteBuffer? {
-            try await self.stream.next()
-        }
-    }
-
-    public func makeAsyncIterator() -> AsyncIterator {
-        AsyncIterator(stream: IteratorStream(bag: self.bag))
+    
+    @inlinable public init(){
```
Lukasa, line 48:
```suggestion
    @inlinable public init() {
```

Lukasa: Is there any reason not to make this a memberwise initializer with these defaults already provided?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -33,98 +33,74 @@ public struct HTTPClientResponse: Sendable {
...
-            self.transaction = transaction
+extension HTTPClientResponse {
+    /// A representation of the response body for an HTTP response.
+    ///
+    /// The body is streamed as an `AsyncSequence` of `ByteBuffer`, where each `ByteBuffer` contains
+    /// an arbitrarily large chunk of data. The boundaries between `ByteBuffer` objects in the sequence
+    /// are entirely synthetic and have no semantic meaning.
+    public struct Body: AsyncSequence, Sendable {
+        public typealias Element = ByteBuffer
+        @usableFromInline typealias Storage = Either<TransactionBody, AnyAsyncSequence<ByteBuffer>>
```
Lukasa, line 65: TBH, I think the use of `Either` is excessively clever, and it leads to some unclear code (`case a` and `case b`). I'd rather see a specific-case enum for this use-case.

FranzBusch (commented):
In general, this looks like a fine approach to me. However, I would like to discuss the similarities and differences between the approach we took in gRPC and here. In gRPC we are just providing a testing interface which you can use to drive the sequence. Here we allow the user to provider their own `AsyncSequence`.

The former makes it slightly easier to get started testing since you can just use that interface to `yield` and `finish`. The latter provides a bit more customisation because you can drive more than just the `next` method, i.e. also the `makeAsyncIterator`.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
dnadoba: We eventually want to add support for trailing headers and this would require us to add another `init` for it if we want to allow trailing headers in the init too. 
However, I'm fine with adding a member wise init already today if you are too. We aren't likely to add any more properties to this struct (famous last words) and one additional init in the future isn't too bad.

dnadoba:
I think that in AHC we usually do not want to write individual `ByteBuffer`s to the stream. My guess is that the most common use case will be to just init the `Body` with a complete `ByteBuffer`. 

We may also want to think about aligning the Request and Response API in AHC.
A ` HTTPClientRequest.Body` can be created in the following ways:
```swift
extension HTTPClientRequest.Body {
    public static func bytes(_ byteBuffer: ByteBuffer) -> Self

    public static func bytes<Bytes: RandomAccessCollection & Sendable>(
        _ bytes: Bytes
    ) -> Self where Bytes.Element == UInt8 {
        Self._bytes(bytes)
    }
    public static func bytes<Bytes: Sequence & Sendable>(
        _ bytes: Bytes,
        length: Length
    ) -> Self where Bytes.Element == UInt8

    public static func bytes<Bytes: Collection & Sendable>(
        _ bytes: Bytes,
        length: Length
    ) -> Self where Bytes.Element == UInt8

    public static func stream<SequenceOfBytes: AsyncSequence & Sendable>(
        _ sequenceOfBytes: SequenceOfBytes,
        length: Length
    ) -> Self where SequenceOfBytes.Element == ByteBuffer
    
    public static func stream<Bytes: AsyncSequence & Sendable>(
        _ bytes: Bytes,
        length: Length
    ) -> Self where Bytes.Element == UInt8
}
```

We could provide a similar interface for `HTTPClientResponse.Body` excluding the `length` parameter.

Sources/AsyncHTTPClient/AsyncAwait/TransactionBody.swift
```diff
@@ -0,0 +1,65 @@
...
+        deinit {
+            self.transaction.responseBodyIteratorDeinited(streamID: self.id)
+        }
+        // TODO: this should be @inlinable
+        @usableFromInline func next() async throws -> ByteBuffer? {
+            try await self.transaction.nextResponsePart(streamID: self.id)
+        }
+    }
+
+    @inlinable func makeAsyncIterator() -> AsyncIterator {
```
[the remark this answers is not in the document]
dnadoba: We do something similar already here:
https://github.com/swift-server/async-http-client/blob/f17a47e91684541c84a0cdd8864af68e7e30c463/Sources/AsyncHTTPClient/AsyncAwait/Transaction%2BStateMachine.swift#L641-L653
This implementation is a bit more clever. It will allow the creation of multiple iterators and will only fail slightly delayed if someone tries to iterate over more than one of them. I think this is actually too clever and we should remove this eventually and just fail early during the creation of a second iterator. I'm not aware of any use case for creating unused iterators but maybe @fabianfett has some in mind.

Sources/AsyncHTTPClient/AsyncAwait/AsyncSequenceFromSyncSequence.swift
```diff
@@ -0,0 +1,52 @@
...
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension AsyncLazySequence: Sendable where Base: Sendable {}
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension AsyncLazySequence.AsyncIterator: Sendable where Base.Iterator: Sendable {}
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension Sequence {
+    /// Turns `self` into an `AsyncSequence` by vending each element of `self` asynchronously.
+    @inlinable var async: AsyncLazySequence<Self> {
```
fabianfett, line 49: This modifier already exists, doesn't it? I see it is local here, but I assume that we want to make it public eventually...

Sources/AsyncHTTPClient/AsyncAwait/SingleIteratorPrecondition.swift
```diff
@@ -0,0 +1,43 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Atomics
+
+/// Makes sure that a consumer of this `AsyncSequence`
```
Lukasa, line 17: Make sure they what?

Sources/AsyncHTTPClient/AsyncAwait/TransactionBody.swift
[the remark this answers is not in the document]
Lukasa: I'm inclined to agree with Fabian: there's no good sense in breaking the use-case that already exists, even though it'd be better defined if we did.

Lukasa (approved):

Sources/AsyncHTTPClient/AsyncAwait/TransactionBody.swift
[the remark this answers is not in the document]
dnadoba: Changed it back to the old behaviour. However, we now only enforce this on user provided `AsyncSequence`s.  Can also change that to use the more relaxed precondition if you want.

FranzBusch (approved):

PR #633, opened 2022-10-06
Fix request hang if delegate fails promise returned by `didReceiveBodyPart`

dnadoba:
`RequestBag.consumeMoreBodyData0(resultOfPreviousConsume:)` should be called for errors thrown by the delegate instead of `RequestBag.fail(_:)`.

Test added in the first commit hangs without the second commit which calls the right method.


Lukasa (approved):
LGTM.

fabianfett (approved):
Awesome fix! Could not have it done it better :P

PR #637, opened 2022-10-07
Handle ResponseAccumulator not being able to buffer large response in memory

dnadoba:
## Motivation

If the response being buffered by the `ResponseAccumulator` is too large (larger than `UInt32.max`), a crash will occur.

## Modifications

* Introduce an new `public var maxBodySize: Int` on `ResponseAccumulator`
* throw a new `ResponseTooBigError` if the response bodies size is larger than `maxBodySize`

## Result

Will now fail graciously if too big a response is buffered in memory via the `ResponseAccumulator`.
In addition, a user can configure the max size themself.


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -367,16 +367,48 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
...
+
+    /// Maximum size in bytes of the HTTP response body that ``ResponseAccumulator`` will accept
+    /// until it will abort the request and throw an ``ResponseTooBigError``.
+    ///
+    /// Default is 2^32.
+    /// - precondition: not allowed to exceed 2^32 because ``ByteBuffer`` can not store more bytes
+    public var maxBodySize: Int = maxByteBufferSize {
+        didSet {
+            precondition(
+                self.maxBodySize <= Self.maxByteBufferSize,
```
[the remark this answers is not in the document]
dnadoba: Added in https://github.com/swift-server/async-http-client/pull/637/commits/f0947e948d5bea5a46ac28bd608b90710ef8b55c

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -367,16 +367,49 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
         case error(Error)
     }
 
+    public struct ResponseTooBigError: Error, CustomStringConvertible {
+        var maxBodySize: Int
```
Lukasa, line 371: This should be `public`, and we should add a `public init`.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -367,16 +367,49 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
...
     let request: HTTPClient.Request
 
+    static let maxByteBufferSize = Int(UInt32.max)
+
+    /// Maximum size in bytes of the HTTP response body that ``ResponseAccumulator`` will accept
+    /// until it will abort the request and throw an ``ResponseTooBigError``.
+    ///
+    /// Default is 2^32.
+    /// - precondition: not allowed to exceed 2^32 because ``ByteBuffer`` can not store more bytes
+    public var maxBodySize: Int = maxByteBufferSize {
```
Lukasa, line 388: This isn't a great pattern: it risks thread-safety issues when mutated. We should take it in the constructor and make it immutable.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -367,16 +367,49 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
...
         self.request = request
     }
 
     public func didReceiveHead(task: HTTPClient.Task<Response>, _ head: HTTPResponseHead) -> EventLoopFuture<Void> {
         switch self.state {
         case .idle:
+            if let contentLength = head.headers.first(name: "Content-Length"),
+               let announcedBodySize = Int(contentLength),
+               announcedBodySize > self.maxBodySize {
+                let error = ResponseTooBigError(maxBodySize: maxBodySize)
```
Lukasa, line 408: We want to be a _little_ careful here: if this is a response to a HEAD request, we don't expect a response body at all so we don't need to worry about this.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -367,16 +367,49 @@ public class ResponseAccumulator: HTTPClientResponseDelegate {
...
+
     public init(request: HTTPClient.Request) {
         self.request = request
     }
 
     public func didReceiveHead(task: HTTPClient.Task<Response>, _ head: HTTPResponseHead) -> EventLoopFuture<Void> {
         switch self.state {
         case .idle:
+            if let contentLength = head.headers.first(name: "Content-Length"),
+               let announcedBodySize = Int(contentLength),
```
Lukasa, line 406: Should we be tolerating the possibility of a non-integer Content-Length field here?

Lukasa (changes requested):

Sources/AsyncHTTPClient/HTTPHandler.swift
dnadoba: That's at least what we do in swift-nio so far:
https://github.com/apple/swift-nio/blob/41f0b2d3983038ae68bbb902ea43e123a5466494/Sources/NIOHTTP1/HTTPTypes.swift#L1508

Lukasa (approved):

PR #639, opened 2022-10-10
Speedup tests

dnadoba:
### Motiviation
We like fast tests.

### Modification
- set `networkFrameworkWaitForConnectivity` and `connectionPool.retryConnectionEstablishment` to `false` for most test cases
- move long running test cases into separate classes so they can run in parallel

### Results

| Mode      | Old                  | New     | Relative|
| :---        |    ---: |      ---: | ---: |
| Serial      | 180s (3m 0s) | 48s | 3.75x |
| Parallel   | 62s (1m 2s)    | 12s  | 5.1x |


FranzBusch (approved):
Nice stuff! Looks good to me!

PR #640, opened 2022-10-11
Replace `NIOSendable` with `Sendable`

dnadoba:
We now require Swift 5.5.2 which includes `Sendable` so we no longer require the `NIOSendable` workaround.
`NIOSendable` is just a `typealias` of `Sendable` so this is a non source breaking change.
This allows `swift-nio` to eventually deprecate `NIOSendable`.


Lukasa (commented):
What’s the rationale for this?

dnadoba:
I have updated the description.

Lukasa (approved):

PR #641, opened 2022-10-12
Remove `#if compiler(>=5.5)`

dnadoba:
### Motivation
We only support Swift 5.5.2+. 

### Modification
Remove a lot of our `#if swift(>=5.5)` conditional compilation blocks.

### Result
less branching


Tests/AsyncHTTPClientTests/TransactionTests.swift
```diff
@@ -567,7 +545,7 @@ actor SharedIterator<Wrapped: AsyncSequence> where Wrapped.Element: Sendable {
 
 /// non fail-able promise that only supports one observer
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-fileprivate actor Promise<Value> {
+private actor Promise<Value> {
```
dnadoba, line 548: The change from `fileprivate` to `private` is the result of running `SwiftFormat`

Lukasa (approved):

PR #643, opened 2022-10-26
FileDownloadDelegate: mark `Progress` as `Sendable`

MaxDesiatov:
This is a trivial struct that should be `Sendable`, as it's a common use case to pass `Progress` values to `@Sendable` closures.


Lukasa (approved):
Thanks!

Lukasa:

PR #645, opened 2022-11-03
Remove obsolete notes in README.md

dnadoba:
AsyncHTTPClient now requires at least Swift 5.5.2. The async/await APIs are therefore available in all supported Swift compiler versions.

Swift Concurrency and HTTP/2 are now also supported for some time now. No need to call out the specific version in the readme. This information can still be found in the release notes if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,19 +2,13 @@
 This package provides an HTTP Client library built on top of SwiftNIO.
 
 This library provides the following:
-- First class support for Swift Concurrency (since version 1.9.0)
+- First class support for Swift Concurrency
 - Asynchronous and non-blocking request methods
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,20 +2,14 @@
 This package provides an HTTP Client library built on top of SwiftNIO.
 
 This library provides the following:
-- First class support for Swift Concurrency (since version 1.9.0)
+- First class support for Swift Concurrency
```
PeterAdams-A, line 5: I was tempted to suggest leaving the version numbers here but seeing as the inclusion code snippet before uses `from: "1.9.0"` it feels consistent to drop it.

PeterAdams-A (approved):
Seems fine to me.

PR #646, opened 2022-11-08
Tolerate shutdown message after channel is closed

dnadoba:
### Motivation
A channel can close unexpectedly if something goes wrong. We may in the meantime have scheduled the connection for graceful shutdown but the connection has not yet seen the message. We need to still accept the shutdown message and just ignore it if we are already closed.

### Modification
- ignore calls to shutdown if the channel is already closed
- add a test which would previously crash because we have transition from the closed state to the closing state and we hit the deinit precondition
- include the current state in preconditions if we are in the wrong state

### Result

We don’t hit the precondition in the deinit in the scenario described above and have more descriptive crashes if something still goes wrong.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
@@ -129,7 +129,7 @@ final class HTTP2Connection {
             delegate: delegate,
             logger: logger
         )
-        return connection.start().map { maxStreams in (connection, maxStreams) }
+        return connection.start0().map { maxStreams in (connection, maxStreams) }
     }
 
     func executeRequest(_ request: HTTPExecutableRequest) {
@@ -164,7 +164,7 @@ final class HTTP2Connection {
         return promise.futureResult
     }
 
-    func start() -> EventLoopFuture<Int> {
+    func start0() -> EventLoopFuture<Int> {
         self.channel.eventLoop.assertInEventLoop()
 
         let readyToAcceptConnectionsPromise = self.channel.eventLoop.makePromise(of: Int.self)
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
@@ -267,7 +267,7 @@ final class HTTP2Connection {
         self.channel.eventLoop.assertInEventLoop()
 
         switch self.state {
-        case .active, .starting:
+        case .active:
             self.state = .closing
 
             // inform all open streams, that the currently running request should be cancelled.
@@ -279,10 +279,11 @@ final class HTTP2Connection {
             // are closed.
             self.channel.triggerUserOutboundEvent(HTTPConnectionEvent.shutdownRequested, promise: nil)
             
-        case .closed:
-            // we are already closed and we need to tolerate this
+        case .closed, .closing:
+            // we are already closing/closed and we need to tolerate this
             break
-        case .initialized, .closing:
+            
+        case .initialized, .starting:
             preconditionFailure("invalid state \(self.state)")
         }
     }
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
@@ -111,7 +111,7 @@ final class HTTP2Connection {
 
     deinit {
         guard case .closed = self.state else {
-            preconditionFailure("Connection must be closed, before we can deinit it")
+            preconditionFailure("Connection must be closed, before we can deinit it. Current state: \(self.state)")
         }
     }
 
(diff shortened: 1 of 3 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -258,16 +266,25 @@ final class HTTP2Connection {
...
-        self.channel.triggerUserOutboundEvent(HTTPConnectionEvent.shutdownRequested, promise: nil)
+            // inform the idle connection handler, that connection should be closed, once all streams
+            // are closed.
+            self.channel.triggerUserOutboundEvent(HTTPConnectionEvent.shutdownRequested, promise: nil)
+            
+        case .closed:
+            // we are already closed and we need to tolerate this
+            break
+        case .initialized, .closing:
+            preconditionFailure("invalid state \(self.state)")
```
fabianfett, line 286: I think we should also support shutdown calls if we are already closing. No reason to crash here. Also do we want to support shutdown, if we haven't even connected? We could go straight to `closed`, which is the only way to drop a connection currently.

Tests/AsyncHTTPClientTests/HTTP2ConnectionTests.swift
```diff
@@ -41,6 +41,28 @@ class HTTP2ConnectionTests: XCTestCase {
...
+        
+        let logger = Logger(label: "test.http2.connection")
+        let connection = HTTP2Connection(
+            channel: embedded,
+            connectionID: 0,
+            decompression: .disabled,
+            delegate: TestHTTP2ConnectionDelegate(),
+            logger: logger
+        )
+        _ = connection.start()
```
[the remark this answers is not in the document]
dnadoba: The problem is that the only way to for the the future returned by the `HTTP2Connection.start()` to fulfill is by sending `ByteBuffer`s of `HTTP2Frame`s through the `Channel`. `HTTP2Frame` de/encoding is currently not publicly available.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
dnadoba: Agree, changed in https://github.com/swift-server/async-http-client/pull/646/commits/653858d01b989f9f206d29ebf9e87f0898dd01b3

Tests/AsyncHTTPClientTests/HTTP2ConnectionTests.swift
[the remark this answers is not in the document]
fabianfett: But we don't need that future, or am I missing something here?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -258,16 +266,26 @@ final class HTTP2Connection {
...
+            // inform the idle connection handler, that connection should be closed, once all streams
+            // are closed.
+            self.channel.triggerUserOutboundEvent(HTTPConnectionEvent.shutdownRequested, promise: nil)
+
+        case .closed, .closing:
+            // we are already closing/closed and we need to tolerate this
+            break
+
+        case .initialized, .starting:
+            preconditionFailure("invalid state \(self.state)")
```
fabianfett, line 287: Is there a reason for us not to support going from initialized straight to closed?

Tests/AsyncHTTPClientTests/HTTP2ConnectionTests.swift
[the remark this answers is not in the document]
dnadoba: We do need it as only its success value contains the `HTTP2Connection` instance. We otherwise don't have access to it and therefore can't call `.shutdown()`.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -164,15 +164,23 @@ final class HTTP2Connection {
         return promise.futureResult
     }
 
-    private func start() -> EventLoopFuture<Int> {
+    func start() -> EventLoopFuture<Int> {
```
fabianfett, line 167: and we should follow the internal naming pattern here: `start0`. as a small cleanup.

Tests/AsyncHTTPClientTests/HTTP2ConnectionTests.swift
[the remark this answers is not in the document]
fabianfett: stupid me... makes sense. can we rename it to start0 though. to make sure it is called on el, if someone else internally tries to call it?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
dnadoba: It is an invalid transition. We require a call to `start()` before any other interaction which transitions to the starting case. Further interaction is only allowed after the promise returned by `.start()` is completed which transitions to either the `.active` or `.closed` state. This makes sure we don't violate this contract.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
dnadoba: changed in https://github.com/swift-server/async-http-client/pull/646/commits/ba76fe445fcd25559a564364229c742db8c63316

Tests/AsyncHTTPClientTests/HTTP2ConnectionTests.swift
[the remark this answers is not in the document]
dnadoba: Sure changed in https://github.com/swift-server/async-http-client/pull/646/commits/ba76fe445fcd25559a564364229c742db8c63316

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -164,15 +164,23 @@ final class HTTP2Connection {
         return promise.futureResult
     }
 
-    private func start() -> EventLoopFuture<Int> {
+    func start0() -> EventLoopFuture<Int> {
```
Lukasa, line 167: Nit: should we give this an underscore?

Lukasa (approved):
One very minor nit, otherwise this looks great!

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
dnadoba: Added in https://github.com/swift-server/async-http-client/pull/646/commits/510d6355ea9820d5509e93e428f42e3918479693

Issue #650, opened 2022-12-06
Doesn't work with usefixie and quotaguard with `invalidProxyResponse `


MihaelIsaev:
I tried to use it with usefixie.com and quotaguard.com and getting `invalidProxyResponse` error.

I tried to set breakpoints to debug the issue and see that under the hood these services returns `400 Bad Request`

```swift
let auth = HTTPClient.Authorization.basic(username: "fixie", password: "___")
let proxyConfig = HTTPClient.Configuration.Proxy.server(host: "___.usefixie.com", port: 80, authorization: auth)
app.http.client.configuration.proxy = proxyConfig
//app.http.client.configuration.redirectConfiguration = .follow(max: 100, allowCycles: true)
do {
    let res = try app.client.get(URI("https://ip.quotaguard.com/")).wait()
    app.logger.critical("\(res.body?.string)")
} catch {
    app.logger.critical("Error: \(error)") // getting HTTPClientError.invalidProxyResponse here
}
```

at the same time with nodejs everything works fine

```js
var http = require('http');

function buildAuthHeader(user, pass) {
    return 'Basic ' + new Buffer(user + ':' + pass).toString('base64');
}

proxy = '___.usefixie.com';
proxy_port = 80;
host = 'ip.quotaguard.com';
url = 'https://ip.quotaguard.com';
user = 'fixie';
pass = '___';

var options = {
    port: proxy_port,
    host: proxy,
    path: url,
    headers: {
        Host: host,
        'Proxy-Authorization': buildAuthHeader(user, pass),
    }
};

http.get(options, function(res) {
  console.log("StatusCode: " + res.statusCode + " Message: " + res.statusMessage);
  let chunks = []
    res.on('data', chunk => chunks.push(chunk))
    res.on('end', () => {
      console.log('DONE', Buffer.concat(chunks).toString('utf8'))
    })
}).end();
```

Lukasa:
Can you use Wireshark or something similar to capture the request to the proxy and then provide it here?

MihaelIsaev:
@Lukasa thanks for pointing me to Wireshark, I completely forgot about it!

I realized that `Host` header is missing which causes `400 Bad Request: missing required Host header`

I added missing header into `HTTP1ProxyConnectionHandler` and now it works like a charm!

MihaelIsaev:
@Lukasa thank you very much for merging it that fast! Need it in production asap, could you please tag it?

PR #651, opened 2022-12-06
Add `Host` header (#650)

MihaelIsaev:
I realized that `Host` header is missing which causes `400 Bad Request: missing required Host header`.

I added missing header into `HTTP1ProxyConnectionHandler` and now it works like a charm!

Also added `Host` header check into tests.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
@@ -155,6 +155,7 @@ final class HTTP1ProxyConnectHandler: ChannelDuplexHandler, RemovableChannelHand
             method: .CONNECT,
             uri: "\(self.targetHost):\(self.targetPort)"
         )
+        head.headers.replaceOrAdd(name: "host", value: "\(self.targetHost)")
         if let authorization = self.proxyAuthorization {
             head.headers.replaceOrAdd(name: "proxy-authorization", value: authorization.headerValue)
         }
--- a/Tests/AsyncHTTPClientTests/HTTP1ProxyConnectHandlerTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTP1ProxyConnectHandlerTests.swift
@@ -43,6 +43,7 @@ class HTTP1ProxyConnectHandlerTests: XCTestCase {
 
         XCTAssertEqual(head.method, .CONNECT)
         XCTAssertEqual(head.uri, "swift.org:443")
+        XCTAssertEqual(head.headers["host"].first, "swift.org")
         XCTAssertNil(head.headers["proxy-authorization"].first)
         XCTAssertEqual(try embedded.readOutbound(as: HTTPClientRequestPart.self), .end(nil))
 
@@ -76,6 +77,7 @@ class HTTP1ProxyConnectHandlerTests: XCTestCase {
 
         XCTAssertEqual(head.method, .CONNECT)
         XCTAssertEqual(head.uri, "swift.org:443")
+        XCTAssertEqual(head.headers["host"].first, "swift.org")
         XCTAssertEqual(head.headers["proxy-authorization"].first, "Basic abc123")
         XCTAssertEqual(try embedded.readOutbound(as: HTTPClientRequestPart.self), .end(nil))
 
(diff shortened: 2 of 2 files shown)


Lukasa (approved):
Really nice change, thanks!

Lukasa:

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -155,6 +155,7 @@ final class HTTP1ProxyConnectHandler: ChannelDuplexHandler, RemovableChannelHand
             method: .CONNECT,
             uri: "\(self.targetHost):\(self.targetPort)"
         )
+        head.headers.replaceOrAdd(name: "host", value: "\(self.targetHost)")
```
dnadoba, line 158: This is not the right host in all cases. We allow users to overwrite the host and also allow non-default ports.
We already have a method which can take care of this for us:
https://github.com/swift-server/async-http-client/blob/49abfc30ae80951dbd9e8d12bef6d7722a4837dd/Sources/AsyncHTTPClient/RequestValidation.swift#L114-L129

MihaelIsaev: It seems I can't use this nice method cause I don't have `DeconstructedURL` in this context.

Tried to build `DeconstructedURL` but it requires many lines of code.
I just implemented check on place using what we have.

Sending one more pull request #652 since this one is closed already.
Please let me know if I should build `DeconstructedURL` and use `addHostIfNeeded ` instead.

MihaelIsaev: @dnadoba could you please take a look at [#652](https://github.com/swift-server/async-http-client/pull/652)

Lukasa: I'm pretty sure this is the right host in all cases: the Host header for CONNECT should be the Host of the target URI, the same thing we put in the request line. That's what this should be.

dnadoba: Given the following request:
```swift
var request = HTTPClientRequest(url: "http://127.0.0.1/")
request.headers.add(name: "host", value: "swift.org")
let response = try await httpClient.execute(request, timeout: .seconds(10))
```
This will currently send a request header with 
```
Host: swift.org
``` 
and **not** 
```
Host: 127.0.0.1
```
This can be useful to e.g. to test a local nginx configuration without modifying `/etc/hosts`. AFAIK browsers don't support this for security reasons.

If we now add a CONNECT proxy into the mix with the following request:
```swift
var configuration = HTTPClient.Configuration()
configuration.proxy = .server(host: "example.com", port: 80)
var request = HTTPClientRequest(url: "http://127.0.0.1/")
request.headers.add(name: "host", value: "swift.org")
```
we will make a connection to `example.com` and send it this request:
```
CONNECT 127.0.0.1:80 HTTP/1.1
Host: 127.0.0.1
```
and afterwards will send a HTTP request with the following header:
```
Host: swift.org
```

The [HTTP/1.1 RFC](https://www.rfc-editor.org/rfc/rfc9110) doesn't really mention what the `Host` headers value should be for in the initial CONNECT request. I thought that the Host should be the same as the Host in the subsequent request. This would allow the proxy to connect you to a server that can handle the request. However I might be wrong, that is just a guess on my side.

curl doesn't use the `Host` in the header but will use 
```
% curl -v -curl -H "Host: swift.org" --proxy "http://[::1]:8888" https://127.0.0.1/ 
*   Trying ::1:8888...
* Connected to ::1 (::1) port 8888 (#0)
* allocate connect buffer
* Establish HTTP proxy tunnel to 127.0.0.1:443
> CONNECT 127.0.0.1:443 HTTP/1.1
> Host: 127.0.0.1:443
> User-Agent: curl/7.85.0
> Proxy-Connection: Keep-Alive
```

I have tried setting up a proxy with `URLSession` to also see its behaviour but have failed to do so.

I think we should at least follow curl here and also use the host specified in the request URL for the initial CONNECT and not the host & port of the proxy. 

What you you think @Lukasa?

Lukasa: We are following curl here, no? Looking at the test, we appear to be doing exactly the same thing as curl.

PR #652, opened 2022-12-06
Fix `Host` header (#650)

MihaelIsaev:
Added check to add host only if needed, and also building host the right way.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
@@ -155,6 +155,7 @@ final class HTTP1ProxyConnectHandler: ChannelDuplexHandler, RemovableChannelHand
             method: .CONNECT,
             uri: "\(self.targetHost):\(self.targetPort)"
         )
+        head.headers.replaceOrAdd(name: "host", value: "\(self.targetHost)")
         if let authorization = self.proxyAuthorization {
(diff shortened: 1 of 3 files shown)


Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/HTTP1ProxyConnectHandler.swift
```diff
@@ -155,7 +160,13 @@ final class HTTP1ProxyConnectHandler: ChannelDuplexHandler, RemovableChannelHand
             method: .CONNECT,
             uri: "\(self.targetHost):\(self.targetPort)"
         )
-        head.headers.replaceOrAdd(name: "host", value: "\(self.targetHost)")
+        if !head.headers.contains(name: "host") {
```
dnadoba, line 163: The `HTTPRequestHead` is created in the lines above and we can therefore say for sure we will never have a `host` set. This makes sense because we don't need to look into the headers we have created here but from the original HTTP request from the user. This will require plumping the `HTTPHeader`s from the original request through the stack. 

FYI: @rnro is currently working on moving `HTTP1ProxyConnectHandler` to `swift-nio-extras` [in this PR](https://github.com/apple/swift-nio-extras/pull/185). In this new version the `HTTPHeader`s are passed into the `HTTP1ProxyConnectHandler.init` which makes this a bit easier. We still need to get the original request headers to where we create the `HTTP1ProxyConnectHandler` in AHC.

dnadoba (changes requested):
Thanks! This is almost good to go.

PR #657, opened 2023-01-24
Fix HTTP2StreamChannel leak

fabianfett:
We leak `HTTP2StreamChannel`s in long living HTTPClients. The reason for this is that we registered the callback to remove an `HTTP2StreamChannel` from the openStreams set in HTTP2Connection on the HTTP2 connection channel and not on the `HTTP2StreamChannel` itself. (Side note: `self.channel` vs local `var channel` is always crazy dangerous)

### Changes

- Actual bug fix (remove a `self.`)
- Ensuring that we always close the `HTTP2StreamChannel` when the request has finished
- Add a test for a previous uncovered code path (request receives full response before being done writing)

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
@@ -238,6 +238,9 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
     private func runSuccessfulFinalAction(_ action: HTTPRequestStateMachine.Action.FinalSuccessfulRequestAction, context: ChannelHandlerContext) {
         switch action {
         case .close, .none:
+            // The actions returned here come from an `HTTPRequestStateMachine` that assumes http/1.1
+            // semantics. For this reason we can ignore the close here, since an h2 stream is closed
+            // after every request anyway.
             break
 
         case .sendRequestEnd(let writePromise):
@@ -246,8 +249,9 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
     }
 
     private func runFailedFinalAction(_ action: HTTPRequestStateMachine.Action.FinalFailedRequestAction, context: ChannelHandlerContext, error: Error) {
-        // We must close the http2 stream after the request has finished. This breaks a reference
-        // cycle in HTTP2Connection.
+        // We must close the http2 stream after the request has finished. Since the request failed,
+        // we have no idea what the h2 streams state was. To be on the save side, we explicitly close
+        // the h2 stream. This will break a reference cycle in HTTP2Connection.
         context.close(promise: nil)
 
         switch action {
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
@@ -239,21 +239,21 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
 
     private func runSuccessfulFinalAction(_ action: HTTPRequestStateMachine.Action.FinalSuccessfulRequestAction, context: ChannelHandlerContext) {
         switch action {
-        case .close:
-            context.close(promise: nil)
+        case .close, .none:
+            break
 
         case .sendRequestEnd(let writePromise):
             context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: writePromise)
-
-        case .none:
-            break
         }
     }
 
     private func runFailedFinalAction(_ action: HTTPRequestStateMachine.Action.FinalFailedRequestAction, context: ChannelHandlerContext, error: Error) {
+        // We must close the http2 stream after the request has finished. This breaks a reference
+        // cycle in HTTP2Connection.
+        context.close(promise: nil)
+
         switch action {
         case .close(let writePromise):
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
```diff
@@ -241,7 +241,7 @@ final class HTTP2Connection {
                     // before.
                     let box = ChannelBox(channel)
                     self.openStreams.insert(box)
-                    self.channel.closeFuture.whenComplete { _ in
+                    channel.closeFuture.whenComplete { _ in
```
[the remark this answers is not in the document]
dnadoba: I have run into this just today. If we send a too large header and the server sends a go away frame the child channel is closed but the parent channel not. This might be a good way to test this.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -239,21 +239,29 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
...
-            context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: writePromise)
-
-        case .none:
-            break
+            let promise = writePromise ?? context.eventLoop.makePromise(of: Void.self)
+            context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: promise)
+            promise.futureResult.whenComplete { _ in
+                // Once we have written the last message we must close the http2 stream, since this
+                // break a reference cycle in HTTP2Connection.
+                context.close(promise: nil)
```
Lukasa, line 253: I'm a little surprised that this is necessary. swift-nio-http2 should notify you that the stream is closed when both sides have sent their `.end`. Are you entirely confident this is required?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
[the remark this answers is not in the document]
Lukasa: I think we can try to test this by triggering the shutdown event and finding which channels get sent this user event. So long as we can hook the parent channel we should observe that it doesn't see it.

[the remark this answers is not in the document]
fabianfett: The question is: How do I get access to the stream/child channel? All the ideas I had so far a quite locked down. Of course I can add an internal function that gives me the currently active channels.

```
func forTestingGetActiveChildChannels -> EventLoopFuture<[Channel]>
```

wdyt?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
fabianfett: I'm not. I was only testing the child channel. And wanted to be as defensive as possible. Do you want me to remove it again?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2Connection.swift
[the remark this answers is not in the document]
Lukasa: I think adding a testing function is fine. You can guard it behind an SPI, which is probably sufficient.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
Lukasa: Yeah, I think we shouldn't add code that we don't really need.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -239,21 +239,21 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
 
     private func runSuccessfulFinalAction(_ action: HTTPRequestStateMachine.Action.FinalSuccessfulRequestAction, context: ChannelHandlerContext) {
         switch action {
-        case .close:
-            context.close(promise: nil)
+        case .close, .none:
```
glbrntt, line 242: `.close` no longer closes; should the state machine still produce this action?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -239,21 +239,21 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
...
-
-        case .none:
-            break
         }
     }
 
     private func runFailedFinalAction(_ action: HTTPRequestStateMachine.Action.FinalFailedRequestAction, context: ChannelHandlerContext, error: Error) {
+        // We must close the http2 stream after the request has finished. This breaks a reference
+        // cycle in HTTP2Connection.
+        context.close(promise: nil)
```
glbrntt, line 253: The `.none` action feels like a bit of a lie now since we always `close`. Should the state machine be updated to reflect that?

glbrntt (approved):
Looks good modulo a couple of nits.

fabianfett:
@glbrntt added/updated code comments to point out the http/1.1 semantics of the action

glbrntt (approved):

dnadoba (approved):

PR #658, opened 2023-01-25
Add test for HTTP1 request with large header

dnadoba:
### Motivation
We currently don't handle large headers well which trigger a channel writability change event.

### Modification
- refactor `HTTP1ConnectionHandler` to make it more testable
- Add failing (but currently skipped) tests which reproduces the issue

### Result
We can reliably reproduce the large request header issue in an integration and unit test.
Note that the actual fix is not included to make reviewing easier and will come in a follow up PR.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
@@ -108,6 +108,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
 
         let action = self.state.writabilityChanged(writable: context.channel.isWritable)
         self.run(action, context: context)
+        context.fireChannelWritabilityChanged()
     }
 
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
@@ -156,6 +157,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             metadata: req.requestFramingMetadata
         )
         self.run(action, context: context)
+        promise?.succeed(())
     }
 
     func read(context: ChannelHandlerContext) {
@@ -434,6 +436,11 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
         self.run(action, context: context)
     }
 }
+
+#if swift(>=5.6)
+@available(*, unavailable)
+extension HTTP1ClientChannelHandler: Sendable {}
+#endif
 
 extension HTTP1ClientChannelHandler: HTTPRequestExecutor {
     func writeRequestBodyPart(_ data: IOData, request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
@@ -35,7 +35,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
         didSet {
             if let newRequest = self.request {
                 var requestLogger = newRequest.logger
-                requestLogger[metadataKey: "ahc-connection-id"] = connectionIdLoggerMetadata
+                requestLogger[metadataKey: "ahc-connection-id"] = self.connectionIdLoggerMetadata
                 requestLogger[metadataKey: "ahc-el"] = "\(self.eventLoop)"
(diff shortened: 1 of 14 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -156,6 +157,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             metadata: req.requestFramingMetadata
         )
         self.run(action, context: context)
+        promise?.succeed(())
```
Lukasa, line 160: This doesn't seem appropriate. We should presumably send the promise on with the appropriate data, rather than silently succeeding it here.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -59,15 +59,15 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
...
     private var logger: Logger
+    private let eventLoop: EventLoop
+    private let connectionIdLoggerMetadata: Logger.MetadataValue
 
-    let connection: HTTP1Connection
-    let eventLoop: EventLoop
-
-    init(connection: HTTP1Connection, eventLoop: EventLoop, logger: Logger) {
-        self.connection = connection
+    var onRequestCompleted: () -> Void = {}
```
Lukasa, line 65: Why are we adding this indirection here?

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
dnadoba: Good catch. I have reverted it now. I though I needed it because `try channel.writeOutbound(request)` would block indefinitely but I have now replaced it with `channel.writeAndFlush(request, promise: nil)` which doesn't block and is closer to the production code.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
dnadoba: We need it to break the dependency between `HTTP1Connection` and `HTTP1ClientChannelHandler`. `HTTP1ClientChannelHandler` only needs a reference to `HTTP1Connection` to get the connection id for logging and to signal that the last request is complete. The `connection.id` is now passed as `connectionIdLoggerMetadata: Logger.MetadataValue` in the init. `connection.taskComplete()` is replaced with `onRequestCompleted()` indirection.

I think in the long run we should replace this with a write promise we pass together with the request here:
https://github.com/swift-server/async-http-client/blob/817d9aa99fe4c71e7bea18a9a74bc7a5fc166692/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1Connection.swift#L101
This will cost us an additional allocation per request but this should be fine.

Lukasa: Yes, this was my reaction: having unstructured callbacks like this is a bit of a worry for me. Promises are a substantial improvement.

Lukasa: Is there any reason not to make that change soon?

Lukasa: Ok, I'm asking only because I suspect this mystery-meat callback will live here forever if we don't make the change now.

dnadoba: So I have tried to do that. The problem I run into is that it is currently not really used like a promise and may never be called. It is used to signal that the connection is again able to receive further request which it may never be for various reasons. This case is currently handled by closing the channel instead of calling this callback. At the end it just calls down to `connection.delegate.http1ConnectionReleased(connection)`

I think changing and testing this is quite some effort which I would like not to do right now.

dnadoba: I have at least renamed it to `onConnectionIdle` to better explain what it actually does: https://github.com/swift-server/async-http-client/pull/658/commits/5f5dc1e3e5c2d0610bf0a8a18b83d04034cdd8b3

Lukasa (approved):

PR #661, opened 2023-01-25
Fix crash for large HTTP request headers

dnadoba:
### Motivation
Fixes tests added in #658 and #659

### Modification
Gracefully handle writability change events while headers are send.

### Result
Large headers can be send. This could actually happen randomly for small headers too, especially in the HTTP/2 case. They just need to fill the buffer and trigger back pressure.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
@@ -108,6 +108,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
 
         let action = self.state.writabilityChanged(writable: context.channel.isWritable)
         self.run(action, context: context)
+        context.fireChannelWritabilityChanged()
     }
 
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
@@ -156,6 +157,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             metadata: req.requestFramingMetadata
         )
         self.run(action, context: context)
+        promise?.succeed(())
     }
 
     func read(context: ChannelHandlerContext) {
@@ -434,6 +436,11 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
         self.run(action, context: context)
     }
 }
+
+#if swift(>=5.6)
+@available(*, unavailable)
+extension HTTP1ClientChannelHandler: Sendable {}
+#endif
 
 extension HTTP1ClientChannelHandler: HTTPRequestExecutor {
     func writeRequestBodyPart(_ data: IOData, request: HTTPExecutableRequest, promise: EventLoopPromise<Void>?) {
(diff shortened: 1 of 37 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -183,9 +183,19 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
 
     private func run(_ action: HTTP1ConnectionStateMachine.Action, context: ChannelHandlerContext) {
         switch action {
-        case .sendRequestHead(let head, startBody: let startBody):
-            self.sendRequestHead(head, startBody: startBody, context: context)
+        case .sendRequestHead(let head, let sendEnd):
+            self.sendRequestHead(head, sendEnd: sendEnd, context: context)
+        case .notifyRequestHeadSendSuccessfully(let resumeRequestBodyStream, let startIdleTimer):
 
+            self.request!.requestHeadSent()
```
Lukasa, line 190: Why is this force-unwrap safe?

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -183,9 +183,19 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
...
+            if resumeRequestBodyStream {
+                self.request!.resumeRequestBodyStream()
```
Lukasa, line 192: Relatedly, why is this safe? Can it end up `nil` after the preceding outcall?

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -140,9 +140,18 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
 
     private func run(_ action: HTTPRequestStateMachine.Action, context: ChannelHandlerContext) {
         switch action {
-        case .sendRequestHead(let head, let startBody):
-            self.sendRequestHead(head, startBody: startBody, context: context)
-
+        case .sendRequestHead(let head, let sendEnd):
+            self.sendRequestHead(head, sendEnd: sendEnd, context: context)
+        case .notifyRequestHeadSendSuccessfully(let resumeRequestBodyStream, let startIdleTimer):
+            self.request!.requestHeadSent()
```
Lukasa, line 146: Same notes here about the force unwraps.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
dnadoba: Added a comment

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
dnadoba: Great catch! This is not safe. I added a test which crashed previously bit is fixed by chaining the implantation to unwrap safely: https://github.com/swift-server/async-http-client/pull/661/commits/18f6505fd957e3ac2e8aeca148faa95f05987b38

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
dnadoba: Added a comment

Lukasa:

Lukasa (approved):

PR #668, opened 2023-02-13
Mark Task.wait() noasync and provide Task.get()

Lukasa:
Motivation

Task.wait() is a convenience method to avoid needing to wait for the response future. This has had the effect of "laundering" a noasync warning from EventLoopFuture.wait(), hiding it within a purely sync call that may itself be used in an async context.

We should discourage using these and prefer using .get() instead.

Modifications

Mark Task.wait() noasync.
Add Task.get() for backward compatibility.

Result

Safer migration to Swift concurrency


dnadoba (approved):

dnadoba:
You need to generate linux tests, otherwise LGTM.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -750,13 +750,33 @@ extension HTTPClient {
...
         }
 
+        #if swift(>=5.7)
         /// Waits for execution of this request to complete.
         ///
-        /// - returns: The value of the `EventLoopFuture` when it completes.
-        /// - throws: The error value of the `EventLoopFuture` if it errors.
+        /// - returns: The value of  ``futureResult`` when it completes.
+        /// - throws: The error value of ``futureResult`` if it errors.
+        @available(*, noasync, message: "wait() can block indefinitely, prefer get()", renamed: "get()")
```
ktoso, line 758: You should be able to just `#if` around the available attribute and avoid the duplicate impl, if you wanted to

ktoso (approved):

Sources/AsyncHTTPClient/HTTPHandler.swift
dnadoba: This is a Swift 5.8 feature: https://github.com/apple/swift-evolution/blob/main/proposals/0367-conditional-attributes.md

PR #672, opened 2023-03-16
Collect function fix

carolinacass:
Motivation:
not accumulate too many bytes

Modifications:
Implementing collect function to use NIOCore version to prevent overflowing


Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -83,6 +83,15 @@ extension HTTPClientResponse {
...
             .init(storage: self.storage.makeAsyncIterator())
         }
+
+        /// Accumulates an ``Swift/AsyncSequence`` of ``ByteBuffer``s into a single ``ByteBuffer``.
+        /// - Parameters:
+        ///   - maxBytes: The maximum number of bytes this method is allowed to accumulate
+        /// - Throws: `NIOTooManyBytesError` if the the sequence contains more than `maxBytes`.
+        /// - Returns: the number of bytes collected over time
+        func collect(maxBytes: Int) async throws -> ByteBuffer {
+            return try await self.collect(upTo: maxBytes)
```
dnadoba, line 93: If the response already announces its body size we can check that size before downloading anything. This is e.g. the case if the response contains a `Content-Length` header which contains the expected body size. If it is bigger than `maxBytes` we should throw before calling `self.collection(upTo:)`. To actually do that we need to move this extension from `HTTPClientResponse.Body` to  `HTTPClientResponse` or have some kind of reference in `HTTPClientResponse.Body` to the headers.

Be careful though, there are edge cases where we don't expect a body even if `Content-Length` is set. This is e.g. the case if the request was just a `HEAD` request. I think there a are more but that's the only one I know of the top of my head.

dnadoba (changes requested):

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -83,6 +83,15 @@ extension HTTPClientResponse {
         @inlinable public func makeAsyncIterator() -> AsyncIterator {
             .init(storage: self.storage.makeAsyncIterator())
         }
+
+        /// Accumulates an ``Swift/AsyncSequence`` of ``ByteBuffer``s into a single ``ByteBuffer``.
```
Lukasa, line 87: We can be clearer here: this accumulates `Body`.

Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -592,166 +592,30 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
...
             XCTAssertEqual(response2.headers["content-length"], [])
             guard let body = await XCTAssertNoThrowWithResult(
-                try await response2.body.collect()
+                try await response2.body.collect(maxBytes: 1024)
             ) else { return }
             XCTAssertEqual(body, ByteBuffer(string: "1234"))
         }
     }
 
-    func testRejectsInvalidCharactersInHeaderFieldNames_http1() {
```
Lukasa, line 601: Can we un-delete these tests?

Lukasa (changes requested):

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
carolinacass: I do not think this is appropriate for helper functions to check the `Content-Length`, but will confirm that it deals with it correctly.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -83,6 +96,48 @@ extension HTTPClientResponse {
...
+        /// - Parameters:
+        ///   - maxBytes: The maximum number of bytes this method is allowed to accumulate
+        /// - Throws: `NIOTooManyBytesError` if the the sequence contains more than `maxBytes`.
+        /// - Returns: the number of bytes collected over time
+        @inlinable public func collect(upTo maxBytes: Int) async throws -> ByteBuffer {
+            switch storage {
+            case .transaction(let transactionBody):
+                if let contentLength = transactionBody.expectedContentLength {
+                    print(maxBytes)
+                    print("contentLength", contentLength)
```
dnadoba, line 110:
```suggestion
```

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -83,6 +96,48 @@ extension HTTPClientResponse {
...
+
+        }
+
+    }
+}
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientResponse {
+//    need to pass in the arguments
+    static func expectedContentLength(requestMethod: HTTPMethod?, headers: HTTPHeaders, status: HTTPResponseStatus) -> Int? {
```
dnadoba, line 131:
```suggestion
    static func expectedContentLength(requestMethod: HTTPMethod, headers: HTTPHeaders, status: HTTPResponseStatus) -> Int? {
```
We make request method non-optional now

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -83,6 +96,48 @@ extension HTTPClientResponse {
...
+        if let requestMethod {
+            if status == .notModified {
+               return 0
+           } else if requestMethod == .HEAD {
+               return 0
+           }
+       }
+        let contentLengths = headers["content-length"].first.flatMap({Int($0, radix: 10)})
+        return contentLengths
```
dnadoba, line 140:
```suggestion
        let contentLength = headers["content-length"].first.flatMap({Int($0, radix: 10)})
        return contentLength
```

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -83,6 +96,48 @@ extension HTTPClientResponse {
...
+                    print(maxBytes)
+                    print("contentLength", contentLength)
+                    if contentLength > maxBytes {
+                        throw NIOTooManyBytesError()
+                    }
+                }
+            case .anyAsyncSequence:
+                break
+            }
+            func collect<Body: AsyncSequence>(_ body: Body, maxBytes: Int) async throws -> ByteBuffer where Body.Element == ByteBuffer {
```
dnadoba, line 118: Could you add a doc comment which explains that we need this helper method to call the method defined on `AsyncSequence<ByteBuffer>`?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -132,10 +187,10 @@ extension HTTPClientResponse.Body.Storage.AsyncIterator: AsyncIteratorProtocol {
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 extension HTTPClientResponse.Body {
     init(_ body: TransactionBody) {
-        self.init(.transaction(body))
+        self.init(.transaction(body), expectedContentLength: body.expectedContentLength)
     }
 
-    @usableFromInline init(_ storage: Storage) {
+    @usableFromInline init(_ storage: Storage, expectedContentLength: Int?) {
```
dnadoba, line 193:
```suggestion
    @usableFromInline init(_ storage: Storage) {
```
`expectedContentLength` is no longer used and can be removed.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -132,10 +187,10 @@ extension HTTPClientResponse.Body.Storage.AsyncIterator: AsyncIteratorProtocol {
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 extension HTTPClientResponse.Body {
     init(_ body: TransactionBody) {
-        self.init(.transaction(body))
+        self.init(.transaction(body), expectedContentLength: body.expectedContentLength)
```
dnadoba, line 190:
```suggestion
        self.init(.transaction(body))
```

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -146,7 +201,7 @@ extension HTTPClientResponse.Body {
     @inlinable public static func stream<SequenceOfBytes>(
         _ sequenceOfBytes: SequenceOfBytes
     ) -> Self where SequenceOfBytes: AsyncSequence & Sendable, SequenceOfBytes.Element == ByteBuffer {
-        self.init(.anyAsyncSequence(AnyAsyncSequence(sequenceOfBytes.singleIteratorPrecondition)))
+        self.init(.anyAsyncSequence(AnyAsyncSequence(sequenceOfBytes.singleIteratorPrecondition)), expectedContentLength: nil)
```
dnadoba, line 204:
```suggestion
        self.init(.anyAsyncSequence(AnyAsyncSequence(sequenceOfBytes.singleIteratorPrecondition)))
```

Tests/AsyncHTTPClientTests/HTTPClientResponseTests.swift
```diff
@@ -0,0 +1,65 @@
...
+        guard #available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *) else { return }
+        XCTAsyncTest {
+            var response = HTTPClientResponse(status: .notModified , requestMethod: .GET)
+            response.headers.replaceOrAdd(name: "content-length", value: "1025")
+            guard let body = await XCTAssertNoThrowWithResult(
+                try await response.body.collect(upTo: 1024)
+            ) else { return }
+            XCTAssertEqual(0, body.readableBytes)
+        }
+    }
```
dnadoba, line 64: This is no longer testing what we want. I think we can just remove it

Tests/AsyncHTTPClientTests/HTTPClientResponseTests.swift
```diff
@@ -0,0 +1,65 @@
...
+    var config = HTTPClient.Configuration()
+    config.tlsConfiguration = .clientDefault
+    config.tlsConfiguration?.certificateVerification = .none
+    config.httpVersion = .automatic
+    return HTTPClient(
+        eventLoopGroupProvider: eventLoopGroupProvider,
+        configuration: config,
+        backgroundActivityLogger: Logger(label: "HTTPClient", factory: StreamLogHandler.standardOutput(label:))
+    )
+}
```
dnadoba, line 35: After we have removed the last test case, this is no longer used

Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -745,17 +744,34 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
...
+        guard #available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *) else { return }
+        XCTAsyncTest {
+            let bin = HTTPBin(.http2(compress: false)) { _ in HTTPEchoHandler() }
+            defer { XCTAssertNoThrow(try bin.shutdown()) }
+            let client = makeDefaultHTTPClient()
+            defer { XCTAssertNoThrow(try client.syncShutdown()) }
+            let logger = Logger(label: "HTTPClient", factory: StreamLogHandler.standardOutput(label:))
+            var request = HTTPClientRequest(url: "https://localhost:\(bin.port)/")
+            request.method = .GET
+            request.body = .bytes(ByteBuffer(string: "1235"))
```
dnadoba, line 758: Nit:
```suggestion
            request.body = .bytes(ByteBuffer(string: "1234"))
```

dnadoba (changes requested):
Great, a couple of small tweaks and we should be good to go.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
dnadoba: Still missing

Tests/AsyncHTTPClientTests/HTTPClientResponseTests.swift
dnadoba: Still here

Tests/AsyncHTTPClientTests/HTTPClientResponseTests.swift
dnadoba: still here

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -32,16 +32,30 @@ public struct HTTPClientResponse: Sendable {
...
     init(
-        bag: Transaction,
-        version: HTTPVersion,
-        status: HTTPResponseStatus,
-        headers: HTTPHeaders
+        version: HTTPVersion = .http1_1,
+        status: HTTPResponseStatus = .ok,
+        headers: HTTPHeaders = [:],
+        body: Body = Body(),
+        requestMethod: HTTPMethod?
```
dnadoba, line 42: We can remove the `requestMethod` argument name again

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -32,19 +32,8 @@ public struct HTTPClientResponse: Sendable {
...
-        headers: HTTPHeaders
-    ) {
-        self.version = version
-        self.status = status
-        self.headers = headers
-        self.body = Body(TransactionBody(bag))
-    }
 
-    @inlinable public init(
+    @inlinable init(
```
dnadoba, line 36: This `init` needs to stay `public`

dnadoba (approved):

Sources/AsyncHTTPClient/AsyncAwait/TransactionBody.swift
```diff
@@ -20,9 +20,11 @@ import NIOCore
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 @usableFromInline final class TransactionBody: Sendable {
     @usableFromInline let transaction: Transaction
+    @usableFromInline let expectedContentLength: Int?
 
-    init(_ transaction: Transaction) {
+    init(_ transaction: Transaction, expextedContentLength: Int?) {
```
Lukasa, line 25: Typo here.

Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -745,17 +744,32 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
...
+        guard #available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *) else { return }
+        XCTAsyncTest {
+            let bin = HTTPBin(.http2(compress: false)) { _ in HTTPEchoHandler() }
+            defer { XCTAssertNoThrow(try bin.shutdown()) }
+            let client = makeDefaultHTTPClient()
+            defer { XCTAssertNoThrow(try client.syncShutdown()) }
+            let logger = Logger(label: "HTTPClient", factory: StreamLogHandler.standardOutput(label:))
+            var request = HTTPClientRequest(url: "https://localhost:\(bin.port)/")
+            request.method = .GET
+            request.body = .bytes(ByteBuffer(string: "1234"))
```
Lukasa, line 758: Hrm, I think this muddies the waters of the test somewhat. I don't think this test would fail if you took away the content-length checks, because the actual `collect` call will still end up violating the content-length.

What you want is for the HTTPbin server to send you back a content-length that's too long and no body at all. That should still throw this error.

Lukasa (changes requested):

Lukasa:

Lukasa (approved):
Nice one, this LGTM.

Lukasa:
Ah, sorry, can you run `generate_linux_tests.rb` please?

Lukasa (approved):

PR #675, opened 2023-03-29
Allow DNS override

dnadoba:
Sometimes it can be useful to connect to one host e.g. `x.example.com` but request and validate the certificate chain as if we would connect to `y.example.com`. This is what this PR adds support for by adding a `dnsOverride` configuration to `HTTPClient.Configuration`. This is similar to curls `—resolve` option but only allows overriding host and not ports for now.


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -434,7 +434,7 @@ extension HTTPConnectionPool.ConnectionFactory {
         }
         #endif
 
-        let sslServerHostname = self.key.connectionTarget.sslServerHostname
+        let sslServerHostname = self.key.serverNameIndicatorOverride ?? self.key.connectionTarget.sslServerHostname
```
Lukasa, line 437: Should we hide this in a computed property on the key so as to centralise the knowledge of how to do this?

Lukasa (commented):
Mostly looks really good! One small note.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
dnadoba: Done. Actually found through this another place where I forgot to use the override, namely if we proxy:
https://github.com/swift-server/async-http-client/pull/675/commits/f296aa4cd01a074ff580791dcf7f6717f3c6a93a

Lukasa (approved):

PR #677, opened 2023-04-05
Replace `TransactionBody` with `NIOAsyncSequenceProducer`

dnadoba:
### Motivation
We want to properly respect task cancelation during the call to `HTTPClientResponse.Body.AsyncIterator.next()`. 

### Modification
Replace our custom `AsyncSequence` implementation by delegate most of the work to `NIOAsyncSequenceProducer` from `NIOCore`

### Result
We fully support cancelation throughout the stack and get the performance benefits of `NIOAsyncSequenceProducer`.


dnadoba:
We now also throw `CancellationError` instead of `HTTPClientError.cancelled` if the task is cancelled while awaiting the response head.

glbrntt (approved):
I think this looks good but I'm not super familiar with AHC.

Lukasa (approved):

PR #678, opened 2023-04-11
Add a control to limit connection reuses

glbrntt:
Motivation:

Sometimes it can be helpful to limit the number of times a connection
can be used before discarding it. AHC has no such support for this at
the moment.

Modifications:

- Add a `maximumUsesPerConnection` configuration option which defaults
  to `nil` (i.e. no limit).
- For HTTP1 we count down uses in the state machine and close the
  connection if it hits zero.
- For HTTP2, each use maps to a stream so we count down remaining uses
  in the state machine which we combine with max concurrent streams to
  limit how many streams are available per connection. We also count
  remaining uses in the HTTP2 idle handler: we treat no remaining uses
  as receiving a GOAWAY frame and notify the pool which then drains the
  streams and replaces the connection.

Result:

Users can control how many times each connection can be used.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
@@ -192,6 +192,7 @@ extension HTTPConnectionPool {
             case .active(let conn, let maxStreams, var usedStreams, let lastIdle, let remainingUses):
                 usedStreams += count
                 precondition(usedStreams <= maxStreams, "tried to lease a connection which is not available")
+                precondition(remainingUses.map { $0 >= count } ?? true, "tried to lease streams from a connection which does not have enough remaining streams")
                 self.state = .active(conn, maxStreams: maxStreams, usedStreams: usedStreams, lastIdle: lastIdle, remainingUses: remainingUses.map { $0 - count })
                 return conn
             }
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
@@ -18,12 +18,12 @@ extension HTTPConnectionPool {
     private struct HTTP2ConnectionState {
         private enum State {
             /// the pool is establishing a connection. Valid transitions are to: .backingOff, .active and .closed
-            case starting
+            case starting(maximumUses: Int?)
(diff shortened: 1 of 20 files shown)


Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
```diff
@@ -176,10 +189,10 @@ extension HTTPConnectionPool {
             case .starting, .backingOff, .draining, .closed:
                 preconditionFailure("Invalid state for leasing a stream: \(self.state)")
 
-            case .active(let conn, let maxStreams, var usedStreams, let lastIdle):
+            case .active(let conn, let maxStreams, var usedStreams, let lastIdle, let remainingUses):
                 usedStreams += count
                 precondition(usedStreams <= maxStreams, "tried to lease a connection which is not available")
-                self.state = .active(conn, maxStreams: maxStreams, usedStreams: usedStreams, lastIdle: lastIdle)
+                self.state = .active(conn, maxStreams: maxStreams, usedStreams: usedStreams, lastIdle: lastIdle, remainingUses: remainingUses.map { $0 - count })
```
dnadoba, line 195: can we add a precondition for `remainingUses >= count`

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -758,6 +758,18 @@ public class HTTPClient {
...
         public var networkFrameworkWaitForConnectivity: Bool
 
+        /// The maximum number of times each connection can be used before it is replaced with a new one. Use `nil` (the default)
+        /// if no limit should be applied to each connection.
+        ///
+        /// - Precondition: The value must be greater than zero.
+        public var maximumUsesPerConnection: Int? {
+            willSet {
+                if let newValue = newValue {
+                    precondition(newValue > 0, "maximumUsesPerConnection must be greater than zero or nil")
```
dnadoba, line 768: Lets make this a fatalError so this prints the reason in release builds as well.

dnadoba (approved):
LGTM minus two nits

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift
glbrntt: Fixed in d5134bb0a46b12a2d70cb161bed5cc406e9112fc

Sources/AsyncHTTPClient/HTTPClient.swift
glbrntt: Fixed in d5134bb0a46b12a2d70cb161bed5cc406e9112fc

PR #681, opened 2023-04-12
Pass request `Task` to `FileDownloadDelegate` `reportHead` and `reportProgress` closures

dnadoba:
### Motivation
If the HTTP response status is not 2xx or the file being downloaded becomes too big, it is desirable to cancel the file download eagerly. This is currently quite hard because the `reportHead` and `reportProgress` closures do not have direct access to the `HTTPClient.Task`.

### Modifications
- Pass `HTTPClient.Task` additionally to `reportHead` and `reportProgress` closures

### Result
A file download can now easily be cancelled at any time.


MaxDesiatov (approved):
This looks like a quite more elegant solution. Much appreciated!

Lukasa (approved):

PR #683, opened 2023-04-13
Add support for custom cancellation error

dnadoba:
### Motivation
A users might want to provide more information why a request got cancelled. For example if we cancel a file download because the response status code is not 2xx, it is desirable to propagate which status code was returned.

### Modification
- add `Task.fail(_:)` in addition to `Task.cancel()` that takes a custom Error which is used to complete the underlying request promise.

### Result
Improved error messages.


MaxDesiatov (approved):
Splendid, thank you!

Lukasa (approved):
Really nice change, I like this a lot.

Issue #684, opened 2023-04-14
'Transaction' is only available in iOS 13.0 or newer


chocoford:
In File [Transaction.swift](https://github.com/swift-server/async-http-client/blob/5b4f03de0600da906f9e46e9f636ec26218da080/Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift#L328).

The extension of `Transaction` only marked as `@available(macOS 10.15, *)`

```swift
@available(macOS 10.15, *) // <-- here
extension Transaction: NIOAsyncSequenceProducerDelegate {
    @usableFromInline
    func produceMore() {
...
```

PR #685, opened 2023-04-14
Fix building transaction extension for Apple Platforms

cbaker6:
The repo currently won't build for other OS's because the extension doesn't match: https://github.com/swift-server/async-http-client/blob/2588b8cdd4be62f72da849f42ec1d1b25c985937/Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift#L21-L22


dnadoba:

dnadoba:
Good catch and thanks a lot for opening a PR to fix it! 
resolves https://github.com/swift-server/async-http-client/issues/684

dnadoba (approved):

PR #688, opened 2023-05-15
Tolerate new request after connection error happened

dnadoba:
### Motivation
A connection and the connection pool may be on different threads and therefore run concurrently. If a connection error happens the connections closes itself and notifies the pool about it. While this is happening the connection pool could already decide to run a new connection on the closing connection. We need to tolerate this happening. We already partially do but only if the connection is fully closed but not currently still closing.

### Result
- change all `preconditionFailure`s in `HTTP1ConnectionStateMachine.swift` to `fatalError`s to see the message in release builds
- tolerate `.closing` in addition to `.closed` in `runNewRequest`
- add test which previously crashed

### Result
AHC no longer crashes in the race condition described above.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ConnectionStateMachine.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ConnectionStateMachine.swift
@@ -106,14 +106,14 @@ struct HTTP1ConnectionStateMachine {
             return .wait
 
         case .modifying:
-            preconditionFailure("Invalid state: \(self.state)")
+            fatalError("Invalid state: \(self.state)")
         }
     }
 
     mutating func channelInactive() -> Action {
         switch self.state {
         case .initialized:
-            preconditionFailure("A channel that isn't active, must not become inactive")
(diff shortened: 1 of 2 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ConnectionStateMachine.swift
```diff
@@ -182,15 +182,15 @@ struct HTTP1ConnectionStateMachine {
...
             // connection pool state machine must not send a new request to the connection, if the
             // connection is `.initialized`, `.closing` or `.inRequest`
-            preconditionFailure("Invalid state: \(self.state)")
+            fatalError("Invalid state: \(self.state)")
 
-        case .closed:
+        case .closing, .closed:
             // The remote may have closed the connection and the connection pool state machine
             // was not updated yet because of a race condition. New request vs. marking connection
             // as closed.
```
dnadoba, line 196: This is the actual bug fix.

Lukasa (approved):
LGTM

fabianfett (approved):
LGTM. Great change!

PR #698, opened 2023-07-13
Fix flaky `AsyncAwaitEndToEndTests.testImmediateDeadline` test

dnadoba:
We have observed that `AsyncAwaitEndToEndTests.testImmediateDeadline` can also return a `HTTPClientError.remoteConnectionClosed` in rare circumstances. If closing  happens really fast we might shutdown the pipeline before we fail the request. If the pipeline is closed we may receive a  `.remoteConnectionClosed` which can execute before we push the previous error to the request state machine. We can't change this in the current architecture so we will just tolerate this error in tests as well. 

The `AsyncAwaitEndToEndTests.testDeadline()` is less likely to experience this issue but I have changed it there too just to be safe.

### DIFF
--- a/Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
+++ b/Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
@@ -389,8 +389,10 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
                 guard let error = error as? HTTPClientError else {
                     return XCTFail("unexpected error \(error)")
                 }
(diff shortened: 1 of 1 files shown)


dnadoba:

Test seems to have timed out:
```
[502/503] Testing AsyncHTTPClientTests.RequestBagTests/testCancelFailsTaskAfterRequestIsSent
```
Have added a timeout to the `ELF.wait()` in #699 to better debug this in the future. CI might also just got throttled as this is the second last test.

rnro (approved):

Issue #392, opened 2021-07-07
RFC: design suggestion: Make this a "3-tier library"


weissi:
At the moment, we have a pretty gnarly issue with AHC's API: Settings can either be supplied globally or per request. So if you want to set certain configuration for a certain component in your app (say TLS config, redirect config, ...) then you either have to specify it for every single request (tedious) or globally (expensive as you'll get a new connection pool, need to maintain lifecycle etc).

What I suggest here is making AHC a "3-tier library":

- tier 1: What is currently `HTTPClient`: the highly stateful, expensive, and lifecycle-managed object that owns the `TLSContext`, the connection pool, and if not `.shared` also the ELG
- tier 2: (doesn't exist right now) A very lightweight (probably `struct`) and not lifecycle managed client object which is essentially a reference to the tier 1 object and a bag of configuration
- tier 3: the individual requests

An API sketch could look like this:

```
// tier 1
let http = HTTPService(eventLoopGroupProvider: .shared(myELG)) // This is currently called `HTTPClient`
defer {
    try! http.syncShutdown()
}

// tier 2 (creating this is ~free, doesn't need to be shut down)
let httpClient = http.client(tlsConfiguration: specialConfig, redirectConfig: ...)

// tier 3
let result = httpClient.get("https://...").wait()
```

A few things are important:

#### Tier 1

- do not take any configuration apart from the ELG

#### Tier 2

- doesn't need lifecycle management
- holds bags of configuration
- ~free to create

#### Tier 3

- can override most(all?) of the configuration set in Tier 2

fabianfett:
I agree with most of the things you mention. And I'm +1 on the general proposal. However I don't get why we should be able to override most settings in tier3. If tier2 is basically free to create, I don't see any benefit by overriding the settings. In my point of view the client should be a struct with value semantics.

weissi:
@fabianfett good point! I don't know either :D

Lukasa:
Different TLS configurations interact somewhat poorly with the connection pool, as they produce new keys. So this isn't entirely a panacea for the connection pool re-use. Otherwise agreed.

weissi:
> Different TLS configurations interact somewhat poorly with the connection pool, as they produce new keys. So this isn't entirely a panacea for the connection pool re-use. Otherwise agreed.

Agreed, but we already have that issue/feature today. And we added it because people very commonly requested it, usually because the need a different trust roots/client certs to talk to different services.

fabianfett:
[...]

PR #709, opened 2023-08-15
HTTPClientRequest: allow custom TLS config

weissi:
In the transition from `HTTPClient.Request` to `HTTPClientRequest` we unfortunately lost some important functionality: Custom TLS configuration.

The can be important to override on a per-request basis but more importantly it was the most significant step which allows us to unlock the 3-tier architecture (#392). For the 3-tier architecture it's very important to get rid of any top-level configuration.

This PR brings back the lost functionality.

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
@@ -15,6 +15,7 @@
 import Algorithms
 import NIOCore
 import NIOHTTP1
+import NIOSSL
 
 @usableFromInline
 let bagOfBytesToByteBufferConversionChunkSize = 1024 * 1024 * 4
@@ -32,6 +33,9 @@ let byteBufferMaxSize = Int(UInt32.max)
 /// A representation of an HTTP request for the Swift Concurrency HTTPClient API.
 ///
 /// This object is similar to ``HTTPClient/Request``, but used for the Swift Concurrency API.
+///
+/// - note: For many ``HTTPClientRequest/body-swift.property`` configurations, this type is _not_ a value type
+///    (https://github.com/swift-server/async-http-client/issues/708).
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 public struct HTTPClientRequest: Sendable {
     /// The request URL, including scheme, hostname, and optionally port.
@@ -46,11 +50,15 @@ public struct HTTPClientRequest: Sendable {
     /// The request body, if any.
     public var body: Body?
 
+    /// Request-specific TLS configuration, defaults to no request-specific TLS configuration.
+    public var tlsConfiguration: TLSConfiguration?
+
     public init(url: String) {
         self.url = url
         self.method = .GET
         self.headers = .init()
         self.body = .none
+        self.tlsConfiguration = nil
     }
 }
 
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -3525,4 +3525,49 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
         let response = try client.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
         XCTAssertEqual(.ok, response.status)
     }
+
+    func testAsyncExecuteWithCustomTLS() async throws {
+        let httpsBin = HTTPBin(.http1_1(ssl: true))
+        defer {
+            XCTAssertNoThrow(try httpsBin.shutdown())
+        }
+
+        // A client with default TLS settings, i.e. it won't accept `httpsBin`'s fake self-signed cert
+        let client = HTTPClient(eventLoopGroup: MultiThreadedEventLoopGroup.singleton)
+        defer {
+            XCTAssertNoThrow(try client.shutdown().wait())
+        }
+
+        var request = HTTPClientRequest(url: "https://localhost:\(httpsBin.port)/get")
+
+        // For now, let's allow bad TLS certs
+        request.tlsConfiguration = TLSConfiguration.clientDefault
+        // ! is safe, assigned above
+        request.tlsConfiguration!.certificateVerification = .none
+
+        let response1 = try await client.execute(request, timeout: /* infinity */ .hours(99))
+        XCTAssertEqual(.ok, response1.status)
+
+        // For the second request, we reset the TLS config
+        request.tlsConfiguration = nil
+        do {
+            let response2 = try await client.execute(request, timeout: /* infinity */ .hours(99))
+            XCTFail("shouldn't succeed, self-signed cert: \(response2)")
+        } catch {
+            switch error as? NIOSSLError {
+            case .some(.handshakeFailed(_)):
+                () // ok
+            default:
+                XCTFail("unexpected error: \(error)")
+            }
+        }
+
+        // And finally we allow it again.
+        request.tlsConfiguration = TLSConfiguration.clientDefault
+        // ! is safe, assigned above
+        request.tlsConfiguration!.certificateVerification = .none
+
+        let response3 = try await client.execute(request, timeout: /* infinity */ .hours(99))
+        XCTAssertEqual(.ok, response3.status)
+    }
 }
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
@@ -78,7 +78,7 @@ extension HTTPClient {
         // this loop is there to follow potential redirects
         while true {
             let preparedRequest = try HTTPClientRequest.Prepared(currentRequest, dnsOverride: configuration.dnsOverride)
-            let response = try await executeCancellable(preparedRequest, deadline: deadline, logger: logger)
+            let response = try await self.executeCancellable(preparedRequest, deadline: deadline, logger: logger)
 
             guard var redirectState = currentRedirectState else {
                 // a `nil` redirectState means we should not follow redirects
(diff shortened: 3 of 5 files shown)


fabianfett (approved):
Love it! I should have done this way earlier! I especially like that we add this property to the request itself and don't pass it as a function parameter!

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -14,10 +14,14 @@
 
 import NIOCore
 import NIOHTTP1
+import NIOSSL
```
fabianfett, line 17: CI tells us that TLSConfiguration isn't marked as Sendable in Swift 5.6.

```suggestion
#if swift(<5.7)
@preconcurrency import NIOSSL
#else
import NIOSSL
#endif
```

Lukasa: That's not the right fix, it'll warn. The right fix is to put it in an `UnsafeTransferBox` in 5.6. The reason this is this way is that on 5.6 we couldn't mark the closures in the TLSConfiguration object as `@preconcurrency Sendable`.

weissi: @Lukasa it being `var tlsConfiguration: TLSConfig?` --> `var tlsConfiguration: UnsafeTransfer<TLSConfig>?` for 5.6 only?

weissi: @Lukasa actually I think `@preconcurrency` will work warning-free. That's because the `@preconcurrency` isn't actually unused. But let's see

dnadoba:
As discussed in https://github.com/swift-server/async-http-client/issues/392, there seems to be no point in tier 3 if tier 2 is ~free to create.

weissi:
> As discussed in #392, there seems to be no point in tier 3 if tier 2 is ~free to create.

Convenience and you don't need to repeat everything.

For example, I might want to ~always set a certain header in my tier 2 but in tier 3 I definitely want to add headers.

Like

```
actor FooClient {
    private let client: HTTPClient = {
        var client = HTTPClient(...)
        client.headers["user-agent"] = "my super lib/123"
        [...]
        return client
    }

    func upgradeFoo() {
        var request = HTTPClientRequest()
        request.url = "..."
        request.headers["foo-level"] = "max"
        try await self.client.execute(request)
    }
}
```

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
weissi: @Lukasa yes worked, all green.

dnadoba:
I agree with that, default headers on tier 2 are great. What I don't agree with is that tier 3 has override for everything in tier 2. Tier 3 should just be `(HTTPRequest, Body)` where `HTTPRequest` is the new type form [`swift-http-types`](https://github.com/apple/swift-http-types) and `Body` is some kind of `AsyncSequence`. Nothing else.

weissi:
> I agree with that, default headers on tier 2 are great. What I don't agree with is that tier 3 has override for everything in tier 2. Tier 3 should just be `(HTTPRequest, Body)` where `HTTPRequest` is the new type form [`swift-http-types`](https://github.com/apple/swift-http-types) and `Body` is some kind of `AsyncSequence`. Nothing else.

I think that's important. But also irrelevant for this PR. This PR brings back accidentally lost functionality when we added `HTTPClientRequest` a close friend of `HTTPClient.Request` (even documented that way). `HTTPClient.Request` has `tlsConfiguration` and so should `HTTPClientRequest`. 

In the future when we pick up swift-http-types the whole nomenclature of what a "request" even is will change. Then we can discuss where the various configuration options can go. But this is the wrong place IMHO, we're just adding stuff back that we forgot before.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -14,10 +14,18 @@
 
 import NIOCore
 import NIOHTTP1
+#if compiler(>=5.7)
```
dnadoba, line 17: We have just drop 5.6, we can remove this now.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3525,4 +3525,47 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+            case .some(.handshakeFailed(_)):
+                () // ok
+            default:
+                XCTFail("unexpected error: \(error)")
+            }
+        }
+
+        // And finally we allow it again.
+        request.tlsConfiguration = TLSConfiguration.clientDefault
+        request.tlsConfiguration!.certificateVerification = .none
```
dnadoba, line 3566: lets get rid of the force unwrap.
```suggestion
        // And finally we allow it again.
        let tlsConfiguration = TLSConfiguration.clientDefault
        tlsConfiguration.certificateVerification = .none
        request.tlsConfiguration = tlsConfiguration
```

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3525,4 +3525,47 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+        let client = HTTPClient(eventLoopGroup: MultiThreadedEventLoopGroup.singleton)
+        defer {
+            XCTAssertNoThrow(try client.shutdown().wait())
+        }
+
+        var request = HTTPClientRequest(url: "https://localhost:\(httpsBin.port)/get")
+
+        // For now, let's allow bad TLS certs
+        request.tlsConfiguration = TLSConfiguration.clientDefault
+        request.tlsConfiguration!.certificateVerification = .none
```
dnadoba, line 3545: lets get ride of the force unwrap here as well.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3525,4 +3525,47 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+
+        let response1 = try await client.execute(request, timeout: .hours(99))
+        XCTAssertEqual(.ok, response1.status)
+
+        // For the second request, we reset the TLS config
+        request.tlsConfiguration = nil
+        do {
+            let response2 = try await client.execute(request, timeout: .hours(99))
```
dnadoba, line 3553: that's a bit long

dnadoba (commented):
After thinking about this again I came to the conclusion that it doesn't hurt to add this before we implement the 3-tier architecture as we will likely deprecate `HTTPClientRequest` anyway. 

Minor nits, afterwards we are good to go from my side.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: but that would hide issues. Force unwraps here are good because they'd catch that going wrong at the right place here

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
weissi: cool, will remove

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: I mean I can do that but what's bad about force unwraps used correctly? This one won't trigger

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: I wanted to use `.hours(.max)` to make this clear but https://github.com/apple/swift-nio/issues/2532

weissi:
@dnadoba can we get this over the line please?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
dnadoba: No, `nil` is the default configuration and it would be reasonable to normalise the input value to become nil after setting it to the default value.

Note that this shouldn't be an Optional property to begin with but that's the current API.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: I'm assigning it literally the line above. If this fails then AHC has a severe bug and I'd like to know about it rather than hide it.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
weissi: No, that would not be reasonable, that would be user hostile. If this breaks, AHC is broken, let's not hide this

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
dnadoba: I'm not against using force unwraps in general. However, you should only use them if you can reason about it locally. `tlsConfigurataion` isn't a local variable but a property of a type that is define in another module. Keep in mind that properties in Swift can be computed so it can be arbitrary code that is executed on the getter and setter. If you want to know that the force unwrap is safe, you and everyone that looks at that tests in the future, needs to look at the implementation.

In theory you can also add a comment to the force unwrap explaining why it is safe but the code that circumvents it is straight forward and also just a single line.

weissi: @dnadoba Apologies but this is backwards. 

If a user writes

```
        request.tlsConfiguration = TLSConfiguration.clientDefault
        request.tlsConfiguration!.certificateVerification = .none
```

then this has to work. Yes, it is _possible_ to break this code if you were so inclined but then the `HTTPClientRequest` type would just be outright broken.

Remember, it's not the goal to be user hostile. If this type were defined in another package that was known as having user hostile types then your comment might be appropriate. But this is a type from this package and it shouldn't be user hostile.

weissi: added the comment. The guidelines aren't meant for test code but for production code.

dnadoba:

dnadoba (approved):

PR #710, opened 2023-08-30
Automatically chunk large request bodies

dnadoba:
### Motivation
If the `HTTPClientRequest.Body` is created from a type that is not `ByteBuffer`, we need to convert it to it at some point. 
`ByteBuffer` can only hold up to `Int32.max` on 32-bit platforms and `UInt32.max` on 64-bit platforms and crashes if it exceeds this size limit. 

### Modification
- Convert from `some Sequence<UInt8>`, some Collection<UInt8>`, some RandomAccessCollection<UInt8>` and `String` in lazy chunks if the input exceeds 4 MB. 
- `some Sequence<UInt8>` is special as it can't be efficiently chunked lazily. Therefore we need to do the chunking eagerly if it implements the fast path `withContiguousStorageIfAvailable`. Otherwise we do the chunking lazy but "slowly".
- conform `HTTPClientRequest.Body` to `AsyncSequence` to make testing easier.

### Result
Converting from `some Sequence<UInt8>`, some Collection<UInt8>`, some RandomAccessCollection<UInt8>` and `String` to `ByteBuffer` will no longer crash for inputs that are larger than 4GB. Conversion above 4MB is now more memory efficient as we do it lazily and not eagerly all at once.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,6 +29,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.4"),
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.2"),
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-algorithms", from: "1.0.0"),
     ],
     targets: [
         .target(name: "CAsyncHTTPClient"),
@@ -48,6 +49,7 @@ let package = Package(
                 .product(name: "NIOTransportServices", package: "swift-nio-transport-services"),
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Atomics", package: "swift-atomics"),
+                .product(name: "Algorithms", package: "swift-algorithms"),
             ]
         ),
         .testTarget(
@@ -64,6 +66,7 @@ let package = Package(
                 .product(name: "NIOSOCKS", package: "swift-nio-extras"),
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Atomics", package: "swift-atomics"),
+                .product(name: "Algorithms", package: "swift-algorithms"),
             ],
             resources: [
                 .copy("Resources/self_signed_cert.pem"),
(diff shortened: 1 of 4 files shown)


Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -12,9 +12,24 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Algorithms
 import NIOCore
 import NIOHTTP1
 
+// TODO: should this be a multiple of a page size?
```
Lukasa, line 19: On all platforms it will be.

Package.swift
```diff
@@ -29,6 +29,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.4"),
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.2"),
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-algorithms", from: "1.0.0"),
```
Lukasa, line 32: I'm a bit on the fence about this dependency. Is it worth us considering not using it?

dnadoba: We could also [copy & paste the implementation](https://github.com/apple/swift-algorithms/blob/fc8fdfd4dcc6d05f9d603c581fcbba643cb676df/Sources/Algorithms/Chunked.swift#L518-L824). Hasn't changed in ~2 years (except documentation).

Lukasa: @FranzBusch thoughts on adding this dep?

FranzBusch: TBH, I think we should treat `swift-algorithms` similar to how we treat `swift-collections` or `swift-atomics`. They are stdlib extensions and follow the best practices around SemVer. IMO totally fine to pull this one in. In the end, any application will probably pull in `swift-collections`, `swift-atomics`, `swift-algorithms`, `swift-log`, `swift-metrics`, `swift-tracing` and in the future probably `swift-async-algorithms` and `swift-service-lifecycle`.

dnadoba:
@Lukasa anything else?

Lukasa (approved):

PR #714, opened 2023-10-25
Fix wrong/outdated `connect` timeout documentation

MahdiBM:
The documentation states `30 seconds` but other documentations and the implementation all agree on `10 seconds` timeout as `HTTPClient.Configuration.connect`.


dnadoba (approved):
Good catch! Thank you!

Lukasa:

PR #716, opened 2023-11-02
testPlatformConnectErrorIsForwardedOnTimeout port reuse

PeterAdams-A:
Motivation:

The above test has been seen to fail with port already in use. The test assumes that a port can be bound to twice in a row. It's possible that another process takes the port between the two binds - this can be stopped by binding a second time before stopping the first.

Modifications:

Allow port reuse and reorder the test to bind a second time before releasing the first.

Result:

Test should no longer be flaky.


dnadoba (approved):
Thanks for looking into this!

fabianfett (approved):

simonjbeaumont:

dnadoba:

PR #718, opened 2023-12-14
Add an idle write timeout

gjcairo:
## Motivation
We currently provide an idle read timeout to fail requests that take more than a given amount of time to produce any response data. We were missing its counterpart: to provide a timeout for when no request data is being written either. 
This could happen, for example, when uploading large files: if for whatever reason (e.g. back pressure from the server, etc) the client does not write any bytes, and there is no data being sent from the server either, the connection could be idle and unused, but remain open for longer than necessary. With an idle write timeout, we can avoid this.

## Modifications
Added a new idle write timeout.

## Result
Users can now configure an idle write timeout in addition to an idle read timeout.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
@@ -53,7 +53,7 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
 
     private var idleReadTimeoutStateMachine: IdleReadStateMachine?
     private var idleReadTimeoutTimer: Scheduled<Void>?
-    
+
     private var idleWriteTimeoutStateMachine: IdleWriteStateMachine?
     private var idleWriteTimeoutTimer: Scheduled<Void>?
 
@@ -91,7 +91,7 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
         if let timeoutAction = self.idleWriteTimeoutStateMachine?.channelWritabilityChanged(context: context) {
             self.runTimeoutAction(timeoutAction, context: context)
         }
-        
+
         let action = self.state.writabilityChanged(writable: context.channel.isWritable)
         self.run(action, context: context)
     }
@@ -124,7 +124,7 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
         // The `HTTPRequestStateMachine` ensures that a `HTTP2ClientRequestHandler` only handles
         // a single request.
         self.request = request
-        
+
         if let timeoutAction = self.idleWriteTimeoutStateMachine?.write() {
             self.runTimeoutAction(timeoutAction, context: context)
         }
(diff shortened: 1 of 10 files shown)


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -938,7 +940,12 @@ extension HTTPClient.Configuration {
         /// - parameters:
         ///     - connect: `connect` timeout. Will default to 10 seconds, if no value is provided.
         ///     - read: `read` timeout.
-        public init(connect: TimeAmount? = nil, read: TimeAmount? = nil) {
+        ///     - write: `write` timeout.
+        public init(
+            connect: TimeAmount? = nil,
+            read: TimeAmount? = nil,
+            write: TimeAmount? = nil
+        ) {
```
dnadoba, line 948: This is a breaking change:
```swift
15:21:52 1 breaking change detected in AsyncHTTPClient:
15:21:52   💔 API breakage: constructor HTTPClient.Configuration.Timeout.init(connect:read:) has been removed
```
Adding a new parameter, even with a default, is breaking because users can write:
```
// type is inferred as `(Timeout?, Timeout?) -> Timeout`
let createTimeout = HTTPClient.Configuration.Timeout.init
let timeout = createTimeout(nil, nil)
```
with you change the inferred type is now `(Timeout?, Timeout?, Timeout?) -> Timeout` which is breaking.

The workaround is to leave the old init as is and default to `nil` for `write`. Add a new init where the `write` parameter is required.

Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
```diff
@@ -295,6 +322,36 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
         }
     }
 
+    private func runTimeoutAction(_ action: IdleWriteStateMachine.Action, context: ChannelHandlerContext) {
+        switch action {
+        case .startIdleWriteTimeoutTimer(let timeAmount):
+            assert(self.idleWriteTimeoutTimer == nil, "Expected there is no timeout timer so far.")
+
+            self.idleWriteTimeoutTimer = self.eventLoop.scheduleTask(in: timeAmount) {
+                guard self.idleWriteTimeoutTimer != nil else { return }
```
dnadoba, line 331: In HTTP1 we use a timer "ID" but here we are just setting the timer to nil. Is there a reason for these two different approaches?

gjcairo: Yes, this is an inconsistency and should be changed because race conditions can happen with this current approach. It was actually fixed for HTTP1 in https://github.com/swift-server/async-http-client/pull/455, but not in HTTP2. I just decided to keep things as they currently are in this PR (without using the timer ID) to avoid mixing multiple changes in a single PR, but I'll send a follow-up that adds a timer and changes these checks.

dnadoba (approved):

PR #720, opened 2023-12-18
Fix potential race conditions when cancelling read/write idle timers

gjcairo:
## Motivation
This PR is a follow-up to https://github.com/swift-server/async-http-client/pull/718. It applies the same modifications done in https://github.com/swift-server/async-http-client/pull/455 to the HTTP2 request handler to avoid a race condition when cancelling idle timers.

## Modifications
A new timer ID for each timer (read and write idle timeouts) has been added to keep track of the current idle timers, just like it's done in the HTTP1 implementation.

## Result
No more race conditions when cancelling timers in HTTP2.


dnadoba (approved):
Perfect! Thanks for taking care of this

PR #727, opened 2024-01-23
Update async usages in README and examples to shutdown properly

michaelmcguire:
Motivation:

In several examples of async/await support, the call to shutdown() on the HTTPClient are made with await but do not call get() on the EventLoopFuture.

Modifications:

Add the .get() call in four locations that need it.

Result:

After fixing this, the client should properly get closed down and there are no longer warnings emitted by the compiler.

### DIFF
--- a/Examples/GetHTML/GetHTML.swift
+++ b/Examples/GetHTML/GetHTML.swift
@@ -29,6 +29,6 @@ struct GetHTML {
             print("request failed:", error)
         }
         // it is important to shutdown the httpClient after all requests are done, even if one failed
-        try await httpClient.shutdown()
+        try await httpClient.shutdown().get()
     }
 }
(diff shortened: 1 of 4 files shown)


FranzBusch (approved):

dnadoba:

dnadoba (changes requested):
There is a [`shutdown()`  that is async](https://github.com/swift-server/async-http-client/blob/09b7eb751e297396f44b270b26754f9c85b105d1/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient%2Bshutdown.swift#L20) and needs to be awaited. I'm not sure why you are seeing the warnings. Can you please post the exact warning you are seeing? Do you have a reproducer?

michaelmcguire:
Sorry for the long delay on this. I will double-check and open again if I hit this.

PR #736, opened 2024-04-08
Two handy extensions to work with HTTPClientResponse body

makoni:
Two extensions for `HTTPClientResponse` to avoid boilerplate code.

Reading a response data currently:
```swift
let expectedBytes = headers.first(name: "content-length").flatMap(Int.init) ?? 1024 * 1024
var bytes = try await body.collect(upTo: expectedBytes)
if let data = bytes.readData(length: bytes.readableBytes) {
    // handle data
}
```

With the extension:
```swift
if let data = try await response.data(upTo: 1024*1024) {
    // handle data
}
```

Additionally an extension for `bytes`. Currently:
```swift
let expectedBytes = headers.first(name: "content-length").flatMap(Int.init) ?? 1024 * 1024
var bytes = try await body.collect(upTo: expectedBytes)
```

With the extension:
```swift
var bytes = try await response.bytes(upTo: 1024*1024)
```

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse+data.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse+data.swift
@@ -13,7 +13,6 @@
 //===----------------------------------------------------------------------===//
 
 import NIOCore
-import NIOHTTP1
 #if canImport(FoundationEssentials)
 import FoundationEssentials
 #else
@@ -27,8 +26,7 @@ extension HTTPClientResponse {
     /// - Parameter maxBytes: The maximum number of bytes this method is allowed to accumulate.
     /// - Returns: Bytes collected over time
     public func bytes(upTo maxBytes: Int) async throws -> ByteBuffer {
-        let expectedBytes = self.headers.first(name: "content-length").flatMap(Int.init) ?? maxBytes
-        return try await self.body.collect(upTo: min(expectedBytes, maxBytes))
+        return try await self.body.collect(upTo: maxBytes)
     }
 }
 


Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -138,6 +138,19 @@ extension HTTPClientResponse {
...
 
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientResponse {
+    /// Response body as `ByteBuffer`.
+    public var bytes: ByteBuffer {
+        get async throws {
+            let expectedBytes = headers
+                .first(name: "content-length")
+                .flatMap(Int.init) ?? 1024 * 1024
+            return try await body.collect(upTo: expectedBytes)
```
Lukasa, line 149: I like this helper, but I think it might be a bit better as a function. In particular I'm very reluctant to let users avoid providing the default maximum body size: just hard-coding this to 1MB is probably going to cause some confusion.

Sources/AsyncHTTPClient/FoundationExtensions.swift
```diff
@@ -64,3 +64,14 @@ extension HTTPClient.Body {
         return self.bytes(data)
     }
 }
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientResponse {
+    /// Response body as `Data`.
+    public var data: Data? {
+        get async throws {
+            var bytes = try await bytes
```
Lukasa, line 73: Mild nit: our style is to use `self.` wherever relevant.

Lukasa (commented):
Generally these look really good! Can you add a couple of tests for them as well?

Sources/AsyncHTTPClient/FoundationExtensions.swift
makoni: Done

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
makoni: Added 2 new tests

fabianfett: I'm not sure we should automatically depend on "content-length" here. A malicious client may send just peta bytes of data, which can bring down the client.

makoni: If you're getting an image from some URL you probably don't know is its size 1mb or 20mb. If you'll pass too small `upTo` value there will be a crash.

Here's an example that gives `Swift/ErrorType.swift:200: Fatal error: Error raised at top level: NIOCore.NIOTooManyBytesError()` with 1.21.0:
```swift
let request = HTTPClientRequest(url: "https://github.githubassets.com/images/modules/logos_page/GitHub-Mark.png")
let response = try await HTTPClient.shared.execute(request, timeout: .seconds(30))
let buffer = try await response.body.collect(upTo: 10)
```

If you don't know the data size it makes sense to provide some big value (like 1012 * 1012 * 100 for 100MB for example), or to use "content-length" header that tells you the size of the data.

If you don't trust the source it's still possible to provide `upTo` param in the `public func bytes(upTo maxBytes: Int? = nil) async throws -> ByteBuffer` method above.

Do you have a better idea? 🙂

Lukasa: A better version might be to provide a maximum value the user is willing to accept here, and then use content-length _or_ the maximum value.

makoni: Thanks Lukasa. I've updated the PR, made `upTo` param required for both helper functions as you suggested. If the content-length is larger than the max there will be an error thrown. Also updated the tests.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
```diff
@@ -138,6 +138,17 @@ extension HTTPClientResponse {
     }
 }
 
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientResponse {
+    /// Response body as `ByteBuffer`.
+    /// - Parameter maxBytes: The maximum number of bytes this method is allowed to accumulate. 
+    /// - Returns: Bytes collected over time
+    public func bytes(upTo maxBytes: Int) async throws -> ByteBuffer {
+        let expectedBytes = self.headers.first(name: "content-length").flatMap(Int.init) ?? maxBytes
```
Lukasa, line 147: This shouldn't just blindly trust the content-length: it should reject values of the content length above the `upTo` value.

Sources/AsyncHTTPClient/FoundationExtensions.swift
```diff
@@ -64,3 +64,14 @@ extension HTTPClient.Body {
         return self.bytes(data)
     }
 }
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientResponse {
+    /// Response body as `Data`.
+    /// - Parameter maxBytes: The maximum number of bytes this method is allowed to accumulate.
+    /// - Returns: Bytes collected over time
+    public func data(upTo maxBytes: Int) async throws -> Data? {
```
Lukasa, line 73: Same note here.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift
makoni: Updated

Sources/AsyncHTTPClient/FoundationExtensions.swift
makoni: Updated

makoni:
It's been a while but I updated it :)

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse+data.swift
```diff
@@ -0,0 +1,41 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa:
```suggestion
// Copyright (c) 2025 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse+data.swift
```diff
@@ -0,0 +1,41 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+import NIOHTTP1
+import Foundation
```
Lukasa, line 19: Let's replace this with a dance that avoids importing all of foundation when not needed:

```suggestion
#if canImport(FoundationEssentials)
import FoundationEssentials
#else
import Foundation
#endif
```

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse+data.swift
```diff
@@ -0,0 +1,41 @@
...
+internal import NIOFoundationCompat
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientResponse {
+    /// Response body as `ByteBuffer`.
+    /// - Parameter maxBytes: The maximum number of bytes this method is allowed to accumulate.
+    /// - Returns: Bytes collected over time
+    public func bytes(upTo maxBytes: Int) async throws -> ByteBuffer {
+        let expectedBytes = self.headers.first(name: "content-length").flatMap(Int.init) ?? maxBytes
```
Lukasa: We should always use `maxBytes` and never the value from `content-length`.

makoni: Done

Issue #739, opened 2024-04-13
Too low of a decompression ratio in `HTTPClient.shared` settings


MahdiBM:
[Penny](https://github.com/vapor/penny-bot) is having difficulty decompressing some GitHub responses after the recent Penny update to use `HTTPClient.shared`.
The decompression limit of 10-ratio seems to be too low for a request to `https://api.github.com/repos/vapor/sql-kit/contributors`, currently. This might not be reproducible sometime soon when the contents of the endpoint changes.

To reproduce:
```bash
docker run --rm -it swift:5.10-jammy bash -c "git clone https://github.com/MahdiBM/AHCLowDecompRatioIssue && cd AHCLowDecompRatioIssue && swift run"
```

MahdiBM:
For future reference:
```bash
curl https://api.github.com/repos/vapor/sql-kit/contributors -v \                                                                                                                             
    -H "User-Agent: AHC-test" \
    &> ahc-decomp-failure.txt
```
[ahc-decomp-failure.txt](https://github.com/swift-server/async-http-client/files/14971322/ahc-decomp-failure.txt)

Lukasa:
I'm not necessarily inclined to change that setting: I think it's appropriate to leave it not far from where it is. You can create another singleton for your use-case :+1:

MahdiBM:
@Lukasa that'd be fine by me to even revert the whole or most of the PR I did for Penny to use `HTTPClient.shared` (I regret it), but from my POV it seems like the decompression limit is just too low.

I didn't find proper references on what ratio would be reasonable when searching through some Go and Rust HTTP client libraries. If a ratio of 10 is a widely accepted value, I'm fine with keeping it at 10. Otherwise the only reference I found on Google (the link didn't even open for me, but was already indexed on Google) mentioned 95% compression is around the max value for normal text files, so I would suggest increasing the ratio to 15/20/25 (each representing 93.3, 95 and 96% compression of a file).

Optionally we could just disable decompression, but that would make the configuration less like what a browser like Firefox would do.

Lukasa:
My concern is only that I don't want to expose users to sudden spikes in memory usage caused by decompression. I'm tentatively open to widening the ratio, but I don't think I'd want us to go much beyond 25.

MahdiBM:
I would have marked this issue as resolved if it wasn't for https://github.com/swiftlang/swiftly/pull/120#discussion_r1639955001...

Let's mention this issue there and see what happens.

[...]

PR #740, opened 2024-04-16
Increase decompression limit ratio of `HTTPClient.shared` 10 -> 25

MahdiBM:
This enables the decompressor to decompress files that have been 96% compressed, instead of the previous 90%, when using `HTTPClient.shared`.

See https://github.com/swift-server/async-http-client/issues/739.

I'm also fine with decreasing the ratio to 20, if 25 seems too high.


MahdiBM:
@Lukasa since you were responding on the issue 🙂

FranzBusch:

Lukasa (approved):
Nice, LGTM.

Lukasa:

Issue #737, opened 2024-04-09
Crash: precondition failure in HTTPConnectionPool+HTTP2Connections.swift:209


nosyjoe:
When trying a GET request to `https://www.rtp.pt/play/podcast/252` using this code 

```swift
import AsyncHTTPClient

let request = HTTPClientRequest(url: "https://www.rtp.pt/play/podcast/252")
let response = try await HTTPClient.shared.execute(request, timeout: .seconds(30))
```

this crash happens:

```
#0	0x00000001989d9564 in _swift_runtime_on_report ()
#1	0x0000000198a9798c in _swift_stdlib_reportFatalErrorInFile ()
#2	0x00000001986860e4 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x00000001986851e0 in _assertionFailure(_:_:file:line:flags:) ()
#4	0x00000001000da554 in HTTPConnectionPool.HTTP2ConnectionState.release() at /async-http-crash-sample-fdvnroepwpjkbacttjkwxnbpltan/SourcePackages/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift:209
#5	0x00000001000e0528 in HTTPConnectionPool.HTTP2Connections.releaseStream(_:) at /async-http-crash-sample-fdvnroepwpjkbacttjkwxnbpltan/SourcePackages/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2Connections.swift:644
#6	0x00000001000e9fd0 in HTTPConnectionPool.HTTP2StateMachine.http2ConnectionStreamClosed(_:) at /async-http-crash-sample-fdvnroepwpjkbacttjkwxnbpltan/SourcePackages/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP2StateMachine.swift:404
#7	0x00000001000f56f0 in closure #2 in HTTPConnectionPool.StateMachine.http2ConnectionStreamClosed(_:) at /async-http-crash-sample-fdvnroepwpjkbacttjkwxnbpltan/SourcePackages/checkouts/async-http-client/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+StateMachine.swift:218
[...]

PR #741, opened 2024-04-29
Disable SETTINGS_ENABLE_PUSH HTTP/2 setting

gjcairo:
This solves https://github.com/swift-server/async-http-client/issues/737

We are currently creating H2 connections using the default NIO settings, but this means we weren't sending `SETTING_ENABLE_PUSH=0`. This caused the server to push streams to the client, which messed up the stream count logic (as we were assuming only the client could create new streams), resulting in a precondition failure when releasing streams.

This PR updates the settings to disable the `enablePush` setting.


FranzBusch (commented):
Can we add a tests that shows push streams are being rejected?

gjcairo:

FranzBusch (approved):

PR #744, opened 2024-06-16
Make ConnectionPool's `retryConnectionEstablishment` public

### LINKED ISSUE #743: `Connection refused` not throwing right away
AsyncHTTPClient commit hash: a22083713ee90808d527d0baa290c2fb13ca3096 (Release 1.21.1)

While trying to use `HTTPClientRequest` on macOS (with `NIOTransportServices`), I noticed that refused connection errors are only thrown when the connection timeout expires.
With the following code, it takes 90 seconds (the connection timeout specified in `HTTPClient.Configuration.singletonConfiguration`) until the error is thrown:

```swift
import AsyncHTTPClient

var request = HTTPClientRequest(url: "https://localhost:12345/")
let response = try await HTTPClient.shared.execute(request, timeout: .hours(1))
```
I'd consider a refused connection a non-recoverable error and would expect `execute` to throw right away.

- Is awaiting the timeout expected behavior here?
- Is there a way to not wait in this situation while using `HTTPClientRequest`?

<br/>

**Environment:**
$ swift --version
swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: arm64-apple-macosx14.0

Operating system: macOS Sonoma 14.5 (23F79)

andreasley:
Currently, there is no way to instantly return when a connection attempt fails. The connection timeout is always awaited, regardless of the connection error. When used within a service, this behavior makes sense – network conditions may temporarily prevent establishing a connection. However, in a user-interactive environment, it's preferable to fail quickly and display the error to the user.

#625 introduced the property [`retryConnectionEstablishment`](https://github.com/swift-server/async-http-client/blob/0ae99db85b2b9d1e79b362bd31fd1ffe492f7c47/Sources/AsyncHTTPClient/HTTPClient.swift#L1028), which defaults to `true` but can be set to `false` to **not** retry on connection failure. The property was added to speed up unit tests, but would also be suitable for the use-case mentioned above.

This PR makes the property `retryConnectionEstablishment`, which is already documented and well tested, `public`.

This would fix #743.

The PR also cleans up some unit tests to consistently use [`enableFastFailureModeForTesting()`](https://github.com/swift-server/async-http-client/blob/0ae99db85b2b9d1e79b362bd31fd1ffe492f7c47/Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift#L595) instead of setting `retryConnectionEstablishment` directly.

**Questions:**

- When using NIOTS, the HTTPClient.Configuration's property [`networkFrameworkWaitForConnectivity`](https://github.com/swift-server/async-http-client/blob/0ae99db85b2b9d1e79b362bd31fd1ffe492f7c47/Sources/AsyncHTTPClient/HTTPClient.swift#L727) **also** needs to be set to `false` to enable fast fail. Would it be worthwhile to have the ability to set both properties at once?

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -1031,10 +1031,10 @@ extension HTTPClient.Configuration {
             self.init(idleTimeout: idleTimeout, concurrentHTTP1ConnectionsPerHostSoftLimit: 8)
         }
 
-        public init(idleTimeout: TimeAmount, concurrentHTTP1ConnectionsPerHostSoftLimit: Int) {
+        public init(idleTimeout: TimeAmount, concurrentHTTP1ConnectionsPerHostSoftLimit: Int, retryConnectionEstablishment: Bool = true) {
             self.idleTimeout = idleTimeout
             self.concurrentHTTP1ConnectionsPerHostSoftLimit = concurrentHTTP1ConnectionsPerHostSoftLimit
-            self.retryConnectionEstablishment = true
+            self.retryConnectionEstablishment = retryConnectionEstablishment
         }
     }
 
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -1012,11 +1012,11 @@ extension HTTPClient.Configuration {
     public struct ConnectionPool: Hashable, Sendable {
         /// Specifies amount of time connections are kept idle in the pool. After this time has passed without a new
         /// request the connections are closed.
-        public var idleTimeout: TimeAmount
+        public var idleTimeout: TimeAmount = .seconds(60)
 
         /// The maximum number of connections that are kept alive in the connection pool per host. If requests with
         /// an explicit eventLoopRequirement are sent, this number might be exceeded due to overflow connections.
-        public var concurrentHTTP1ConnectionsPerHostSoftLimit: Int
+        public var concurrentHTTP1ConnectionsPerHostSoftLimit: Int = 8
 
         /// If true, ``HTTPClient`` will try to create new connections on connection failure with an exponential backoff.
         /// Requests will only fail after the ``HTTPClient/Configuration/Timeout-swift.struct/connect`` timeout exceeded.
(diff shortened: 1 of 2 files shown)


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1025,16 +1025,16 @@ extension HTTPClient.Configuration {
...
         /// ``HTTPClient`` will automatically mitigate these kind of issues if this flag is turned on.
-        var retryConnectionEstablishment: Bool
+        public var retryConnectionEstablishment: Bool
 
         public init(idleTimeout: TimeAmount = .seconds(60)) {
             self.init(idleTimeout: idleTimeout, concurrentHTTP1ConnectionsPerHostSoftLimit: 8)
         }
 
-        public init(idleTimeout: TimeAmount, concurrentHTTP1ConnectionsPerHostSoftLimit: Int) {
+        public init(idleTimeout: TimeAmount, concurrentHTTP1ConnectionsPerHostSoftLimit: Int, retryConnectionEstablishment: Bool = true) {
```
Lukasa, line 1034: Unfortunately it's not API stable to add this extra parameter. Rather than us keep proliferating new `init`s, it might be nice for us to just add a zero-element initializer and encourage users towards setting the fields after default construction.

Lukasa (commented):
Very nice, thank you! Just a single quick note.

Lukasa:

Sources/AsyncHTTPClient/HTTPClient.swift
andreasley: Thanks for taking a look! I've...

- Reverted the change to the existing initializer.
- Added a parameterless initializer.
- Removed the default value for the parameter `idleTimeout` in the respective initializer (because it overlaps with the new parameterless initializer). Is that safe in terms of API stability?
- Moved the default values to the respective property declarations (to prevent them having to be specified at multiple points in the initializer code). This seems cleaner to me than initializer delegation, but if you prefer a different style, let me know.

Lukasa:

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1025,16 +1025,17 @@ extension HTTPClient.Configuration {
         /// - warning: We highly recommend leaving this on.
         /// It is very common that connections establishment is flaky at scale.
         /// ``HTTPClient`` will automatically mitigate these kind of issues if this flag is turned on.
-        var retryConnectionEstablishment: Bool
+        public var retryConnectionEstablishment: Bool = true
 
-        public init(idleTimeout: TimeAmount = .seconds(60)) {
-            self.init(idleTimeout: idleTimeout, concurrentHTTP1ConnectionsPerHostSoftLimit: 8)
+        public init() {}
+
```
Lukasa, line 1031: There are a few spaces here that the soundness check objects to.

```suggestion

```

Lukasa (commented):
Basically fine, one quick note.

fabianfett:

Lukasa (approved):

Lukasa:
I'm ignoring API breakage here: it's concerned about us removing the default argument, but we've replaced it with a zero-argument constructor so it's safe from an API perspective.

PR #746, opened 2024-06-24
Add support for request body to be larger than 2GB on 32-bit devices

aryan-25:
### Motivation:

- The properties that store the request body length and the cumulative number of bytes sent as part of a request are of type `Int`. 
- On 32-bit devices, when sending requests larger than `Int32.max`, these properties overflow and cause a crash.
- To solve this problem, the properties should use the explicit `Int64` type.

### Modifications:

- Changed the type of the `known` field of the `RequestBodyLength` enum to `Int64`.
- Changed the type of `expectedBodyLength` and `sentBodyBytes` in `HTTPRequestStateMachine` to `Int64?` and `Int64` respectively.
- Deprecated the `public var length: Int?` property of `HTTPClient.Body` and backed it with a new property: `contentLength: Int64?`
  - Added a new initializer and "overloaded" the `stream` function in `HTTPClient.Body` to work with the new `contentLength` property.
  - **Note:** The newly added `stream` function has different parameter names (`length` -> `contentLength` and `stream` -> `bodyStream`) to avoid ambiguity problems.
- Added a test case that streams a 3GB request -- verified this fails with the types of the properties set explicitly to `Int32`. 

### Result:

- 32-bit devices can send requests larger than 2GB without integer overflow issues.

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
@@ -140,7 +140,7 @@ extension HTTPClientRequest.Body {
     ///
     /// Caution should be taken with this method to ensure that the `length` is correct. Incorrect lengths
     /// will cause unnecessary runtime failures. Setting `length` to ``Length/unknown`` will trigger the upload
-    /// to use `chunked` `Transfer-Encoding`, while using ``Length/known(_:)`` will use `Content-Length`.
+    /// to use `chunked` `Transfer-Encoding`, while using ``Length/known(_:)-9q0ge`` will use `Content-Length`.
     ///
     /// - parameters:
     ///     - bytes: The bytes of the request body.
@@ -225,7 +225,7 @@ extension HTTPClientRequest.Body {
     ///
     /// Caution should be taken with this method to ensure that the `length` is correct. Incorrect lengths
     /// will cause unnecessary runtime failures. Setting `length` to ``Length/unknown`` will trigger the upload
-    /// to use `chunked` `Transfer-Encoding`, while using ``Length/known(_:)`` will use `Content-Length`.
+    /// to use `chunked` `Transfer-Encoding`, while using ``Length/known(_:)-9q0ge`` will use `Content-Length`.
     ///
     /// - parameters:
     ///     - bytes: The bytes of the request body.
@@ -265,7 +265,7 @@ extension HTTPClientRequest.Body {
     ///
     /// Caution should be taken with this method to ensure that the `length` is correct. Incorrect lengths
     /// will cause unnecessary runtime failures. Setting `length` to ``Length/unknown`` will trigger the upload
-    /// to use `chunked` `Transfer-Encoding`, while using ``Length/known(_:)`` will use `Content-Length`.
+    /// to use `chunked` `Transfer-Encoding`, while using ``Length/known(_:)-9q0ge`` will use `Content-Length`.
     ///
     /// - parameters:
     ///     - sequenceOfBytes: The bytes of the request body.
@@ -293,7 +293,7 @@ extension HTTPClientRequest.Body {
     ///
     /// Caution should be taken with this method to ensure that the `length` is correct. Incorrect lengths
     /// will cause unnecessary runtime failures. Setting `length` to ``Length/unknown`` will trigger the upload
-    /// to use `chunked` `Transfer-Encoding`, while using ``Length/known(_:)`` will use `Content-Length`.
+    /// to use `chunked` `Transfer-Encoding`, while using ``Length/known(_:)-9q0ge`` will use `Content-Length`.
     ///
     /// - parameters:
     ///     - bytes: The bytes of the request body.
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
@@ -341,11 +341,11 @@ extension HTTPClientRequest.Body {
         public static let unknown: Self = .init(storage: .unknown)
 
         /// The size of the request body is known and exactly `count` bytes
-        @_disfavoredOverload
+        @available(*, deprecated, message: "Use `known(_ count: Int64)` instead")
         public static func known(_ count: Int) -> Self {
             .init(storage: .known(Int64(count)))
         }
-        
+
         /// The size of the request body is known and exactly `count` bytes
         public static func known(_ count: Int64) -> Self {
             .init(storage: .known(count))
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
@@ -125,7 +125,7 @@ extension HTTPClientRequest.Body {
     public static func bytes<Bytes: RandomAccessCollection & Sendable>(
         _ bytes: Bytes
     ) -> Self where Bytes.Element == UInt8 {
-        self.bytes(bytes, length: .known(bytes.count))
+        self.bytes(bytes, length: .known(Int64(bytes.count)))
     }
 
     /// Create an ``HTTPClientRequest/Body-swift.struct`` from a `Sequence` of bytes.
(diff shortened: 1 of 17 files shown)


fabianfett:

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -70,16 +70,35 @@ extension HTTPClient {
...
 
         @usableFromInline typealias StreamCallback = @Sendable (StreamWriter) -> EventLoopFuture<Void>
 
         @inlinable
+        @available(*, deprecated, message: "Use initializer with `contentLength: Int64` instead")
         init(length: Int?, stream: @escaping StreamCallback) {
-            self.length = length
+            self.contentLength = length.flatMap { Int64($0) }
+            self.stream = stream
+        }
```
fabianfett, line 97: This is an internal function. We should just remove it.

Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -184,6 +184,56 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
...
+        let sequence = AsyncSequenceByteBufferGenerator(
+            chunkSize: 4_194_304, // 4MB chunk
+            totalChunks: 768 // Total = 3GB
+        )
+        request.body = .stream(sequence, length: .unknown)
+        
+        let response: HTTPClientResponse = try await client.execute(request, deadline: .now() + .seconds(30), logger: logger)
+        XCTAssertEqual(response.headers["content-length"], [])
+        
+        _ = try await response.body.collect(upTo: 3_221_225_472)
```
fabianfett, line 234: Please don't collect here. :) Otherwise this makes the tests extremely memory hungry :)

```swift
var receivedBytes: Int64 = 0
for try await part in response.body {
    receivedBytes += Int64(part.readableBytes)
}
XTCAssertEqual(receivedBytes, expectedSentBytes)
```

Tests/AsyncHTTPClientTests/RequestBagTests.swift
```diff
@@ -19,31 +19,32 @@ import NIOEmbedded
...
         let logger = Logger(label: "test")
 
-        var writtenBytes = 0
-        var writes = 0
-        let bytesToSent = (3000...10000).randomElement()!
-        let expectedWrites = bytesToSent / 100 + ((bytesToSent % 100 > 0) ? 1 : 0)
-        var streamIsAllowedToWrite = false
+        let writtenBytes: ManagedAtomic<Int> = .init(0)
+        let writes: ManagedAtomic<Int> = .init(0)
+        let bytesToSent: ManagedAtomic<Int> = .init((3000...10000).randomElement()!)
```
fabianfett, line 32: `bytesToSend` doesn't look mutable here. Why do we need an atomic here?

Tests/AsyncHTTPClientTests/RequestBagTests.swift
```diff
@@ -19,31 +19,32 @@ import NIOEmbedded
...
+        let expectedWrites = bytesToSent.load(ordering: .relaxed) / 100 + ((bytesToSent.load(ordering: .relaxed) % 100 > 0) ? 1 : 0)
```
fabianfett, line 33: `expectedWrites` doesn't look mutable here. Why do we need an atomic here?

fabianfett (commented):
This already looks very god! Great patch! @Lukasa can you please check the API changes?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -341,7 +341,13 @@ extension HTTPClientRequest.Body {
         public static let unknown: Self = .init(storage: .unknown)
 
         /// The size of the request body is known and exactly `count` bytes
+        @_disfavoredOverload
         public static func known(_ count: Int) -> Self {
```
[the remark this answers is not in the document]
Lukasa: I think we have to deprecate.

The number of times this is called with an integer literal is actually very low. This will usually be called with a computed value, which will be of type `Int`. That means the issue isn't actually at this location: the user failed to even _compute_ the right size.

We need to push users over to computing on sizes in `Int64`.

One option, if we wanted to do it, would be to confine the deprecation to the 32-bit platforms (which we can do by a series of clever games with `#if`). But I have the voice of @weissi in my head telling me that that's a bad idea because it produces warnings that most libraries will never see, meaning they won't actually address their issues. So I guess I tentatively favour deprecation.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -100,18 +119,30 @@ extension HTTPClient {
         /// header is set with the given `length`.
         ///     - stream: Body chunk provider.
         @preconcurrency
+        @available(*, deprecated, message: "Use `stream` with `length: Int64` instead")
```
Lukasa, line 122: We don't need to use messages here, we can use `renamed` and write out the new function name:

```suggestion
        @available(*, deprecated, renamed: "stream(contentLength:bodyStream:)")
```

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -70,16 +70,35 @@ extension HTTPClient {
 
         /// Body size. If nil,`Transfer-Encoding` will automatically be set to `chunked`. Otherwise a `Content-Length`
         /// header is set with the given `length`.
-        public var length: Int?
+        @available(*, deprecated, message: "Use `contentLength: Int64?` instead")
```
Lukasa, line 73: Again, use `renamed` here.

Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -184,6 +184,56 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
...
+        init(chunkSize: Int, totalChunks: Int) {
+            self.chunkSize = chunkSize
+            self.totalChunks = totalChunks
+        }
+        
+        mutating func next() async throws -> ByteBuffer? {
+            guard self.chunksGenerated < self.totalChunks else { return nil }
+
+            self.chunksGenerated += 1
+            return ByteBuffer(repeating: 1, count: self.chunkSize)
```
Lukasa, line 203: Just generate the one chunk and send it again and again.

Tests/AsyncHTTPClientTests/HTTP1ConnectionTests.swift
```diff
@@ -116,8 +116,8 @@ class HTTP1ConnectionTests: XCTestCase {
         XCTAssertNoThrow(maybeRequest = try HTTPClient.Request(
             url: "http://localhost/hello/swift",
             method: .POST,
-            body: .stream(length: 4) { writer -> EventLoopFuture<Void> in
-                func recursive(count: UInt8, promise: EventLoopPromise<Void>) {
+            body: .stream(contentLength: 4) { writer -> EventLoopFuture<Void> in
+                @Sendable func recursive(count: UInt8, promise: EventLoopPromise<Void>) {
```
Lukasa, line 120: I don't really think we should have needed this annotation, tbh: the compiler should have been able to infer it.

Tests/AsyncHTTPClientTests/RequestBagTests.swift
Lukasa: I don't think we should be using atomics on any of this, it's a bit too overly clever. Locks are just fine.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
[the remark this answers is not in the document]
weissi: +1 on deprecation. I think it's beneficial:

- As @Lukasa points out: the use cases for `.known(...)` are limited -> blast radius of people who we annoy needlessly is small
- We do want the people who use `.known(Int)` to migrate

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -341,7 +341,13 @@ extension HTTPClientRequest.Body {
         public static let unknown: Self = .init(storage: .unknown)
 
         /// The size of the request body is known and exactly `count` bytes
+        @available(*, deprecated, message: "Use `known(_ count: Int64)` instead")
```
Lukasa, line 344: I think this message can be more helpful: let's explicitly point out we need the argument to be Int64 instead of Int.

Lukasa (approved):

Lukasa:

fabianfett (approved):
Great change!

PR #747, opened 2024-06-28
Fix leading slash issue in relative URL requests

aryan-25:
### Motivation:

- When the leading slash is omitted in the `string` argument of a relative `URL` instance constructed from `init(string:relativeTo)`, requests made with this `URL` instance fail. This is because the underlying HTTP parser fails when there is no leading slash in the URI.

### Modifications:

- The URI of a `URL` instance is obtained from a computed property `uri` in `HTTPHandler.swift`, which in turn calls the `percentEncodedPath` computed property.
  - The `percentEncodedPath` property uses `URLComponents` to extract out the path.
  - The `resolvingAgainstBaseURL` argument to the `URLComponents` constructor was modified from `false` to `true` 

### Result:

- Regardless of whether a user provides or omits a leading slash in the path of a relative `URL`, the `uri` property will always contain a leading slash.

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
@@ -142,6 +142,24 @@ class HTTPClientInternalTests: XCTestCase {
         XCTAssertEqual(request12.url.uri, "/some%2Fpathsegment1/pathsegment2")
     }
 
+    func testURIOfRelativeURLRequest() throws {
+        let requestNoLeadingSlash = try Request(
+            url: URL(
+                string: "percent%2Fencoded/hello",
+                relativeTo: URL(string: "http://127.0.0.1")!
+            )!
+        )
+
+        let requestWithLeadingSlash = try Request(
+            url: URL(
+                string: "/percent%2Fencoded/hello",
+                relativeTo: URL(string: "http://127.0.0.1")!
+            )!
+        )
+
+        XCTAssertEqual(requestNoLeadingSlash.url.uri, requestWithLeadingSlash.url.uri)
+    }
+
     func testChannelAndDelegateOnDifferentEventLoops() throws {
         class Delegate: HTTPClientResponseDelegate {
             typealias Response = ([Message], [Message])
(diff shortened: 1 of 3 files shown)


Tests/AsyncHTTPClientTests/HTTPClientInternalTests.swift
```diff
@@ -142,6 +142,24 @@ class HTTPClientInternalTests: XCTestCase {
...
+        )
+
+        let requestWithLeadingSlash = try Request(
+            url: URL(
+                string: "/percent%2Fencoded/hello",
+                relativeTo: URL(string: "http://127.0.0.1")!
+            )!
+        )
+
+        XCTAssertEqual(requestNoLeadingSlash.url.uri, requestWithLeadingSlash.url.uri)
```
fabianfett, line 160: mind putting the absolute values here?

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -495,6 +495,19 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+        let noLeadingSlashURL = URL(string: "percent%2Fencoded/hello", relativeTo: URL(string: self.defaultHTTPBinURLPrefix)!)!
+        let withLeadingSlashURL = URL(string: "/percent%2Fencoded/hello", relativeTo: URL(string: self.defaultHTTPBinURLPrefix)!)!
+
+        let noLeadingSlashURLRequest = try HTTPClient.Request(url: noLeadingSlashURL, method: .GET)
+        let withLeadingSlashURLRequest = try HTTPClient.Request(url: withLeadingSlashURL, method: .GET)
+
+        let noLeadingSlashURLResponse = try self.defaultClient.execute(request: noLeadingSlashURLRequest).wait()
+        let withLeadingSlashURLResponse = try self.defaultClient.execute(request: withLeadingSlashURLRequest).wait()
+
+        XCTAssertEqual(noLeadingSlashURLResponse.status, withLeadingSlashURLResponse.status)
```
fabianfett, line 508: can we compare against absolute values here please?

fabianfett (commented):
Looks great. Please compare against static absolute values in tests.

PR #749, opened 2024-07-03
Resolve ambiguity issue for the `stream` function

aryan-25:
### Motivation:

- The `stream` function in `HTTPClientRequest.Body` (newly added in #746) is ambiguous with the existing `stream` function as the first argument has a default value and the second argument is a closure. This is API-breaking. 

### Modifications:

- Removed the `@deprecated` marking from the original `stream` function and instead added `@_disfavoredOverload`.

### Result:

- There are no more ambiguity issues.


glbrntt (approved):

fabianfett (approved):
LGTM

FranzBusch:
API break is expected. Merging over it

PR #758, opened 2024-08-07
Assume http2 connection by default, instead of http1

ayush1794:
Since most of the servers now conform to http2, the change here updates the behaviour of assuming the connection to be http2 and not http1 by default. It will migrate to http1 if the server only supports http1. 
One can set the `httpVersion` in `ClientConfiguration` to `.http1Only` which will start with http1 instead of http2.

Additional Changes:

- Fixed an off by one error in the maximum additional general purpose connection check
- Updated tests

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
@@ -596,10 +596,10 @@ extension HTTPConnectionPool {
                     eventLoop: eventLoop,
                     maximumUses: self.maximumConnectionUses
                 )
-                
+
                 /// Mark as overflow connection if any request is pending on a starting connection
                 /// on the same event loop, else mark as general purpose connection
-                if requiredEventLoopPendingRequests.first(where: {$0.0.id == eventLoop.id}) != nil {
+                if requiredEventLoopPendingRequests.first(where: { $0.0.id == eventLoop.id }) != nil {
                     self.connections.append(newConnection)
                 } else {
                     self.connections.insert(newConnection, at: self.overflowIndex)
@@ -619,10 +619,10 @@ extension HTTPConnectionPool {
                 )
                 // TODO: Maybe we want to add a static init for backing off connections to HTTP1ConnectionState
                 backingOffConnection.failedToConnect()
-                
+
                 /// Mark as overflow connection if any request is pending on a backing off connection
                 /// on the same event loop, else mark as general purpose connection
-                if requiredEventLoopPendingRequests.first(where: {$0.0.id == eventLoop.id}) != nil {
+                if requiredEventLoopPendingRequests.first(where: { $0.0.id == eventLoop.id }) != nil {
                     self.connections.append(backingOffConnection)
                 } else {
                     self.connections.insert(backingOffConnection, at: self.overflowIndex)
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
@@ -307,7 +307,7 @@ extension HTTPConnectionPool {
         }
 
         private var maximumAdditionalGeneralPurposeConnections: Int {
-            self.maximumConcurrentConnections - (self.overflowIndex - 1)
+            self.maximumConcurrentConnections - (self.overflowIndex)
         }
 
         /// Is there at least one connection that is able to run requests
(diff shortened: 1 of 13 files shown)


czechboy0:

dnadoba:

dnadoba:
LGTM. I'm wondering if we need to expose this settings not only through `.http1Only` but rather as initial preference that starts with http1 but still allows http2. @weissi what do you think?

weissi:
> LGTM. I'm wondering if we need to expose this settings not only through `.http1Only` but rather as initial preference that starts with http1 but still allows http2. @weissi what do you think?

TIL that we're not already doing that. Seems sensible

dnadoba:
>Yes, we'll likely need that but that could also be a follow-up?

agree

ayush1794:
> > LGTM. I'm wondering if we need to expose this settings not only through `.http1Only` but rather as initial preference that starts with http1 but still allows http2. @weissi what do you think?
> 
> TIL that we're not already doing that. Seems sensible to default to HTTP/2.
> 
> Regarding the 'prefer H1' override in public API: Yes, we'll likely need that but that could also be a follow-up?

Just out of curiosity what would be a real use case for this? 
If we do this, we should likely also need to change the [ALPN ordering](https://github.com/swift-server/async-http-client/blob/54d1006dc90b9ee77b4d04d63ad8688d2215dc1e/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool%2BFactory.swift#L405) in that case to prefer h1 over h2 when TLS is needed (_though again not sure of a TLS use case on h1_)

weissi:
> > Regarding the 'prefer H1' override in public API: Yes, we'll likely need that but that could also be a follow-up?
> 
> Just out of curiosity what would be a real use case for this? If we do this, we should likely also need to change the [ALPN ordering](https://github.com/swift-server/async-http-client/blob/54d1006dc90b9ee77b4d04d63ad8688d2215dc1e/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool%2BFactory.swift#L405) in that case to prefer h1 over h2 when TLS is needed (_though again not sure of a TLS use case on h1_)

Four reasons come to mind:

1. At the moment you can instruct AHC to create an arbitrary amount of HTTP/1.1 connections to a given host. That means you can for example have 1,000 concurrent requests to a given server. But at present, with H2, AHC will only ever do one connection which usually allows up to 100 concurrent requests (technically negotiated using `MAX_CONCURRENT_STREAMS` but in practice it's always 100)
2. Buggy software where HTTP/1.1 works better than HTTP/2 but the software claims to support HTTP/2
3. (very) Low-latency connections: HTTP/1.1 can be performant than HTTP/2 if you have essentially 0 latency (e.g. on `localhost`). HTTP/2 trades off more CPU usage on client & server for fewer roundtrips. That's usually a good tradeoff but the lower the latency, the worse it gets.
4. Integration tests for server software that wants to check that HTTP/1.x also works. Of course this means we shouldn't actually have a `preferHTTP1: Bool` but rather `{prefer,require}HTTPProtocolVersion: .v1_1 | .v2 | .v1_0`

ayush1794:
@swift-server-bot test

ayush1794:

Tests/AsyncHTTPClientTests/HTTPConnectionPool+HTTP2StateMachineTests.swift
```diff
@@ -1146,11 +1151,12 @@ class HTTPConnectionPool_HTTP2StateMachineTests: XCTestCase {
             idGenerator: .init(),
             maximumConcurrentHTTP1Connections: 8,
             retryConnectionEstablishment: true,
+            preferHTTP1: false,
             maximumConnectionUses: nil
         )
 
         var connectionIDs: [HTTPConnectionPool.Connection.ID] = []
-        for el in [el1, el2, el2] {
+        for el in [el1, el2] {
```
fabianfett, line 1159: why did you remove `el2` here?

Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1Connections.swift
```diff
@@ -584,21 +584,30 @@ extension HTTPConnectionPool {
...
-                self.connections.insert(newConnection, at: self.overflowIndex)
-                /// If we can grow, we mark the connection as a general purpose connection.
-                /// Otherwise, it will be an overflow connection which is only used once for requests with a required event loop
-                if self.canGrow {
-                    self.overflowIndex = self.connections.index(after: self.overflowIndex)
+
+                /// Mark as overflow connection if any request is pending on a starting connection
+                /// on the same event loop, else mark as general purpose connection
+                if requiredEventLoopPendingRequests.first(where: { $0.0.id == eventLoop.id }) != nil {
+                    self.connections.append(newConnection)
```
fabianfett, line 603: Why do we have this preference for overflow connections here?

Tests/AsyncHTTPClientTests/HTTPConnectionPool+HTTP2StateMachineTests.swift
ayush1794: This is because we now directly start with h2 rather than h1 and then migrating to h1. 
h2 has checks to not allow creating connections for same EL. The second el2 just fails the test because the check fails.

fabianfett (approved):
Thanks for pushing this through!

PR #760, opened 2024-08-07
enable StrictConcurrency build setting

weissi:
Set

```
swiftSettings: [
    .enableExperimentalFeature("StrictConcurrency")
]
```

so we know where there's work.


Lukasa (changes requested):
This is currently incompatible with warnings-as-errors, so we need some CI tweaks to handle it.

weissi:
> This is currently incompatible with warnings-as-errors, so we need some CI tweaks to handle it.

Ah yeah, should I just tweak the CI?

docker/docker-compose.yaml
```diff
@@ -32,7 +32,7 @@ services:
     
   test:
     <<: *common
-    command: /bin/bash -xcl "swift test --parallel -Xswiftc -warnings-as-errors --enable-test-discovery $${SANITIZER_ARG-} $${IMPORT_CHECK_ARG-}"
```
weissi, line 35: @Lukasa removed that

Lukasa:
That can work, though @FranzBusch is in the process of cutting over to a new CI scheme so we may want to hold off and tackle this slightly later.

It's also not clear that AHC can easily do this without us sorting out NIO first.

Issue #751, opened 2024-07-17
OOM if setting `concurrentHTTP1ConnectionsPerHostSoftLimit: .max`


weissi:
I would like to make the pool create as many concurrent connections as possible. So I thought the best way of spelling this is

```
        var httpClientConfig = HTTPClient.Configuration()
        httpClientConfig.connectionPool = .init(
            idleTimeout: .hours(1),
            concurrentHTTP1ConnectionsPerHostSoftLimit: .max
        )
        httpClientConfig.timeout = .init(connect: .seconds(10), read: .seconds(100), write: .seconds(100))
```

unfortunately, the `concurrentHTTP1ConnectionsPerHostSoftLimit: .max` makes AHC OOM even before it creates any connections because `HTTPConnectionPool.HTTP1Connections.init()` preallocates an array of that size.

It shouldn't do that or at least cap it to something reasonable like 1024.

natikgadzhi:
@weissi Do you think it should cap the actual pool size at 1024, or it should pre-initialize the first 1024, but allow making more of the app is actually trying to use more?

I'm not 100% certain but it feels like 1024 _should be enough for any single connection pool?_

MahdiBM:
> or it should pre-initialize the first 1024, but allow making more of the app is actually trying to use more?

That's what which has been implemented.

> I'm not 100% certain but it feels like 1024 should be enough for any single connection pool?

I wouldn't put a relatively arbitrary limit like that. We connect to our own internal services over http and there can be a lot of connections, although I don't think we've reached 1024.

PR #763, opened 2024-08-08
Fix OOM issue when setting `concurrentHTTP1ConnectionsPerHostSoftLimit` to `Int.max`

aryan-25:
### Motivation:

When a user wishes to make the connection pool create as many concurrent connections as possible, a natural way to achieve this would be to set `.max` to the `concurrentHTTP1ConnectionsPerHostSoftLimit` property.

```swift
HTTPClient.Configuration().connectionPool = .init(
    idleTimeout: .hours(1),
    concurrentHTTP1ConnectionsPerHostSoftLimit: .max
)
```

The `concurrentHTTP1ConnectionsPerHostSoftLimit` property is of type `Int`. Setting it to `Int.max` leads to `Int.max` being passed as an argument to `Array`s `.reserveCapacity(_:)` method, causing an OOM issue.

Addresses Github Issue #751 

### Modifications:

Capped the argument to `self.connections.reserveCapacity(_:)` to 1024 in `HTTPConnectionPool.HTTP1Connections`

### Result:

Users can now set the `concurrentHTTP1ConnectionsPerHostSoftLimit` property to `.max` without causing an OOM issue.

### DIFF
--- a/Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
+++ b/Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
@@ -642,6 +642,25 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
         }
     }
 
+    func testInsanelyHighConcurrentHTTP1ConnectionLimitDoesNotCrash() throws {
+        let bin = HTTPBin(.http1_1(compress: false))
+        defer { XCTAssertNoThrow(try bin.shutdown()) }
+
+        var httpClientConfig = HTTPClient.Configuration()
+        httpClientConfig.connectionPool = .init(
+            idleTimeout: .hours(1),
+            concurrentHTTP1ConnectionsPerHostSoftLimit: Int.max
+        )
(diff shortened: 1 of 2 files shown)


Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -642,6 +642,25 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
...
+            concurrentHTTP1ConnectionsPerHostSoftLimit: Int.max
+        )
+        httpClientConfig.timeout = .init(connect: .seconds(10), read: .seconds(100), write: .seconds(100))
+
+        let httpClient = HTTPClient(eventLoopGroupProvider: .shared(self.clientGroup), configuration: httpClientConfig)
+        defer {
+            XCTAssertNoThrow(try httpClient.syncShutdown())
+        }
+
+        let response = httpClient.get(url: "https://localhost:\(bin.port)", deadline: .now() + .seconds(2))
```
fabianfett, line 661: we need to use the async/await API here.

fabianfett (approved):

Lukasa (approved):
Love this, very nice.

PR #766, opened 2024-08-28
 Add an option to enable Multipath TCP on clients

Aperence:
Multipath TCP (MPTCP) is a TCP extension allowing to enhance the reliability of the network by using multiple interfaces. This extension provides a seamless handover between interfaces in case of deterioration of the connection on the original one. In the context of iOS and Mac OS X, it could be really interesting to leverage the capabilities of MPTCP as they could benefit from their multiple interfaces (ethernet + Wi-fi for Mac OS X, Wi-fi + cellular for iOS).

This contribution introduces patches to HTTPClient.Configuration and establishment of the Bootstraps. A supplementary field "enableMultipath" was added to the configuration, allowing to request the use of MPTCP. This flag is then used when creating the channels to configure the client.

Note that in the future, it might also be potentially interesting to offer more precise configuration options for MPTCP on MacOS, as the Network framework allows also to select a type of service, instead of just offering the option to create MPTCP connections. Currently, when enabling MPTCP, only the Handover mode is used.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -738,6 +738,10 @@ public class HTTPClient {
             }
         }
 
+        /// Whether ``HTTPClient`` will use Multipath TCP or not
+        /// By default, don't use it
+        public var enableMultipath: Bool
+
         public init(
             tlsConfiguration: TLSConfiguration? = nil,
             redirectConfiguration: RedirectConfiguration? = nil,
@@ -755,6 +759,7 @@ public class HTTPClient {
             self.decompression = decompression
             self.httpVersion = .automatic
             self.networkFrameworkWaitForConnectivity = true
+            self.enableMultipath = false
         }
 
         public init(tlsConfiguration: TLSConfiguration? = nil,
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -3600,8 +3600,10 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
             }
             let response = try client.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
             XCTAssertEqual(.ok, response.status)
-        } catch let error as IOError where error.errnoCode == EINVAL {
+        } catch let error as IOError where error.errnoCode == EINVAL || error.errnoCode == EPROTONOSUPPORT || error.errnoCode == ENOPROTOOPT {
             // some old Linux kernels don't support MPTCP, skip this test in this case
+            // see https://www.mptcp.dev/implementation.html for details about each type
+            // of error
             throw XCTSkip()
         }
     }
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -3600,7 +3600,7 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
             }
             let response = try client.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
             XCTAssertEqual(.ok, response.status)
-        }catch let error as IOError where error.errnoCode == EINVAL{
+        } catch let error as IOError where error.errnoCode == EINVAL {
             // some old Linux kernels don't support MPTCP, skip this test in this case
             throw XCTSkip()
         }
--- a/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTPClientTests.swift
@@ -3590,6 +3590,22 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
         XCTAssertEqual(.ok, response.status)
     }
 
+    func testClientWithMultipath() throws {
+        do {
+            var conf = HTTPClient.Configuration()
+            conf.enableMultipath = true
+            let client = HTTPClient(configuration: conf)
+            defer {
+                XCTAssertNoThrow(try client.shutdown().wait())
+            }
+            let response = try client.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
+            XCTAssertEqual(.ok, response.status)
+        }catch let error as IOError where error.errnoCode == EINVAL{
+            // some old Linux kernels don't support MPTCP, skip this test in this case
+            throw XCTSkip()
+        }
+    }
+
     func testSingletonClientWorks() throws {
         let response = try HTTPClient.shared.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
         XCTAssertEqual(.ok, response.status)
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
@@ -322,6 +322,7 @@ extension HTTPConnectionPool.ConnectionFactory {
         if #available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *), let tsBootstrap = NIOTSConnectionBootstrap(validatingGroup: eventLoop) {
             return tsBootstrap
                 .channelOption(NIOTSChannelOptions.waitForActivity, value: self.clientConfiguration.networkFrameworkWaitForConnectivity)
+                .channelOption(NIOTSChannelOptions.multipathServiceType, value: self.clientConfiguration.enableMultipath ? .handover : .disabled)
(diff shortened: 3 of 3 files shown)


Lukasa:
Hi @Aperence, sorry for the delay in getting a review here. I'll take a look now.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -755,22 +760,25 @@ public class HTTPClient {
...
         }
 
         public init(tlsConfiguration: TLSConfiguration? = nil,
                     redirectConfiguration: RedirectConfiguration? = nil,
                     timeout: Timeout = Timeout(),
                     proxy: Proxy? = nil,
                     ignoreUncleanSSLShutdown: Bool = false,
-                    decompression: Decompression = .disabled) {
+                    decompression: Decompression = .disabled,
+                    enableMultipath: Bool = false) {
```
Lukasa, line 772: Unfortunately, we can't add new parameters to public methods. This will need a new method that has the extra parameter.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -738,14 +738,19 @@ public class HTTPClient {
...
         public init(
             tlsConfiguration: TLSConfiguration? = nil,
             redirectConfiguration: RedirectConfiguration? = nil,
             timeout: Timeout = Timeout(),
             connectionPool: ConnectionPool = ConnectionPool(),
             proxy: Proxy? = nil,
             ignoreUncleanSSLShutdown: Bool = false,
-            decompression: Decompression = .disabled
+            decompression: Decompression = .disabled,
+            enableMultipath: Bool = false
```
Lukasa, line 753: Unfortunately, we can't add new parameters here. We need a new method, with the new parameter.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -780,7 +788,8 @@ public class HTTPClient {
                     maximumAllowedIdleTimeInConnectionPool: TimeAmount = .seconds(60),
                     proxy: Proxy? = nil,
                     ignoreUncleanSSLShutdown: Bool = false,
-                    decompression: Decompression = .disabled) {
+                    decompression: Decompression = .disabled,
+                    enableMultipath: Bool = false) {
```
Lukasa, line 792: Same note here.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -799,7 +809,8 @@ public class HTTPClient {
                     proxy: Proxy? = nil,
                     ignoreUncleanSSLShutdown: Bool = false,
                     decompression: Decompression = .disabled,
-                    backgroundActivityLogger: Logger?) {
+                    backgroundActivityLogger: Logger?,
+                    enableMultipath: Bool = false) {
```
Lukasa, line 813: Same note here.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -808,23 +819,26 @@ public class HTTPClient {
...
         }
 
         public init(certificateVerification: CertificateVerification,
                     redirectConfiguration: RedirectConfiguration? = nil,
                     timeout: Timeout = Timeout(),
                     proxy: Proxy? = nil,
                     ignoreUncleanSSLShutdown: Bool = false,
-                    decompression: Decompression = .disabled) {
+                    decompression: Decompression = .disabled,
+                    enableMultipath: Bool = false) {
```
Lukasa, line 832: Same note here.

Lukasa (changes requested):
Thanks for this! Generally it looks like a great addition. I've noted that we need to tweak the API surface a bit.

Can I also ask that you add at least one unit test that uses the feature? That'll ensure that we don't break it.

Sources/AsyncHTTPClient/HTTPClient.swift
Aperence: Does it require a new method, or do you prefer simply having the field enableMPTCP as public in the configuration, allowing to set it freely ?

Aperence:
Hello @Lukasa,

First, thanks for the review !

I've removed the changes to the methods' signatures. It is still possible to set the enableMultipath field as it is public, but if you prefer, I can also introduce a method for it. 

I've added also another test that enables MPTCP, and check that we can still fetch https://httpbin.org/get, as requested.

I remain available if other changes are needed. 
Have a nice day !

Lukasa:

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3589,6 +3589,17 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
         let response = try client.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
         XCTAssertEqual(.ok, response.status)
     }
+
```
Lukasa, line 3592: The soundness script doesn't like the whitespace here

```suggestion

```

Lukasa:
Ok, looks like the linux tests fail with `EINVAL`. This happens with kernels that don't support MPTCP, so you'll need to add a try-catch that checks for this case and skips the test in that context.

Aperence:
Just added the do-catch now !

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3590,6 +3590,21 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
         XCTAssertEqual(.ok, response.status)
     }
 
+    func testClientWithMultipath() throws {
+        do{
```
Lukasa, line 3594:
```suggestion
        do {
```

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3590,6 +3590,21 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+            var conf = HTTPClient.Configuration()
+            conf.enableMultipath = true
+            let client = HTTPClient(configuration: conf)
+            defer {
+                XCTAssertNoThrow(try client.shutdown().wait())
+            }
+            let response = try client.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
+            XCTAssertEqual(.ok, response.status)
+        }catch{
```
Lukasa, line 3603: We need to catch the specific error, as other errors should still fail the test. This should be [`IOError`](https://swiftpackageindex.com/apple/swift-nio/2.72.0/documentation/niocore/ioerror), with `errnoCode == EINVAL`. Please also `throw XCTSkip()` in the catch as well.

Aperence:
Applied required changes

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3590,6 +3590,22 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+        do {
+            var conf = HTTPClient.Configuration()
+            conf.enableMultipath = true
+            let client = HTTPClient(configuration: conf)
+            defer {
+                XCTAssertNoThrow(try client.shutdown().wait())
+            }
+            let response = try client.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
+            XCTAssertEqual(.ok, response.status)
+        }catch let error as IOError where error.errnoCode == EINVAL{
```
Lukasa, line 3603: Quick cleanup:

```suggestion
        } catch let error as IOError where error.errnoCode == EINVAL {
```

Lukasa:

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -3590,6 +3590,22 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+        do {
+            var conf = HTTPClient.Configuration()
+            conf.enableMultipath = true
+            let client = HTTPClient(configuration: conf)
+            defer {
+                XCTAssertNoThrow(try client.shutdown().wait())
+            }
+            let response = try client.get(url: self.defaultHTTPBinURLPrefix + "get").wait()
+            XCTAssertEqual(.ok, response.status)
+        } catch let error as IOError where error.errnoCode == EINVAL {
```
Lukasa, line 3603: Just tried running this locally, it turns out `EPROTONOSUPPORT` is also sometimes thrown on Linux, so we need to add that to the list.

Aperence: Added this, I've checked the [documentation](https://www.mptcp.dev/implementation.html) and it seems that it could also raise an `ENOPROTOOPT` error, so I've added this one too to the list

Lukasa:

Lukasa (approved):
Nice one, thanks!

PR #768, opened 2024-09-03
Fix crash when writablity becomes false and races against finishing the http request

fabianfett:
### Motivation

If the channel's writability changed to false just before we finished a request, we currently run into a precondition.

### Changes

- Remove the precondition and handle the case appropiatly

### Result

A crash less.


weissi (approved):
Amazing, thank you!

FranzBusch (approved):

Issue #754, opened 2024-07-23
7% of the overall runtime spent in `SelectableEventLoop.debugDescription`


weissi:
Just noticed that when just running loads of sequential super basic HTTP requests against a fast server (`NIOHTTP1Server`), we spent about 7% of the overall runtime in  `HTTP1ClientChannelHandler.request.setter` which spends most of its time in `_print_unlocked<A, B>(_:_:)` because it's pulling a `debugDescription` for EventLoop for the logger metadata :P (release mode compile of course)

Should be an easy fix.

https://github.com/swift-server/async-http-client/blob/54d1006dc90b9ee77b4d04d63ad8688d2215dc1e/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift#L39

<img width="1248" alt="Screenshot 2024-07-23 at 11 55 05 am" src="https://github.com/user-attachments/assets/a36030ed-39e8-4c8c-ab5e-5767ded511b3">

FWIW, the super basic perf test I used is this: https://github.com/weissi/async-http-client/commit/bf7dbeb933433bc111942808ec5ed9108c44be72 (this code isn't meant to be merged).

Lukasa:
Should have been resolved by #772.

PR #772, opened 2024-09-17
Reduce time spent logging EventLoop description in HTTP1ClientChannelHandler

aryan-25:
### Motivation:

A performance test executing 100,000 sequential requests against a simple [`NIOHTTP1Server`](https://github.com/apple/swift-nio/blob/main/Sources/NIOHTTP1Server/README.md) revealed that 7% of total run time is spent in the setter of the `request` property in `HTTP1ClientChannelHandler` (GitHub Issue #754).

The poor performance comes from [processing the string interpolation `"\(self.eventLoop)"`](https://github.com/swift-server/async-http-client/blob/6df8e1c17e68f0f93de2443b8c8cafca9ddcc89a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift#L39C17-L39C75) which under the hood calls a computed property.

This problem can entirely be avoided by storing `eventLoop.description` when initializing `HTTP1ClientChannelHandler`, and using that stored value in `request`'s setter, rather than computing the property each time.

### Modifications:

- Created a new property `let eventLoopDescription: Logger.MetadataValue` in `HTTP1ClientChannelHandler` that stores the description of the `eventLoop` argument that is passed into the initializer.
- Replaced the string interpolation `"\(self.eventLoop)"` in `request`'s setter with `self.eventLoopDescription`.

### Result:

`HTTP1ClientChannelHandler.eventLoop`'s `description` property is cached upon initialization rather than being computed each time in the `request` property's setter.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
@@ -36,7 +36,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             if let newRequest = self.request {
                 var requestLogger = newRequest.logger
                 requestLogger[metadataKey: "ahc-connection-id"] = self.connectionIdLoggerMetadata
-                requestLogger[metadataKey: "ahc-el"] = "\(self.eventLoop)"
+                requestLogger[metadataKey: "ahc-el"] = self.eventLoopDescription
                 self.logger = requestLogger
 
                 if let idleReadTimeout = newRequest.requestOptions.idleReadTimeout {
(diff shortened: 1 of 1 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -72,11 +72,13 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
     private let backgroundLogger: Logger
     private var logger: Logger
     private let eventLoop: EventLoop
+    private let eventLoopDescription: Logger.MetadataValue
     private let connectionIdLoggerMetadata: Logger.MetadataValue
 
     var onConnectionIdle: () -> Void = {}
     init(eventLoop: EventLoop, backgroundLogger: Logger, connectionIdLoggerMetadata: Logger.MetadataValue) {
         self.eventLoop = eventLoop
+        self.eventLoopDescription = Logger.MetadataValue(stringLiteral: eventLoop.description)
```
Lukasa, line 81: This is not a string literal, so we shouldn't use this. Let's use the string interpolation instead.

Lukasa (approved):

PR #777, opened 2024-10-02
workaround Foundation.URL behavior changes

rnro:
`Foundation.URL` has various behavior changes in Swift 6 to better match RFC 3986 which impact AHC.

In particular it now no longer strips the square brackets in IPv6 hosts which are not tolerated by `inet_pton` so these must be manually stripped.

https://github.com/swiftlang/swift-foundation/issues/957
https://github.com/swiftlang/swift-foundation/issues/958
https://github.com/swiftlang/swift-foundation/issues/962

### DIFF
--- a/Sources/AsyncHTTPClient/DeconstructedURL.swift
+++ b/Sources/AsyncHTTPClient/DeconstructedURL.swift
@@ -48,9 +48,16 @@ extension DeconstructedURL {
 
         switch scheme {
         case .http, .https:
+            #if !canImport(Darwin) && compiler(>=6.0)
+            guard let urlHost = url.host, !urlHost.isEmpty else {
+                throw HTTPClientError.emptyHost
+            }
+            let host = urlHost.trimIPv6Brackets()
+            #else
             guard let host = url.host, !host.isEmpty else {
                 throw HTTPClientError.emptyHost
             }
+            #endif
             self.init(
                 scheme: scheme,
                 connectionTarget: .init(remoteHost: host, port: url.port ?? scheme.defaultPort),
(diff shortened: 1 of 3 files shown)


Sources/AsyncHTTPClient/DeconstructedURL.swift
```diff
@@ -48,9 +48,16 @@ extension DeconstructedURL {
 
         switch scheme {
         case .http, .https:
+            #if os(Linux) && compiler(>=6.0)
```
Lukasa, line 51: This is the wrong check: it's `if !canImport(Darwin)`

Sources/AsyncHTTPClient/DeconstructedURL.swift
```diff
@@ -48,9 +48,16 @@ extension DeconstructedURL {
...
+            guard var urlHost = url.host, !urlHost.isEmpty else {
```
Lukasa, line 52:
```suggestion
            guard let urlHost = url.host, !urlHost.isEmpty else {
```

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -43,8 +43,12 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
         XCTAssertEqual(request2.url.path, "")
 
         let request3 = try Request(url: "unix:///tmp/file")
-        XCTAssertNil(request3.url.host)
         XCTAssertEqual(request3.host, "")
+        #if os(Linux) && compiler(>=6.0)
```
Lukasa, line 47: Same note here.

Lukasa (approved):

PR #778, opened 2024-10-11
Added: ability to set basic authentication on requests

agamdua:
Motivation:

As an HTTP library, async-http-client should have authentication support.

Modifications:

This adds a `setBasicAuth()` method to both HTTPClientRequest and `HTTPClient.Request` and their related unit tests.

Result:

Library users will be able to leverage this method to use basic authentication on their requests without implementing this in their own projects.

Note: I also ran the tests (`swift test`) with the `docker.io/library/swift:6.0-focal` and `docker.io/library/swift:5.10.1-focal` to ensure linux compatibility.

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+auth.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+auth.swift
@@ -1,0 +1,27 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2024 Apple Inc. and the AsyncHTTPClient project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientRequest {
+    /// Set basic auth for a request.
+    ///
+    /// - parameters:
+    ///     - username: the username to authenticate with
+    ///     - password: authentication password associated with the username
+    public mutating func setBasicAuth(username: String, password: String) {
+        self.headers.setBasicAuth(username: username, password: password)
+    }
+}
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -285,6 +285,15 @@ extension HTTPClient {
 
             return (head, metadata)
         }
+
+        /// Set basic auth for a request.
+        ///
+        /// - parameters:
+        ///     - username: the username to authenticate with
+        ///     - password: authentication password associated with the username
+        public mutating func setBasicAuth(username: String, password: String) {
+            self.headers.setBasicAuth(username: username, password: password)
+        }
     }
 
     /// Represents an HTTP response.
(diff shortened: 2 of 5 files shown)


Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+auth.swift
```diff
@@ -0,0 +1,29 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2022 Apple Inc. and the AsyncHTTPClient project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2024 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+auth.swift
```diff
@@ -0,0 +1,29 @@
...
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension HTTPClientRequest {
+    /// Set basic auth for a request.
+    ///
+    /// - parameters:
+    ///     - username: the username to authenticate with
+    ///     - password: authentication password associated with the username
+    mutating public func setBasicAuth(username: String, password: String) {
+        let value = Data(String(format: "%@:%@", username, password).utf8)
```
Lukasa, line 25: I think we can construct this a bit more cheaply. Something like this:

```swift
var value = Data()
value.reserveCapacity(username.utf8.count + password.utf8.count + 1)
value.append(contentsOf: username.utf8)
value.append(UInt8(ascii: ":"))
value.append(contentsOf: password.utf8)
```
```

File: Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+auth.swift [hunk 3]
```diff
@@ -0,0 +1,29 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of AsyncHTTPClient project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```
Lukasa, line 17: Where is this availability annotation coming from?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -285,6 +285,18 @@ extension HTTPClient {
...
+
+        /// Set basic auth for a request.
+        ///
+        /// - parameters:
+        ///     - username: the username to authenticate with
+        ///     - password: authentication password associated with the username
+        mutating public func setBasicAuth(username: String, password: String) {
+            let value = Data(String(format: "%@:%@", username, password).utf8)
+            let encoded = value.base64EncodedString()
+            self.headers.replaceOrAdd(name: "Authorization", value: "Basic \(encoded)")
```
Lukasa, line 297: Let's refactor out the common code here.

Lukasa (commented):
Thanks for this @agamdua! I've left a few notes in the diff. I'm broadly happy with this, but I'd like to see if @FranzBusch thinks the API is suitable for now.

Lukasa:

agamdua:
Sounds good - thank you both! Will address comments soon, they all make sense 👍

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+auth.swift
agamdua: This is from the original type: https://github.com/swift-server/async-http-client/blob/main/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift#L39-L40

Sources/AsyncHTTPClient/HTTPHandler.swift
agamdua: I created a new file and function for the code here, and extended `HTTPHeaders` with internal access to reuse the code effectively. Let me know if you prefer different code organization!

Lukasa (approved):
Nice one, thanks @agamdua!

PR #780, opened 2024-10-28
Adopt GitHub actions

rnro:
Migrate CI to use GitHub Actions.

### Motivation:

To migrate to GitHub actions and centralised infrastructure.

### Modifications:

Changes of note:
* Adopt swift-format using rules from SwiftNIO.
* Remove scripts and docker files which are no longer needed.
* Disabled warnings-as-errors on Swift 6.0 CI pipelines for now.

### Result:

Feature parity with old CI.


.unacceptablelanguageignore
```diff
@@ -0,0 +1 @@
+NOTICE.txt
```
FranzBusch, line 1: If this is flagged due to the `master` we can just switch that to a commit sha instead

FranzBusch (approved):

PR #785, opened 2024-11-22
fix 784: dont crash on huge in-memory bodies

### LINKED ISSUE #784: AHC crashes with HTTP2 when uploading a big body when channel & delegate live on separate EventLoops
This is a pretty fun bug: In (at least) HTTP2 mode, when uploading a huge body, AHC behaves badly in various different ways.
The writes (in a recursive loop) are run in a recursive loop and they are freewheeling meaning there is no real backpressure.

Specifically in https://github.com/swift-server/async-http-client/blob/bdaa3b18358b60e268afe82e8feb16418874f41b/Sources/AsyncHTTPClient/RequestBag.swift#L187-L188 , we _don't_ return the actual `write` promise but rather a pre-succeeded future. This won't stop until the state machine switches into a mode to "pause" sending the request body.

The problem is twofold:
1. When delegate & channel are _not_ co-located, the "pause" can arrive quite a bit too late because it's triggered from the channel's `channelWritabilityChanged -> false` which of course happens on the channel's EL. But the writes are triggered on the delegate's EL... (That's bad usually not fatal, we're just eating up some extra memory)
2. Unfortunately, (in https://github.com/swift-server/async-http-client/blob/bdaa3b18358b60e268afe82e8feb16418874f41b/Sources/AsyncHTTPClient/HTTPHandler.swift#L58-L69) we are actually triggering the writes recursively! That means it's actually _impossible_ to get the pause signal until we're done. We'll just recurse on that pre-fulfilled future (which will immediately execute) and never enter the event loop again where a "please pause sending bytes" is waiting to execute. This is _always_ fatal if we can't tolerate `O(bodySize / 4MB)` recursive stack frames. 5 GB always does the job.

weissi:
fixes #784


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -49,23 +49,52 @@ extension HTTPClient {
...
+                    if let nextElement = iterator.withLockedValue({ $0.next() }) {
+                        let (index, chunk) = nextElement
+                        self.write(.byteBuffer(ByteBuffer(bytes: chunk))).map {
+                            if (index + 1) % 4 == 0 {
+                                // Let's not stack-overflow if the futures insta-complete which they at least in HTTP/2
+                                // mode.
+                                // Also, we must frequently return to the EventLoop because we may get the pause signal
+                                // from another thread. If we fail to do that promptly, we may balloon our body chunks
+                                // into memory.
+                                allDone.futureResult.eventLoop.execute {
```
weissi, line 80: This is the crucial line (apart from the thread unsafety from `UnsafeMutableTransferBox`)

Tests/AsyncHTTPClientTests/HTTPClientTestUtils.swift
```diff
@@ -1548,3 +1586,88 @@ private let key = """
     oYQsPj00S3/GA9WDapwe81Wl2A==
     -----END PRIVATE KEY-----
     """
+
+final class BasicInboundTrafficShapingHandler: ChannelDuplexHandler {
```
weissi, line 1590: This turned out to not be necessary, but it's a test-only utility which can't hurt I thought :).

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -49,23 +49,52 @@ extension HTTPClient {
...
+                // - https://github.com/swift-server/async-http-client/issues/194
+                // - https://github.com/swift-server/async-http-client/issues/264
+                // - We're not told the EventLoop the task runs on and the user is free to return whatever EL they
+                //   want.
+                // One important consideration then is that we must lock around the iterator because we could be hopping
+                // between threads.
+                let iterator = NIOLockedValueBox(
+                    bytes.chunks(ofCount: maxChunkSize).enumerated().makeIterator()
+                )
+                guard let chunk = iterator.withLockedValue({ $0.next() }) else {
```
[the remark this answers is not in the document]
weissi: ah good point, will do

Lukasa (approved):

PR #798, opened 2024-12-27
import FoundationEssentials instead of Foundation when available

t089:
This PR replaces `import Foundation` with `import FoundationEssentials` when available.

Linking `Foundation` on Linux comes with a significant binary size increase due to the included icu data. None of this is necessary for the HTTP client. Instead we can use `FoundationEssentials` to get access to the essential types like `URL`, `Data`, `Date`, ... without including icu data.

According to @Lukasa, sadly, this might be a semver major change. :(


Lukasa:
Indeed it is. See apple/swift-crypto#317 for the full discussion.

Issue #596, opened 2022-06-16
Add functionality to add "Debug handlers" to the channel pipelines


Lukasa:
A useful trick we learned in grpc-swift is that it can be handy to allow users to insert a "debug handler" (or several) into each channel pipeline. This should usually be given the option to insert itself anywhere in the pipeline (meaning we should probably name our handlers), and can enable all kinds of useful observability and testing. We may need to be careful for HTTP/2, and we should consider @glbrntt's grpc experience here.

weissi:
Yes please, I use that functionality pretty much on every gRPC server/client I've ever created :).

PR #801, opened 2025-01-21
Add "debug initializer" hook for channels

clintonpi:
Motivation:

As requested in #596, it can be handy to have a lower-level access to channels (HTTP/1 connection, HTTP/2 connection, or HTTP/2 stream) to enable a more fine-grained interaction for, say, observability, testing, etc.

Modifications:

- Add 3 new properties (`http1_1ConnectionDebugInitializer`, `http2ConnectionDebugInitializer` and `http2StreamChannelDebugInitializer`) to `HTTPClient.Configuration` with access to the respective channels. These properties are of `Optional` type `@Sendable (Channel) -> EventLoopFuture<Void>` and are called when creating a connection/stream.

Result:

Provides APIs for a lower-level access to channels.


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -847,14 +847,32 @@ public class HTTPClient {
...
             proxy: Proxy? = nil,
             ignoreUncleanSSLShutdown: Bool = false,
-            decompression: Decompression = .disabled
+            decompression: Decompression = .disabled,
+            http1_1ConnectionDebugInitializer:
+                (@Sendable (Channel) -> EventLoopFuture<Void>)? = nil,
+            http2ConnectionDebugInitializer:
+                (@Sendable (Channel) -> EventLoopFuture<Void>)? = nil,
+            http2StreamChannelDebugInitializer:
+                (@Sendable (Channel) -> EventLoopFuture<Void>)? = nil
```
glbrntt, line 875: Adding parameters to a function is a SemVer breaking change, even if they have defaults.

The reason is that you could store the function in a variable:

```swift
struct Foo {
  var bar: Int
  init(bar: Int) { ... }
}

// This is totally value
let makeFoo = Foo.init(bar:)
let foo = makeFoo(bar: 42) // equivalent to Foo(bar: 42)
```

If you subsequently add a parameter to the `init` then this is an API break because `init(bar:)` won't exist, _even_ if there's a default for the param you add.


The upshot of this is that you should add a new `init` rather than updating the existing ones.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -4306,4 +4306,89 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+        // The stream channel debug initializer must be called only as much as the number of
+        // requests made.
+        XCTAssertEqual(streamChannelDebugInitializerUtil.executionCount, numberOfRequests)
+    }
+}
+
+class DebugInitializerUtil {
+    var executionCount: Int
+
+    @Sendable
```
glbrntt, line 4384: You've fallen into the trap where you've done what the compiler told you to but the compiler was wrong.

This is a class with mutable state and no synchronisation mechanism on that state: `executionCount` is modified from an event loop and subsequently read from the test thread.

Generally speaking don't use `@Sendable` on methods, it's rarely correct.

In this instance the utility should be `Sendable` and access to the `executionCount` protected by a lock (we can use a `NIOLockedValueBox`).

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -4306,4 +4306,89 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+        // only once.
+        XCTAssertEqual(connectionDebugInitializerUtil.executionCount, 1)
+
+        // The stream channel debug initializer must be called only as much as the number of
+        // requests made.
+        XCTAssertEqual(streamChannelDebugInitializerUtil.executionCount, numberOfRequests)
+    }
+}
+
+class DebugInitializerUtil {
```
glbrntt, line 4381: This isn't a very helpful name, something like `CountingDebugInitializer` would be more appropriate as it hints to what it actually does.

glbrntt (changes requested):
Good start, I left some feedback inline.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
clintonpi: True! Easy to miss!

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -4306,4 +4306,108 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+        // requests made.
+        XCTAssertEqual(
+            streamChannelDebugInitializerUtil.executionCount.withLockedValue { $0 },
+            numberOfRequests
+        )
+    }
+}
+
+final class CountingDebugInitializerUtil: Sendable {
+    let executionCount: NIOLockedValueBox<Int>
```
glbrntt, line 4400: A useful thing to do is to make this private and have a computed property which returns the value from the box; this makes it easier for any test which wants the value to get it without having to go via `withLockedValue`.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -4306,4 +4306,108 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+
+    /// The acual debug initializer.
+    @Sendable
```
glbrntt, line 4403: I don't think we need this annotation as the type is `Sendable`

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -4306,4 +4306,108 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+            eventLoopGroupProvider: .singleton,
+            configuration: config,
+            backgroundActivityLogger: Logger(
+                label: "HTTPClient",
+                factory: StreamLogHandler.standardOutput(label:)
+            )
+        )
+        defer { XCTAssertNoThrow(client.shutdown()) }
+
+        let bin = HTTPBin(.http1_1(ssl: true, compress: false))
```
glbrntt, line 4334: Presumably the initialiser also works when we're not using SSL? I'm not super familiar with this codebase so would be good to check that.

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
glbrntt: I think that's because you do `http1_1ConnectionDebugInitializer: connectionDebugInitializerUtil.initialize(channel:)` rather than:

```swift
http1_1ConnectionDebugInitializer: { channel in 
   connectionDebugInitializerUtil.initialize(channel: channel) 
}
```

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -4306,4 +4306,119 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+        // requests made.
+        XCTAssertEqual(streamChannelDebugInitializerUtil.executionCount, numberOfRequests)
+    }
+}
+
+final class CountingDebugInitializerUtil: Sendable {
+    private let _executionCount: NIOLockedValueBox<Int>
+    var executionCount: Int { self._executionCount.withLockedValue { $0 } }
+
+    /// The acual debug initializer.
```
Lukasa, line 4414:
```suggestion
    /// The actual debug initializer.
```

Tests/AsyncHTTPClientTests/HTTPClientTests.swift
```diff
@@ -4306,4 +4306,119 @@ final class HTTPClientTests: XCTestCaseHTTPClientTestsBaseClass {
...
+    func initialize(channel: Channel) -> EventLoopFuture<Void> {
+        self._executionCount.withLockedValue { $0 += 1 }
+
+        return channel.eventLoop.makeSucceededVoidFuture()
```
Lukasa, line 4418: It may be useful to add some tests that have the debug initializer _actually delay_ the setup. That is, make sure that we actually do wait for this to complete. Easiest thing to do might be to set some low timeouts, and then confirm the timeouts fire, and that completing the promise doesn't then cause any issues.

Lukasa (commented):
This looks really nice, I've left a couple of suggestions for the tests.

Lukasa (approved):
Nice one, thanks @clintonpi!

glbrntt (approved):

glbrntt:
Thanks @clintonpi -- this'll be a really handy feature!

PR #803, opened 2025-01-28
Avoid precondition failure in write timeout

rnro:
### Motivation:

In some cases we can crash because of a precondition failure when the write timeout fires and we aren't in the running state. This can happen for example if the connection is closed whilst the write timer is active.

### Modifications:

* Remove the precondition and instead take no action if the timeout fires outside of the running state. Instead we take a new `Action`, `.noAction` when the timer fires.
* Clear write timeouts upon request completion. When a request completes we have no use for the idle write timer, we clear the read timer and we should clear the write one too.

### Result:

Fewer crashes.

The supplied tests fails without these changes and passes with either of them.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
@@ -314,6 +314,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             let oldRequest = self.request!
             self.request = nil
             self.runTimeoutAction(.clearIdleReadTimeoutTimer, context: context)
+            self.runTimeoutAction(.clearIdleWriteTimeoutTimer, context: context)
 
             switch finalAction {
             case .close:
@@ -353,6 +354,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
             let oldRequest = self.request!
             self.request = nil
             self.runTimeoutAction(.clearIdleReadTimeoutTimer, context: context)
+            self.runTimeoutAction(.clearIdleWriteTimeoutTimer, context: context)
 
             switch finalAction {
             case .close(let writePromise):
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
@@ -281,7 +281,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
         case .close:
             context.close(promise: nil)
 
-        case .wait:
+        case .wait, .noAction:
             break
 
         case .forwardResponseHead(let head, let pauseRequestBodyStream):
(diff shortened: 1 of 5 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ClientChannelHandler.swift
```diff
@@ -281,7 +281,7 @@ final class HTTP1ClientChannelHandler: ChannelDuplexHandler {
         case .close:
             context.close(promise: nil)
 
-        case .wait:
+        case .wait, .noAction:
```
rnro, line 284: Do we need this new state or can we just re-use `.wait`?

fabianfett: reuse `.wait`. `.wait` is a do nothing. can be renamed in a follow up pr.

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ConnectionStateMachine.swift
```diff
@@ -359,7 +361,7 @@ struct HTTP1ConnectionStateMachine {
 
     mutating func idleWriteTimeoutTriggered() -> Action {
         guard case .inRequest(var requestStateMachine, let close) = self.state else {
-            preconditionFailure("Invalid state: \(self.state)")
+            return .noAction
```
fabianfett, line 364: Do you mind adding another test to HTTP1ConnectionStateMachineTests that just tests this edge?

Sources/AsyncHTTPClient/ConnectionPool/HTTP1/HTTP1ConnectionStateMachine.swift
```diff
@@ -83,6 +83,8 @@ struct HTTP1ConnectionStateMachine {
         case fireChannelActive
         case fireChannelInactive
         case fireChannelError(Error, closeConnection: Bool)
+
+        case noAction
```
fabianfett, line 87: please remove.

Tests/AsyncHTTPClientTests/HTTP1ClientChannelHandlerTests.swift
```diff
@@ -840,6 +840,81 @@ class HTTP1ClientChannelHandlerTests: XCTestCase {
         channel.writeAndFlush(request, promise: nil)
         XCTAssertEqual(request.events.map(\.kind), [.willExecuteRequest, .requestHeadSent])
     }
+
+    class SlowHandler: ChannelOutboundHandler {
```
fabianfett, line 844: this isn't used at all.

Tests/AsyncHTTPClientTests/HTTP1ClientChannelHandlerTests.swift
```diff
@@ -840,6 +840,81 @@ class HTTP1ClientChannelHandlerTests: XCTestCase {
...
+
+        // start a request stream we'll never write to
+        let streamPromise = embedded.eventLoop.makePromise(of: Void.self)
+        let streamCallback = { @Sendable (streamWriter: HTTPClient.Body.StreamWriter) -> EventLoopFuture<Void> in
+            streamPromise.futureResult
+        }
+        request.body = .init(contentLength: nil, stream: streamCallback)
+
+        let delegate = NullResponseDelegate()
+        var maybeRequestBag: RequestBag<NullResponseDelegate>?
```
fabianfett, line 874: Why not use a `ResponseAccumulator` here instead?

fabianfett (commented):
Nits.

Tests/AsyncHTTPClientTests/HTTP1ClientChannelHandlerTests.swift
rnro: I hadn't spotted that, thanks.

PR #806, opened 2025-02-05
baby steps towards a Structured Concurrency API

weissi:
At the moment, `HTTPClient`'s entire API surface violates Structured Concurrency. Both the creation & shutdown of a HTTP client as well as making requests doesn't follow Structured Concurrency. Some of the problems are:

1. Upon return of methods, resources are still in active use in other threads/tasks
2. Cancellation doesn't always work

This PR is baby steps towards a Structured Concurrency API, starting with start/shutdown of the HTTP client.


Lukasa (commented):
I'm happy with this, but would like @fabianfett to review as well.

Sources/AsyncHTTPClient/StructuredConcurrencyHelpers.swift
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@inlinable
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+internal func asyncDo<R: Sendable>(
+    _ body: @escaping @Sendable () async throws -> R,
+    finally: @escaping @Sendable ((any Error)?) async throws -> Void
+) async throws -> R {
```
FranzBusch, line 20:
```suggestion
internal func asyncDo<Result>(
    _ body: () async throws -> sending R,
    finally: @escaping sending((any Error)?) async throws -> Void
) async throws -> sending Result {
```

Sources/AsyncHTTPClient/HTTPClient+StructuredConcurrency.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// This method allows to start & automatically dispose of a ``HTTPClient`` following the principle of Structured Concurrency.
+    /// The ``HTTPClient`` is guaranteed to be shut down upon return, whether `body` throws or not.
+    ///
+    /// This may be particularly useful if you cannot use the shared singleton (``HTTPClient/shared``).
+    public static func withHTTPClient<R: Sendable>(
+        eventLoopGroup: any EventLoopGroup = HTTPClient.defaultEventLoopGroup,
+        configuration: Configuration = Configuration(),
+        backgroundActivityLogger: Logger? = nil,
+        _ body: @escaping @Sendable (HTTPClient) async throws -> R
+    ) async throws -> R {
```
FranzBusch, line 17:
```suggestion
    public static func withHTTPClient<Result>(
        eventLoopGroup: any EventLoopGroup = HTTPClient.defaultEventLoopGroup,
        configuration: Configuration = Configuration(),
        backgroundActivityLogger: Logger? = nil,
        isolation: isolated (any Actor)? = #isolation,
        _ body: (HTTPClient) async throws -> sending Result
    ) async throws -> sending Result {
```

FranzBusch:
I left some Sendable correctness remarks. Overall this looks good to me.

Sources/AsyncHTTPClient/HTTPClient+StructuredConcurrency.swift
weissi: @FranzBusch isn't this all Swift 6.0? We support 5.9, right?

FranzBusch: Partially the `@Sendable @escsaping` isn't generally needed on the body. We need those APIs `isolation/sending` things for `6.0` at least otherwise adopters will get warnings/errors.

weissi: @FranzBusch So what's the recommendation? Add two, one for 5.9 and one for 6.0 using `#if compiler`?

weissi: @FranzBusch Okay, that's weird. So the API actually changes depending on the compiler version you use to compile...

weissi: > Without the #isolation you will get errors in Swift 6.

Well. in certain scenarios, yes. All my stuff is 6 language mode and never got an error. For our software we keep it 5.10 compatible at the minute and migrate to `isolated` once we're 6.0+

Sources/AsyncHTTPClient/HTTPClient+StructuredConcurrency.swift
```diff
@@ -0,0 +1,58 @@
...
+    /// The ``HTTPClient`` is guaranteed to be shut down upon return, whether `body` throws or not.
+    ///
+    /// This may be particularly useful if you cannot use the shared singleton (``HTTPClient/shared``).
+    public static func withHTTPClient<R>(
+        eventLoopGroup: any EventLoopGroup = HTTPClient.defaultEventLoopGroup,
+        configuration: Configuration = Configuration(),
+        backgroundActivityLogger: Logger? = nil,
+        isolation: isolated (any Actor)? = #isolation,
+        _ body: sending (HTTPClient) async throws -> sending R
+    ) async throws -> sending R {
```
FranzBusch, line 19: Can we please use a more descriptive generic type name than `R`. It shows up in docs and autocomplete so it's worth making this better. Also the `body` must not be `sending`. `asyncDo` is not `sending` it anyways. This should run in the same task isolation as the caller of the `withHTTPClient` method

```suggestion
    public static func withHTTPClient<Result>(
        eventLoopGroup: any EventLoopGroup = HTTPClient.defaultEventLoopGroup,
        configuration: Configuration = Configuration(),
        backgroundActivityLogger: Logger? = nil,
        isolation: isolated (any Actor)? = #isolation,
        _ body: (HTTPClient) async throws -> sending Result
    ) async throws -> sending Result {
```

Sources/AsyncHTTPClient/HTTPClient+StructuredConcurrency.swift
```diff
@@ -0,0 +1,58 @@
...
+    /// This method allows to start & automatically dispose of a ``HTTPClient`` following the principle of Structured Concurrency.
+    /// The ``HTTPClient`` is guaranteed to be shut down upon return, whether `body` throws or not.
+    ///
+    /// This may be particularly useful if you cannot use the shared singleton (``HTTPClient/shared``).
+    public static func withHTTPClient<R: Sendable>(
+        eventLoopGroup: any EventLoopGroup = HTTPClient.defaultEventLoopGroup,
+        configuration: Configuration = Configuration(),
+        backgroundActivityLogger: Logger? = nil,
+        _ body: @escaping @Sendable (HTTPClient) async throws -> R
+    ) async throws -> R {
```
FranzBusch, line 44:
```suggestion
    public static func withHTTPClient<Result: Sendable>(
        eventLoopGroup: any EventLoopGroup = HTTPClient.defaultEventLoopGroup,
        configuration: Configuration = Configuration(),
        backgroundActivityLogger: Logger? = nil,
        _ body: (HTTPClient) async throws -> Result
    ) async throws -> Result {
```

Sources/AsyncHTTPClient/StructuredConcurrencyHelpers.swift
```diff
@@ -0,0 +1,60 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=6.0)
+@inlinable
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+internal func asyncDo<R>(
+    isolation: isolated (any Actor)? = #isolation,
+    _ body: sending () async throws -> sending R,
```
FranzBusch, line 20:
```suggestion
    _ body: () async throws -> sending R,
```

Sources/AsyncHTTPClient/StructuredConcurrencyHelpers.swift
```diff
@@ -0,0 +1,60 @@
...
+    } catch {
+        // This _looks_ unstructured but isn't really because we unconditionally always await the return.
+        // We need to have an uncancelled task here to assure this is actually running in case we hit a
+        // cancellation error.
+        try await Task {
+            try await finally(error)
+        }.value
+        throw error
+    }
+    try await finally(nil)
```
FranzBusch, line 35: Why doesn't this `finally` need a cancellation shield? It might still be the case that we cancel right between the `body` returning and us calling `finally`

Sources/AsyncHTTPClient/StructuredConcurrencyHelpers.swift
```diff
@@ -0,0 +1,60 @@
...
+    return result
+}
+#else
+@inlinable
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+internal func asyncDo<R: Sendable>(
+    _ body: @escaping @Sendable () async throws -> R,
```
FranzBusch, line 42:
```suggestion
    _ body: () async throws -> R,
```

Sources/AsyncHTTPClient/HTTPClient+StructuredConcurrency.swift
weissi: That's in the 5.9 part

Sources/AsyncHTTPClient/StructuredConcurrencyHelpers.swift
FranzBusch: Shouldn't matter. It still is not cross an isolation domain nor is it escaping.

Sources/AsyncHTTPClient/StructuredConcurrencyHelpers.swift
weissi: Yes, this is old code I had accidentally pushed.

Sources/AsyncHTTPClient/StructuredConcurrencyHelpers.swift
weissi: Alright, I tightened them all down to what I think is the minimum in the current implementation. Feel slightly uneasy about tightening it so much as it might lock us in. But I'm okay going with what you use as the standard shape.

FranzBusch (approved):

weissi:
Thanks @FranzBusch for advice here! As you know, I still use `Sendable` everything -- but yeah for public SemVer'd APIs that is probably an issue, particularly with `@MainActor` that is of course used heavily on Apple platforms.

PR #811, opened 2025-02-13
Add `head` property to `FileDownloadDelegate`'s `Progress`/`Response` struct

gregcotten:
I needed a way to use a `FileDownloadDelegate` task to fish out the recommended file name.
```swift
let response = try await downloadTask.get()

// access content-disposition
response.head.headers.first(name: "Content-Disposition")
```

The `head` property is an explicitly unwrapped optional because there is no "default value" to set it to, and it won't be accessed by the user until it's already been set anyway. This is a little inelegant, so I could change it to something like below where I fill in bogus init data, but that seems worse for some reason.
```swift
public struct Progress: Sendable {
    public var totalBytes: Int?
    public var receivedBytes: Int
    public var head: HTTPResponseHead
}

private var progress = Progress(
    totalBytes: nil,
    receivedBytes: 0,
    head: .init(
        version: .init(major: 0, minor: 0),
        status: .badRequest
    )
)
```

### DIFF
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -23,10 +23,10 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
     public struct Progress: Sendable {
         public var totalBytes: Int?
         public var receivedBytes: Int
-        public var responseHead: HTTPResponseHead!
+        public var head: HTTPResponseHead!
     }
 
-    private var progress = Progress(totalBytes: nil, receivedBytes: 0, responseHead: nil)
+    private var progress = Progress(totalBytes: nil, receivedBytes: 0, head: nil)
 
     public typealias Response = Progress
 
@@ -136,7 +136,7 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
     ) -> EventLoopFuture<Void> {
         self.reportHead?(task, head)
 
-        self.progress.responseHead = head
+        self.progress.head = head
 
         if let totalBytesString = head.headers.first(name: "Content-Length"),
             let totalBytes = Int(totalBytesString)
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -19,14 +19,22 @@ import NIOPosix
 /// Handles a streaming download to a given file path, allowing headers and progress to be reported.
 public final class FileDownloadDelegate: HTTPClientResponseDelegate {
     /// The response type for this delegate: the total count of bytes as reported by the response
-    /// "Content-Length" header (if available) and the count of bytes downloaded.
+    /// "Content-Length" header (if available), the count of bytes downloaded, and the
+    /// response head.
     public struct Progress: Sendable {
         public var totalBytes: Int?
         public var receivedBytes: Int
-        public var head: HTTPResponseHead!
+        public var head: HTTPResponseHead
     }
 
-    private var progress = Progress(totalBytes: nil, receivedBytes: 0, head: nil)
+    private var progress = Progress(
+        totalBytes: nil,
+        receivedBytes: 0,
+        head: .init(
+            version: .init(major: 0, minor: 0),
+            status: .init(statusCode: 0)
+        )
+    )
 
     public typealias Response = Progress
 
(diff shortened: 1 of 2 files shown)


Lukasa:
Hi @gregcotten, thanks for opening this and sorry for the slow time to review, it's been a busy week. I'll take a look now.

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -23,9 +23,10 @@ public final class FileDownloadDelegate: HTTPClientResponseDelegate {
     public struct Progress: Sendable {
         public var totalBytes: Int?
         public var receivedBytes: Int
+        public var head: HTTPResponseHead!
```
Lukasa, line 26: Let's avoid the IUO here. I agree with your rationale that it's not a good idea for us to set this to a default value, but I think it's fine to leave it as an explicit optional and just set it to `nil`. This is the same behaviour used for `totalBytes`.

Lukasa (commented):
Great, left a note in the diff. Can we also add at least one unit test?

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
gregcotten: @Lukasa

It's just so unfortunate ergonomics-wise because it means user code is going to have to explicitly unwrap (feels bad) or write their own "throw error if head is nil" code even though it will guaranteed to never be `nil` in the `reportProgress` handler or in the result `Response`. I'm almost leaning toward init'ing the head with bogus data now...

gregcotten: I went ahead and pushed a change for this that goes down the other proposed route. I strongly believe this will be much cleaner for library users (like me!)

gregcotten:
OK @Lukasa I, at my own risk, have diverged from your suggestion as I think it's really important that the `head` property is non-nil. I've instead initialize it with bogus data (HTTP 0.0, status code 0) and it's replaced before the library user gets a hold of it.

I've also added its use to existing unit tests, but can make a unique unit test if you have any suggestions.

gregcotten:
Needless to say this should be a squash merge if accepted :)

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
Lukasa: Yeah, the cleanliness argument is interesting, but I'm worried about whether we should be enforcing this invariant somewhere else. If we're _that_ confident that we will always have a result, why not store the internal as an optional and project through it before we pass it to the user? We can even have a release-mode fallback that doesn't crash but just provides garbage, and then in debug mode we'd crash.

gregcotten: @Lukasa

I'm not familiar with this "project through" pattern in Swift... 

Do you mean store a `var headResponse: HTTPHeadResponse?` optional in `FileDownloadDelegate` and construct a `Progress` struct every time reportProgress is called? Or is there some `Unsafe` pattern I'm not familiar with?

dnadoba: I think @Lukasa means something like this:
```swift
private var _head: HTTPResponseHead?
public var head: HTTPResponseHead { _head! }
// or
public var head: HTTPResponseHead { _head ?? .init(...) }
```

Lukasa: Yup, that would work nicely.

gregcotten: OK, done!

Sources/AsyncHTTPClient/FileDownloadDelegate.swift
```diff
@@ -19,13 +19,36 @@ import NIOPosix
...
-    /// "Content-Length" header (if available) and the count of bytes downloaded.
+    /// "Content-Length" header (if available), the count of bytes downloaded, and the
+    /// response head.
     public struct Progress: Sendable {
         public var totalBytes: Int?
         public var receivedBytes: Int
+
+        public var head: HTTPResponseHead {
+            get {
+                #if DEBUG
```
Lukasa, line 30: I think it's probably easier to just write this as `assert(self._head != nil)`. That keeps the code common always, and lets us leverage assertions mode instead of relying on specifically having `DEBUG` defined.

Lukasa: Also nit: we require explicit `self`.

Lukasa (commented):
Very minor nits here but this is looking good. I'll kick off CI for some more info there as well.

Lukasa:
Looks like the formatter wants a few tweaks from you.

gregcotten:
> Looks like the formatter wants a few tweaks from you.

Done!

Issue #753, opened 2024-07-20
`Task` doesn't correctly reference follow-on RequestBag objects so it cannot cancel them after redirect.


gregcotten:
**Update**

Reproducible test case:
https://github.com/gregcotten/ahc-download-cancel-bug

Just run `swift test`

It turns out that a redirect causes a new task to be created, so you can't count on execute(...)'s Task to be the only Task you'll need to cancel. I'll put a workaround on the above repo at some point.

----

**Original**

Stubbing this for now with hope to provide a reproducible test case when I have time...

I'm hoping to use structured concurrency Task cancellation to actually cancel a download, so I did something like this:

```swift
let download = Task {
    let delegate = try FileDownloadDelegate(path: tempDownloadPath)

    let downloadTask = HTTPClient.shared.execute(request: request, delegate: delegate, reportProgress: {
        print("progress: \($0.receivedBytes)")
    })

    _ = try await withTaskCancellationHandler {
        try await downloadTask.get()
    } onCancel: {
        print("cancelling download...")
        downloadTask.cancel()
        print("cancelled download")
    }
}

await Task.sleep(for: .seconds(5))
download.cancel()
try await download.value
```

The actual download progresses until it's done downloading the file, no matter what, and then `downloadTask.get()` returns without throwing an error. 

What am I doing wrong here?

gregcotten:
Follow up: canceling the `task` passed as a parameter in `reportProgress` can actually stop the download and fail the `downloadTask.get()`. Why can't I cancel the download outside of that scope? Obviously that would be something you'd want to do!

Lukasa:
Sorry for this sitting unanswered over the weekend @gregcotten, let me see if I can reproduce this.

Lukasa:
Ok, so I don't reproduce this locally with a simple alternative. Here's the complete code I'm running:

```
import NIOCore
import NIOPosix
import NIOHTTP1
import AsyncHTTPClient

private final class HTTPServer: ChannelInboundHandler {
    typealias InboundIn = HTTPServerRequestPart
    typealias OutboundOut = HTTPServerResponsePart

    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
        switch self.unwrapInboundIn(data) {
        case .head, .body:
            ()
        case .end:
            self.sendResponse(context: context)
        }
    }

    private func sendResponse(context: ChannelHandlerContext) {
[...]

PR #814, opened 2025-02-17
Propagate `Task` cancellation through subsequent redirects

gregcotten:
Discussed in https://github.com/swift-server/async-http-client/issues/753

No test yet, and would love advice on how to best make one...

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -956,7 +956,7 @@ internal struct RedirectHandler<ResponseType> {
         status: HTTPResponseStatus,
         to redirectURL: URL,
         promise: EventLoopPromise<ResponseType>
-    ) {
+    ) -> HTTPClient.Task<ResponseType>? {
         do {
             var redirectState = self.redirectState
             try redirectState.redirect(to: redirectURL.absoluteString)
@@ -976,13 +976,19 @@ internal struct RedirectHandler<ResponseType> {
                 headers: headers,
                 body: body
             )
-            self.execute(newRequest, redirectState).futureResult.whenComplete { result in
+
+            let newTask = self.execute(newRequest, redirectState)
+
+            newTask.futureResult.whenComplete { result in
                 promise.futureResult.eventLoop.execute {
                     promise.completeWith(result)
                 }
             }
+
+            return newTask
         } catch {
             promise.fail(error)
+            return nil
         }
     }
 }
--- a/Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
+++ b/Sources/AsyncHTTPClient/RequestBag+StateMachine.swift
@@ -594,6 +594,7 @@ extension RequestBag.StateMachine {
     enum FailAction {
         case failTask(Error, HTTPRequestScheduler?, HTTPRequestExecutor?)
         case cancelExecutor(HTTPRequestExecutor)
+        case propagateCancellation
         case none
     }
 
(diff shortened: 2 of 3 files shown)


gregcotten:
Should `redirectTask` be stored weakly? Not confident about the ownership model here.

Lukasa:
We avoid `weak` like the plague, so it should definitely be strongly held. I think in the idea solution this would form a singly-linked-list so the ownership won't be a problem, but I want to confirm that when I look at the patch.

Lukasa (commented):
Ok, I've taken a look.

First, I _really_ like this patch @gregcotten. You've done a good job of threading this nicely through the code, and followed all the existing coding style. It's really rare to get contributors who do such a great job of that, so I wanted to express my appreciation.

Secondly, though, I wonder if we can make this simpler. I think the mere _presence_ of a redirect task is sufficient to indicate that we need to cancel that task, as well as ourselves. So should we just unilaterally forward it on?

Relatedly, we need to ensure we only look at the redirect task while holding the lock.

gregcotten:
> First, I _really_ like this patch @gregcotten. You've done a good job of threading this nicely through the code, and followed all the existing coding style. It's really rare to get contributors who do such a great job of that, so I wanted to express my appreciation.

Thanks so much! I'm trying my best to follow the style 🕵


> Secondly, though, I wonder if we can make this simpler. I think the mere _presence_ of a redirect task is sufficient to indicate that we need to cancel that task, as well as ourselves. So should we just unilaterally forward it on?

So remove `FailAction.propagateCancellation` and just cancel the redirect task if it exists?

Something as simple as this? The fail action was `.none` previously since the state was `.finished`, so that's why I wanted to handle it with the `StateMachine`.

```swift
private func fail0(_ error: Error) {
    self.task.eventLoop.assertInEventLoop()

    let action = self.state.fail(error)

    self.executeFailAction0(action)

    self.redirectTask?.cancel()
}
```

> Relatedly, we need to ensure we only look at the redirect task while holding the lock.
Currently it's only being assigned when on `task`'s event loop
```swift
self.task.eventLoop.assertInEventLoop()
```

Lukasa:
> Something as simple as this? 

Yes, modulo the lock.

> Currently it's only being assigned when on task's event loop

Right, but it's read from anywhere. So we need the lock.

Lukasa:
As for the test, I think the right solution here is to rely on a httpbin endpoint with a long delay, plus a redirect. (The httpbin stuff is internal to the project)

You can use `/redirect/target` with the header `X-Target-Redirect-URL` set to where you want to go. That can redirect you to `/wait`, which will never complete, so you can use that as the trigger. Do the redirect, sleep for 500ms, then hit cancel. That should reliably work with your fix, and not reliably work without it.

gregcotten:
> Right, but it's read from anywhere. So we need the lock.

Actually re-reading this, I'm not sure I understand. I've made `redirectTask` a `private var`, like it's other mutable var `state` which is only accessed on the `task`'s event loop (via `self.task.eventLoop.assertInEventLoop()` calls). When I call `self.redirectTask.cancel()`, that is also on the `task` event loop. All reads and writes of `redirectTask` occur on that event loop.

Lukasa:
Oh I apologise, you're quite right. I am clearly losing my mind here.

gregcotten:
OK, @Lukasa, ready!

gregcotten:
Only last minute thought is that instead of simply propagating cancellation, we propagate the actual `fail(...)` `Error`

Lukasa:
Yeah, let's propagate the fail.

gregcotten:
OK @Lukasa, done and ready to go.

Lukasa (approved):
Awesome, I think this is looking really good.

Issue #790, opened 2024-12-06
Obtain Response URL


0xifarouk:
how can I get the response url?
in URLSession we can say: 
```Swift
let (data, response) = try await URLSession.shared.data(for: request)
print(response.url)
```

how to do that using AsyncHttpClient?

Lukasa:
Unfortunately, this information is not exposed today. We'd need some new API surface on the response to expose it, which in my view should also expose the rest of the redirect history.

gregcotten:
This is particularly bumpy problem when trying to replace `URLSession` with `async-http-client`. @Lukasa do you have any ideas on where this API surface should go?

Lukasa:
`Response` would be a fine place to put it, I think.

gregcotten:
> Unfortunately, this information is not exposed today. We'd need some new API surface on the response to expose it, which in my view should also expose the rest of the redirect history.

@Lukasa, I'll probably take a stab at this.

Would we want a simple list of URLs? 
```swift
var visitedURLs: [String]
```

 or actually pairing a url with its corresponding `HTTPResponseHead`
```swift
var visitedURLs: [(url: String, head: HTTPResponseHead)]
```

or something really wild
```swift
var visitedURLs: [(request: HTTPClient.Request, head: HTTPResponseHead)]
```

Lukasa:
So we should definitely avoid using a non-nominal type here, so I think the tuples are out: they will be hard to evolve if we want to add other things.

I wonder if we should go whole-hog and offer:


```swift
var history: [RedirectedRequest]

struct RedirectedRequest {
    var request: HTTPClient.Request
    var response: HTTPResponseHead
}
```

gregcotten:
> So we should definitely avoid using a non-nominal type here, so I think the tuples are out: they will be hard to evolve if we want to add other things.
> 
> I wonder if we should go whole-hog and offer:
> 
> var history: [RedirectedRequest]
> 
> struct RedirectedRequest {
>     var request: HTTPClient.Request
>     var response: HTTPResponseHead
> }

That could work! There are some cases where we would have no history (for example, no redirects occur), and we'd still want to be able to access the `URL` of the response (as desired in this issue). 

We could rename this struct something like `RequestResult` or `VisitedURL`,  and always store the "ultimate" request/response as the last one so we can have some nice API like:

```swift
var url: URL? {
   get {
      self.history.last?.request.url
   }
}
```

Lukasa:
[...]

PR #816, opened 2025-02-20
Add `didVisitURL` delegate method

gregcotten:
Trying to pave the way for closing https://github.com/swift-server/async-http-client/issues/790 with some direction from @Lukasa.

I have no idea where is best to insert this new delegate method. I'm currently doing it first thing in `receiveResponseHead0`, and not using an EventLoopFuture for back pressure management.  The state machine seems pretty fragile and I don't want to leave too much of an imprint. Trying to be a part of an EventLoopFuture chain seems really complicated and would really leave a mark on the codebase, so I'm wondering if it's possible to just warn the user "do not block"?

Anyway, just a jumping-off point and happy to take direction!

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -668,7 +668,16 @@ public protocol HTTPClientResponseDelegate: AnyObject {
     ///     - task: Current request context.
     func didSendRequest(task: HTTPClient.Task<Response>)
 
-    /// Called when response head is received. Will be called once.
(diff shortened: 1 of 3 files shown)


Lukasa:
So this is a totally fine place to begin, if we want to begin here. I think it's entirely acceptable to do the smallest possible work item to make something possible, and this would definitely meet that bar. So if you're happy with this approach, I think I say "let's do it".

gregcotten:
OK, @Lukasa, I think it's ready to go.

Lukasa (approved):
This LGTM, let's start with this.

PR #832, opened 2025-04-28
Fix a few simple sendability issues

glbrntt:
Motivation:

We're about to go on a sendability journey. Let's pick some low hanging fruit to get started.

Modifications:

- Add a few assume-isolated calls
- Stop using static var
- Use a dispatch group instead of a work item to wait for work to be done.

Result:

Fewer warnings

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+StateMachine.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+StateMachine.swift
@@ -26,7 +26,9 @@ extension HTTPConnectionPool {
                 self.connection = connection
             }
 
-            static let none = Action(request: .none, connection: .none)
+            static var none: Action {
+                Action(request: .none, connection: .none)
+            }
         }
 
         enum ConnectionAction {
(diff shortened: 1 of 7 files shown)


Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+StateMachine.swift
```diff
@@ -26,7 +26,9 @@ extension HTTPConnectionPool {
                 self.connection = connection
             }
 
-            static let none = Action(request: .none, connection: .none)
```
gjcairo, line 29: Potentially silly question but was this unsafe?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -223,17 +223,18 @@ public class HTTPClient {
             )
         }
         let errorStorage: NIOLockedValueBox<Error?> = NIOLockedValueBox(nil)
-        let continuation = DispatchWorkItem {}
+        let dispatchGroup = DispatchGroup()
+        dispatchGroup.enter()
```
Lukasa, line 227: Nit: this should probably use the condition variable workaround we used in NIO to avoid triggering the main thread checker.

glbrntt: Ah yeah, good idea

Lukasa (approved):

PR #834, opened 2025-04-28
Make the file download delegate sendable

glbrntt:
Motivation:

Delegates can be passed from any thread and are executed on an arbitrary event loop. That means they need to be Sendable. Rather than making them all Sendable in one go, we'll do the larger ones separately.

Modifications:

- Make FileDownloadDelegate sendable

Result:

Safe to pass FileDownloadDelegate across isolation domains

### DIFF
--- a/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
+++ b/Sources/AsyncHTTPClient/FileDownloadDelegate.swift
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import NIOConcurrencyHelpers
 import NIOCore
 import NIOHTTP1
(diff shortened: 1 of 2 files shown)


glbrntt:
API breakage expected:

```
3 breaking changes detected in AsyncHTTPClient:
  💔 API breakage: constructor FileDownloadDelegate.init(path:pool:reportHead:reportProgress:) is now with @preconcurrency
  💔 API breakage: constructor FileDownloadDelegate.init(path:pool:reportHead:reportProgress:) is now with @preconcurrency
  💔 API breakage: constructor FileDownloadDelegate.init(path:reportHead:reportProgress:) is now with @preconcurrency
```

PR #835, opened 2025-04-29
Make body stream writer sendable

glbrntt:
Motivation:

The body stream writer can be sent across isolation domains so should be sendable.

Modifications:

- Make it explicitly sendable
- Add appropriate preconcurrency annotations
- Wrap an iterator from swift-algorithms as it hasn't yet annotated its types with Sendable

Result:

Body stream writer is sendable

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -53,11 +53,10 @@ extension HTTPClient {
             }
 
             @inlinable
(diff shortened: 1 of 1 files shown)


glbrntt:
This is expected:

```
1 breaking change detected in AsyncHTTPClient:
  💔 API breakage: constructor HTTPClient.Body.StreamWriter.init(closure:) is now with @preconcurrency
```

Lukasa (approved):

PR #838, opened 2025-04-30
Make the ResponseAccumulator Sendable

glbrntt:
Motivation:

The response accumulator is a delegate which must be sendable as it's passed across isolation domains.

Modifications:

- Make delegates have a sendable requirement
- Make the response accumulator sendable

Result:

Delegates, and the response accumulator, are sendable

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPHandler.swift
+++ b/Sources/AsyncHTTPClient/HTTPHandler.swift
@@ -538,8 +538,12 @@ public final class ResponseAccumulator: HTTPClientResponseDelegate {
         }
     }
 
-    var history = [HTTPClient.RequestResponse]()
-    var state = State.idle
+    private struct MutableState: Sendable {
+        var history = [HTTPClient.RequestResponse]()
+        var state = State.idle
+    }
+
+    private let state: NIOLockedValueBox<MutableState>
     let requestMethod: HTTPMethod
     let requestHost: String
 
@@ -573,6 +577,7 @@ public final class ResponseAccumulator: HTTPClientResponseDelegate {
         self.requestMethod = request.method
         self.requestHost = request.host
         self.maxBodySize = maxBodySize
+        self.state = NIOLockedValueBox(MutableState())
     }
 
     public func didVisitURL(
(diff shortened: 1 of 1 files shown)


Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -573,107 +577,118 @@ public final class ResponseAccumulator: HTTPClientResponseDelegate {
...
-                self.state = .error(error)
-                return task.eventLoop.makeFailedFuture(error)
-            }
+        self.state.withLockedValue {
+            switch $0.state {
+            case .idle:
+                if self.requestMethod != .HEAD,
+                    let contentLength = head.headers.first(name: "Content-Length"),
+                    let announcedBodySize = Int(contentLength),
+                    announcedBodySize > self.maxBodySize
```
Lukasa, line 600: Can I suggest that we pull this computation out of the lock?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -968,7 +991,8 @@ extension HTTPClient {
         /// - returns: The value of ``futureResult`` when it completes.
         /// - throws: The error value of ``futureResult`` if it errors.
         @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-        public func get() async throws -> Response {
+        @preconcurrency
+        public func get() async throws -> Response where Response: Sendable {
```
Lukasa, line 995: These `where` clauses are superfluous, you've required `Response` to be `Sendable`.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -894,30 +910,35 @@ extension HTTPClient {
     ///
     /// Will be created by the library and could be used for obtaining
     /// `EventLoopFuture<Response>` of the execution or cancellation of the execution.
-    public final class Task<Response> {
+    @preconcurrency
+    public final class Task<Response: Sendable> {
```
Lukasa, line 914: Is the unchecked Sendable still needed for this? Can it be checked?

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -894,30 +910,35 @@ extension HTTPClient {
...
             set {
-                self.lock.withLock { self._taskDelegate = newValue }
+                self.state.withLockedValue { $0.taskDelegate = newValue }
             }
         }
 
-        private var _isCancelled: Bool = false
-        private var _taskDelegate: HTTPClientTaskDelegate?
-        private let lock = NIOLock()
         private let makeOrGetFileIOThreadPool: () -> NIOThreadPool
```
Lukasa, line 942: Looks like this needs an `@Sendable`.

Sources/AsyncHTTPClient/HTTPHandler.swift
Lukasa, line 914: Also, how sure are we that this constraint is needed?

Lukasa (approved):

PR #840, opened 2025-04-30
Fix a few more Sendability warnings in Sources

glbrntt:
Motivation:

There are a couple of sendability warnings leftover in Sources.

Transaction moves a closure into a task. The closure isn't Sendable (and shouldn't be). However, higher up the stack there's a closure which generates the non-sendable closure which can be sendable.

Modifications:

- Pass the sendable closure generating closure down rather
- Add a few more explicit sendable annotations

Result:

Fewer warnings


Lukasa (approved):
Nice and straightforward.

PR #849, opened 2025-07-25
Add HTTP examples for POST PUT DELETE

iwa:
hi!

thanks a lot for your great lib

one thing really missing tho is more examples to help newbies using your lib, so i've added a few really basic ones to at least show how to use Post, Put & Delete
examples are made about a little local Todo api, nothing extravagant

let me know if something isn't right for y'all!

### DIFF
--- a/Examples/DeleteJSON/DeleteJSON.swift
+++ b/Examples/DeleteJSON/DeleteJSON.swift
@@ -1,0 +1,23 @@
+import AsyncHTTPClient
+import Foundation
+import NIOCore
+import NIOFoundationCompat
+
+@main
+struct PostJSON {
+    static func main() async throws {
+        let httpClient = HTTPClient(eventLoopGroupProvider: .singleton)
+
+        do {
+            var request = HTTPClientRequest(url: "http://localhost:8080/todos/1)")
+            request.method = .DELETE
+
+            let response = try await httpClient.execute(request, timeout: .seconds(30))
+            print("HTTP head", response)
+        } catch {
+            print("request failed:", error)
+        }
+        // it is important to shutdown the httpClient after all requests are done, even if one failed
+        try await httpClient.shutdown()
+    }
+}
--- a/Examples/Package.swift
+++ b/Examples/Package.swift
@@ -26,6 +26,9 @@ let package = Package(
     products: [
         .executable(name: "GetHTML", targets: ["GetHTML"]),
         .executable(name: "GetJSON", targets: ["GetJSON"]),
+        .executable(name: "PostJSON", targets: ["PostJSON"]),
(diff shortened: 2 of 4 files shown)


Examples/DeleteJSON/DeleteJSON.swift
```diff
@@ -0,0 +1,23 @@
...
+import NIOCore
+import NIOFoundationCompat
+
+@main
+struct PostJSON {
+    static func main() async throws {
+        let httpClient = HTTPClient(eventLoopGroupProvider: .singleton)
+
+        do {
+            var request = HTTPClientRequest(url: "http://localhost:8080/todos/1)")
```
glbrntt: I generally think it's best of examples can be run without the user needing to do anything extra like run a local server. Had you thought about using something like [httpbin](https://httpbin.org) instead?

Examples/DeleteJSON/DeleteJSON.swift
```diff
@@ -0,0 +1,23 @@
+import AsyncHTTPClient
+import Foundation
+import NIOCore
+import NIOFoundationCompat
+
+@main
+struct PostJSON {
```
glbrntt: This example does a `DELETE` ;)

Examples/DeleteJSON/DeleteJSON.swift
```diff
@@ -0,0 +1,23 @@
...
+    static func main() async throws {
+        let httpClient = HTTPClient(eventLoopGroupProvider: .singleton)
```
glbrntt: You can actually use `HTTPClient.shared` instead here which you don't need to shutdown

glbrntt (changes requested):

Examples/DeleteJSON/DeleteJSON.swift
iwa: actually that's a better idea, i didnt know about httpbin but we might aswell use jsonplaceholder as they offer specific data formats like todos, posts, users, etc... so it resembles more closely "real-world usage"

Examples/DeleteJSON/DeleteJSON.swift
iwa: ah frick, missed it, fixed in 90ca4d7

Examples/DeleteJSON/DeleteJSON.swift
iwa: fixed by e3e7f31

out of curiosity what is the best practice between using the shared vs creating your own client on the fly?

Examples/DeleteJSON/DeleteJSON.swift
iwa: fixed and working at 2cb31f4

PR #853, opened 2025-08-18
Weakly capture the `HTTPConnectionPool` in the idle connection timeout task

ahoppen:
Currently, the `HTTPConnectionPool` will be kept alive until the idle connection timeout has fired. I don’t see a reason to keep the `HTTPConnectionPool` around if the idle connection timeout future is the only thing that’s still referencing the `HTTPConnectionPool`.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
@@ -394,7 +394,10 @@ final class HTTPConnectionPool:
                 "ahc-connection-id": "\(connectionID)"
             ]
         )
-        let scheduled = eventLoop.scheduleTask(in: self.idleConnectionTimeout) {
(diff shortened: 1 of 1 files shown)


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -394,7 +394,10 @@ final class HTTPConnectionPool:
                 "ahc-connection-id": "\(connectionID)"
             ]
         )
-        let scheduled = eventLoop.scheduleTask(in: self.idleConnectionTimeout) {
+        let scheduled = eventLoop.scheduleTask(in: self.idleConnectionTimeout) { [weak self] in
```
Lukasa, line 397: We should avoid the use of a weak reference here: its performance impact will be very substantial. 

Instead, so long as all of these tasks are cancelled in the shutdown routine (as they should be), there will be no leak, as the cancellation of the tasks will drop the function reference.

PR #856, opened 2025-09-08
Add support for HTTP/1 connection pre-warming

Lukasa:
Motivation

This patch adds support for HTTP/1 connection pre-warming. This allows the user to request that the HTTP/1 connection pool create and maintain extra connections, above-and-beyond those strictly needed to run the pool. This pool can be used to absorb small spikes in request traffic without increasing latency to account for connection creation.

Modifications

- Added new configuration properties for pre-warmed connections.
- Amended the HTTP/1 state machine to create new connections where necessary.
- Added state machine tests.

Results

Pre-warmed connections are available.

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
@@ -108,8 +108,7 @@ final class HTTPConnectionPool:
                 case closeConnectionAndCreateConnection(
                     close: Connection,
                     newConnectionID: Connection.ID,
-                    on: EventLoop,
-                    isShutdown: StateMachine.ConnectionAction.IsShutdown
+                    on: EventLoop
                 )
                 case cleanupConnections(CleanupContext, isShutdown: StateMachine.ConnectionAction.IsShutdown)
                 case migration(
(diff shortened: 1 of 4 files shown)


Sources/AsyncHTTPClient/ConnectionPool/State Machine/HTTPConnectionPool+HTTP1StateMachine.swift
```diff
@@ -495,6 +560,38 @@ extension HTTPConnectionPool {
...
+            let nonLeased = self.connections.generalPurposeStats.nonLeased
+            let neededNonLeased = self.requests.generalPurposeCount + self.preWarmedConnectionCount
+
+            let connectionAction: EstablishedConnectionAction
+            if nonLeased < neededNonLeased && self.connections.canGrow {
+                // We re-use the EL of the connection we just closed.
+                let newConnectionID = self.connections.createNewConnection(on: connectionToClose.eventLoop)
+                connectionAction = .closeConnectionAndCreateConnection(
+                    closeConnection: connectionToClose,
+                    isShutdown: .no,
```
fabianfett: as far as I can see this is always `no`. Therefore we should likely remove it.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool.swift
```diff
@@ -104,6 +105,12 @@ final class HTTPConnectionPool:
             enum Unlocked {
                 case createConnection(Connection.ID, on: EventLoop)
                 case closeConnection(Connection, isShutdown: StateMachine.ConnectionAction.IsShutdown)
+                case closeConnectionAndCreateConnection(
+                    close: Connection,
+                    newConnectionID: Connection.ID,
+                    on: EventLoop,
+                    isShutdown: StateMachine.ConnectionAction.IsShutdown
```
fabianfett: As mentioned elsewhere we don't need this one here.

fabianfett (approved):
This looks really good! Thanks!

PR #857, opened 2025-09-10
Introduce built-in swift-distributed-tracing support

ktoso:
Almost 5 years after the initial tracing proposal PR for this HTTPClient, I'd like to propose it again.

Popular server packages _should_ be embracing distributed tracing, and none is more crucial in this than the async-http-client.

This PR proposes an approach to include tracing with:
- using package traits, from Swift 6.1
- the `TracingSupport` trait is enabled by default
- the trait can be disabled to entirely remove the tracing dependency along with any code using it (and potential overheads).

The details for where to start/end spans are TBD, we can do better than this.

Currently there is a single span that starts before we send the request, and ends as we get the response head. We can further improve this with spans for streaming the request body, spans for the time it takes to _send the request_ etc. However first I'd like to get agreement on this approach for the instrumentation.

### The spans

This creates _very basic_ spans. Basically when you do `client.get()/post()` etc, as well as the asynchronous `try await client.execute()` we create a span that:

- starts: when the request starts -- we could keep improving where EXACTLY this starts, there's lots of opinions to have about this 😉 
- ends: when we receive the response head

There's no spans yet for the response body streaming or the waiting in the pool etc. All this can be added incrementally.

⏭️ The spans don't have fancy attributes yet; 

⏭️ and the error cases are likely not fully covered. If someone more familiar with HTTPClient wants to take that on that'd be best, but I can follow-up with more handling once we're happy with this approach in general.

I think this should be done in follow-ups, so we can keep polishing up the tracing support as we use it in practice.

### Availability woes

It's been somewhat painful to not cause the entire HTTPClient types to suddenly get @availability which would be a breaking change; So I'm playing a bunch of tricks with storing `any Sendable` as storage and casting that... We could probably use unsafe bitcasts, but taking this conservatively for now -- it should work first and foremost, and we can spend time shaving off overhead eventually.

I don't like one bit where we're doing that, the  `mutating func startRequestSpan<T>(tracer: T?) {}` is kinda horrible, but if we make this have `Tracer` we end up forcing the surrounding type to have availability... 

⏭️ We could think of some better ways to express this, but it's all internal so I think we can take this incrementally.

### Configuration

It is possible to set a tracer "per client" by using the Configuration, which in combination with the in memory tracer (Incoming here https://github.com/apple/swift-distributed-tracing/pull/180), will be able to be used for testing. One can also pass `nil` to configuration to disable picking up the global tracer during configuration.

We can use this for testing, and it's important for e.g. the otel library so when exporting while using the httpclient, we don't necessarily also cause spans of that to be exported -- if we don't care we can disable it this way. (alternatively, by making a parent span inside swift-otel that is NOT recording also works)

### Context propagation

✅ This also includes header propagation, although I've ran out of night to make a test for it -- I'll follow up on it shortly.

Resolves https://github.com/slashmo/gsoc-swift-tracing/issues/46
Resolves https://github.com/slashmo/gsoc-swift-tracing/issues/95

Replaces https://github.com/swift-server/async-http-client/pull/320
Replaces https://github.com/swift-server/async-http-client/pull/289

cc @slashmo @simonjbeaumont @czechboy0 @FranzBusch @fabianfett

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,7 @@ jobs:
             linux_5_10_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -warnings-as-errors"
             linux_6_0_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -warnings-as-errors"
             linux_6_1_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -warnings-as-errors"
+            linux_6_2_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -warnings-as-errors"
             linux_nightly_next_arguments_override: "--explicit-target-dependency-import-check error"
             linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error"
 
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -17,6 +17,7 @@ jobs:
             linux_5_10_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -warnings-as-errors"
             linux_6_0_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -warnings-as-errors"
             linux_6_1_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -warnings-as-errors"
+            linux_6_2_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -warnings-as-errors"
             linux_nightly_next_arguments_override: "--explicit-target-dependency-import-check error"
             linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error"
 


Package@swift-6.0.swift
```diff
@@ -0,0 +1,131 @@
...
+                .product(name: "NIOHTTP2", package: "swift-nio-http2"),
+                .product(name: "NIOSSL", package: "swift-nio-ssl"),
+                .product(name: "NIOHTTPCompression", package: "swift-nio-extras"),
+                .product(name: "NIOSOCKS", package: "swift-nio-extras"),
+                .product(name: "NIOTransportServices", package: "swift-nio-transport-services"),
+                .product(name: "Logging", package: "swift-log"),
+                .product(name: "Atomics", package: "swift-atomics"),
+                .product(name: "Algorithms", package: "swift-algorithms"),
+                // Observability support
+                .product(name: "Tracing", package: "swift-distributed-tracing", condition: .when(traits: ["TracingSupport"])),
```
ktoso: The reason we need a -6.0 package file is because this `.when` is only available 6.0+.

Open to other ideas what we can do here, but since we want to keep support for 5.10, I think we need to keep the separate file for now?

czechboy0: Huh, I would have expected this to need to be `-6.1`, since that's when traits arrived?

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
```diff
@@ -51,7 +55,8 @@ final class Transaction:
     }
 
     func cancel() {
-        self.fail(CancellationError())
+        let error = CancellationError()
+        self.fail(error)
```
czechboy0: Needed?

Package@swift-6.0.swift
```diff
@@ -0,0 +1,131 @@
...
+    name: "async-http-client",
+    platforms: [ // FIXME: must remove this
+        .macOS("10.15")
+    ],
+    products: [
+        .library(name: "AsyncHTTPClient", targets: ["AsyncHTTPClient"])
+    ],
+    traits: [
+        .trait(name: "TracingSupport"),
+        .default(enabledTraits: ["TracingSupport"]),
```
czechboy0: I'm surprised this works in a -6.0 manifest.

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
```diff
@@ -25,7 +25,11 @@ extension HTTPClient {
     /// A wrapper for `POSIX` errors thrown by `Network.framework`.
     public struct NWPOSIXError: Error, CustomStringConvertible {
         /// POSIX error code (enum)
+        #if compiler(>=6.1)
+        nonisolated(unsafe) public let errorCode: POSIXErrorCode
+        #else
         public let errorCode: POSIXErrorCode
+        #endif
```
czechboy0: Is this needed for this PR?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -57,6 +61,9 @@ let globalRequestID = ManagedAtomic(0)
 ///         }
 ///     }
 /// ```
+// #if TracingSupport
+// @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+// #endif
```
czechboy0: We won't be adding this, will we? I think that'd be an API-break.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -71,7 +78,14 @@ public final class HTTPClient: Sendable {
     private let state: NIOLockedValueBox<State>
     private let canBeShutDown: Bool
 
-    static let loggingDisabled = Logger(label: "AHC-do-not-log", factory: { _ in SwiftLogNoOpLogHandler() })
+    #if TracingSupport
+    @_spi(Tracing)
+    public var tracer: (any Tracer)? { // Tracer requires macOS 10.15... so we're forced into annotating HTTPClient :-/
```
czechboy0: Can't the availability only be on this property getter, instead of the client itself?

FranzBusch: No it can't since availability can't be added on stored properties. This is going to be a problem since the availability above is going to be an API break as you said. We can work around this though by creating our own protocol and injecting it with runtime availability checks

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -884,6 +919,17 @@ public final class HTTPClient: Sendable {
         /// A method with access to the HTTP/2 stream channel that is called when creating the stream.
         public var http2StreamChannelDebugInitializer: (@Sendable (Channel) -> EventLoopFuture<Void>)?
 
+        #if TracingSupport
+        /// Tracer that should be used by the HTTPClient. 
+        /// 
+        /// This is selected at configuration creation time, and if no tracer is passed explicitly,
+        /// (including `nil` in order to disable traces), the default global bootstrapped tracer will
+        /// be stored in this property, and used for all subsequent requests made by this client.
+        public var tracer: (any Tracer)? = InstrumentationSystem.tracer
```
FranzBusch: Do we really need this? Is this in general our recommendation that libraries are offering `Tracer` APIs for injection?

Sources/AsyncHTTPClient/HTTPClient.swift
czechboy0: In my other PR I just used an erased stored property, with typed accessors behind availability.

Sources/AsyncHTTPClient/HTTPClient.swift
czechboy0: It's important to make it unit testable.

Package@swift-6.0.swift
sebsto: In the Swift Lambda Runtime, we use a simple `define` in 6.0 to replace the traits we use in 6.1

In Package@swift-6.0.swift:
https://github.com/swift-server/swift-aws-lambda-runtime/blob/303559d9eda20c579b4b0a5fe25eef424ebab217/Package%40swift-6.0.swift#L6

In the code:
https://github.com/swift-server/swift-aws-lambda-runtime/blob/303559d9eda20c579b4b0a5fe25eef424ebab217/Sources/AWSLambdaRuntime/FoundationSupport/Lambda%2BJSON.swift#L15

The define and the traits have the same name, allowing us to use just one `#if` in the code.

Sources/AsyncHTTPClient/HTTPClient.swift
slashmo: Not necessarily. Adam worked around this in Hummingbird's tracing middleware tests by using a task-local tracer instead.

czechboy0: It still requires using the process-wide bootstrap and adding the trampoline that then inspects the task local, doesn't it?

That's a fine approach, but is still not great when writing unit tests that you want all to be independent and be able to run in parallel.

Ideally it should be possible not to touch the bootstrap method and pass it explicitly, just like the Logger.

Package@swift-6.0.swift
0xTim: It probably does as long as the compiler is new enough

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: We won't be, that's the dance where we need to do something about the tracer storage that we talked about -- we might need to do an `let _tracer: (any Sendable)?; var tracer: (any Tracer)? { _tracer as! (any Tracer)? }` since we cna't put availability onto a stored property 🤔 

I want to make sure we want the property before I do the ugly stuff like that

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: Here it can be; it can't be on a stored property though, so we'll need a hack in the Config object

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: Yeah, that's something to discuss and decide here I think.

What this enables is:

- `HTTPClient(..., Configuration())` -> default, picks up bootstrapped tracer at configuration time
- `HTTPClient(..., Configuration(tracer: nil))` -> tracing is enabled overall in the process, but in this specific http client it is disabled (for whatever reason)
- `HTTPClient(..., Configuration(tracer: myTracer))` -> use this specific tracer in this specific client
- **disable** the `TracingSupport` trait -> no tracing, ofc

The opposite would be to not have the configuration option and then we just have:
- `HTTPClient()` -> pick bootstrapped tracer when executing requests, e.g. `client.get()` would do the usual `startSpan` (global); 
- **disable** the `TracingSupport` trait -> no tracing, ofc

So for testing in this scenario you could make a tracer that is "Pick task-local tracer", bootstrap that, and then:

```swift
TaskLocalTracer.withValue(myInMemoryTracer) { client.get() }
```

So both allow for testing, just slightly differently. We don't have the `TaskLocalTracer.withValue` yet, but this would be a good utility to add to swift-distributed-tracing along with the in memory tracer that we're adding now. IMHO it's good that the in memory tracer does NOT assume that it is the task local one; but we can combine the two things for testing if we need to.


In Valkey @fabianfett strongly argued for the pattern (1) with the configurable object, and I've done so before as well in the cluster... but yeah, I don't think we have to do this -- we can totally stick to the pattern (2).

It is also true that starting with (2) we can always later on add the configuration if we determine it'd be nicer to do so after all.

Let me know what you think folks; perhaps we can do the simplest (2) for now and if we notice it would have been good, we can later on (if ever) add the configuration "override".

Package@swift-6.0.swift
ktoso: Hm, interesting trick @sebsto! I guess though we might be happy enough with just 6.1; my bad for 6.0 file, removing that!

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
ktoso: removed, thx

Sources/AsyncHTTPClient/HTTPClient.swift
slashmo: > tracing is enabled overall in the process, but in this specific http client it is disabled (for whatever reason)

This is a very good point. One such example where we'd definitely want to turn it off is for the `HTTPClient` in Swift OTel used to export telemetry, even though the users' `HTTPClient`s should be instrumented.

czechboy0: I think the task locals question is a larger one I'd like us to discuss separately (for all of logs, metrics, and tracing).

I like the ability to provide the Tracer explicitly for the same reason we recommend folks have the ability to provide the Logger. I'm not sure why we'd want to diverge, and ask folks to provide Loggers explicitly, but have to jump through the process-global to achieve the same for the Tracer.

Package.swift
```diff
@@ -47,6 +47,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-log.git", from: "1.6.0"),
         .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.2"),
         .package(url: "https://github.com/apple/swift-algorithms.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.2.0"),
```
slashmo:
```suggestion
        .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.3.0"),
```

Should this be `1.3.0` to require the `InMemoryTracer`?

ktoso: Good point! I missed the thanks

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -36,12 +40,27 @@ extension HTTPClient {
...
+            return try await tracer.withSpan("\(request.method)") { span -> (HTTPClientResponse) in
+                let attr = self.configuration.tracing.attributeKeys
+                span.attributes[attr.requestMethod] = request.method.rawValue
+                // Set more attributes on the span
+                return try await doExecute()
+            }
+        }
+        #endif
+
+        return try await doExecute()
```
Lukasa: This feels like the wrong thing was factored out. Rather than pull the `executeAndFollowRedirectsIfNeeded` call into a sub-function, it seems like it might be nicer to have a helper function like this:

```
private func withRequestSpan<ReturnType>(_ request: HTTPClientRequest, spanName: @autoclosure () -> String, _ body: () async throws -> ReturnType) async throws -> ReturnType {
    #if TracingSupport
    if let tracer = self.tracer {
        return try await tracer.withSpan("\(request.method)") { span -> (HTTPClientResponse) in
            let attr = self.configuration.tracing.attributeKeys
            span.attributes[attr.requestMethod] = request.method.rawValue
            // Set more attributes on the span
            return try await body()
        }
    }
    #else
    return try await body()
    #endif
} 
```


The call site then becomes:

```swift
    public func execute(
        _ request: HTTPClientRequest,
        deadline: NIODeadline,
        logger: Logger? = nil
    ) async throws -> HTTPClientResponse {
        self.withRequestSpan(request) {
            try await self.executeAndFollowRedirectsIfNeeded(
                request,
                deadline: deadline,
                logger: logger ?? Self.loggingDisabled,
                redirectState: RedirectState(self.configuration.redirectConfiguration.mode, initialURL: request.url)
            )
        }
    }
```

I think this is a better separation of concerns. It keeps the `#if` out of the main control flow, which now looks unconditional, and it lets the code reader have an easier time seeing what is _supposed_ to happen always.

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
Lukasa: Not removed so far as I can see.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -71,7 +75,15 @@ public final class HTTPClient: Sendable {
...
-    static let loggingDisabled = Logger(label: "AHC-do-not-log", factory: { _ in SwiftLogNoOpLogHandler() })
+    #if TracingSupport
+    @_spi(Tracing)
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    public var tracer: (any Tracer)? {
+        configuration.tracing.tracer
+    }
+    #endif  // TracingSupport
+
+    public static let loggingDisabled = Logger(label: "AHC-do-not-log", factory: { _ in SwiftLogNoOpLogHandler() })
```
Lukasa: Why did this become `public`?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -705,6 +717,22 @@ public final class HTTPClient: Sendable {
...
+        // } else if let tracer = self.tracer {
+        //     let s = tracer.startSpan(request.method.rawValue)
+        //     let attrs = self.configuration.tracing.attributeKeys
+        //     s.attributes[attrs.requestMethod] = request.method.rawValue
+        //     s.attributes["loc"] = "\(#fileID):\(#line)"
+        //     span = s
+        // } else {
+        //     span = nil
+        // }
+        // #endif
```
Lukasa: Why is this commented out?

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
Lukasa: Reminder on this one.

Sources/AsyncHTTPClient/TracingSupport.swift
```diff
@@ -0,0 +1,135 @@
...
+}
+#endif  // TracingSupport
+
+// #if TracingSupport
+// @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+// typealias HTTPClientTracingSupportTracerType = any Tracer
+// #else
+// enum TracingSupportDisabledTracer {}
+// typealias HTTPClientTracingSupportTracerType = TracingSupportDisabledTracer
+// #endif
```
Lukasa: Why is this commented out?

Sources/AsyncHTTPClient/TracingSupport.swift
```diff
@@ -0,0 +1,135 @@
...
+        self.activeSpan = nil
+    }
+
+    /// The request span currently ends when we receive the response head.
+    mutating func endRequestSpan(response: HTTPResponseHead) {
+        guard let span = activeSpan else {
+            return
+        }
+
+        span.attributes["http.response.status_code"] = SpanAttribute.int64(Int64(response.status.code))
```
Lukasa: Presumably this should be in the config.

Tests/AsyncHTTPClientTests/HTTPClientBase.swift
```diff
@@ -61,9 +61,11 @@ class XCTestCaseHTTPClientTestsBaseClass: XCTestCase {
             }
         )
         backgroundLogger.logLevel = .trace
+        let configuration = HTTPClient.Configuration().enableFastFailureModeForTesting()
+
         self.defaultClient = HTTPClient(
             eventLoopGroupProvider: .shared(self.clientGroup),
-            configuration: HTTPClient.Configuration().enableFastFailureModeForTesting(),
+            configuration: configuration,
```
Lukasa: Doesn't seem like this change is needed.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
ktoso: That's indeed nicer, let me do that 👍

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
ktoso: Sorry about that, seems I messed something up while moving between laptops with this change; sorry for the leftovers, following up on those;

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: Doesn't need to, thanks for noticing

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: Sorry for the leftover, was experimenting with different places to instrument; removing this, wasn't the best place

Sources/AsyncHTTPClient/NIOTransportServices/NWErrorHandler.swift
ktoso: Hmmm, on my other laptop 6.2 was really unhappy about this though on my main now can't reproduce this; Will follow up separately why that was -- 

Removing the change meanwhile

Tests/AsyncHTTPClientTests/HTTPClientTracingTests.swift
```diff
@@ -0,0 +1,118 @@
...
+            return
+        }
+        guard let span = tracer.finishedSpans.first else {
+            XCTFail("No span was recorded!")
+            return
+        }
+
+        XCTAssertEqual(span.operationName, "GET")
+    }
+}
```
ktoso: - [ ] TODO: more tests once we're happy enough with the general approach of dependency and instrumentation. Cancellation and error cases specifically need some tests and likely be handled more correctly

Sources/AsyncHTTPClient/TracingSupport.swift
ktoso: Right, need to find a way to weave config over here; but yes 👍

Package.swift
```diff
@@ -73,6 +74,9 @@ let package = Package(
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Atomics", package: "swift-atomics"),
                 .product(name: "Algorithms", package: "swift-algorithms"),
+                // Observability support
```
FranzBusch, line 76: Nit: If you wanna be strict you gotta move Logging here as well :D

Package@swift-6.1.swift
```diff
@@ -0,0 +1,148 @@
...
+
+let package = Package(
+    name: "async-http-client",
+    products: [
+        .library(name: "AsyncHTTPClient", targets: ["AsyncHTTPClient"])
+    ],
+    traits: [
+        .trait(name: "TracingSupport"),
+        .default(enabledTraits: ["TracingSupport"]),
+    ],
```
FranzBusch: I personally am not convinced that this should be guarded by a trait. We don't guard logging by a trait either and IMO there is no difference between logging or tracing. Furthermore, this would be the first trait for AHC which means we can't move existing functionality behind default traits anymore if we want to since that can only be done once with the first trait.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -36,12 +40,33 @@ extension HTTPClient {
...
+    func withRequestSpan<ReturnType>(
+        _ request: HTTPClientRequest,
+        _ body: () async throws -> ReturnType
+    ) async rethrows -> ReturnType {
+        #if TracingSupport
+        if let tracer = self.tracer {
+            return try await tracer.withSpan("\(request.method)") { span in
+                let attr = self.configuration.tracing.attributeKeys
+                span.attributes[attr.requestMethod] = request.method.rawValue
+                // Set more attributes on the span
```
FranzBusch: Should we treat this as a todo?

Package@swift-6.1.swift
ktoso: You folks tell me what to do here.

This adoption has been held up for years for unclear reason, so if we changed our minds and are happy with direct dependency that's IMHO absolutely great and let's do it!

Need a final yay/nay here please :-)

czechboy0: I agree with Franz here, no need for a trait IMO. But it's really more up to @Lukasa and friends

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
ktoso: Yes, I'd like to get confirmation on the dependency approach before I pour more time into polishing this up though

Package@swift-6.1.swift
Lukasa: Fine with dropping the trait.

0xTim: +1

ktoso: Fantastic, let me follow up then!

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
ktoso: Gonna follow up now that we know how to approach the dependency; will prep a more complete PR and leave some TODO and issue for more attributes 👍

Package.swift
ktoso: True, done :)

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
ktoso: Added handling of at status codes and recording >= 400 ones as well as errors

Sources/AsyncHTTPClient/TracingSupport.swift
ktoso: Fixed this in https://github.com/swift-server/async-http-client/pull/857/commits/91ea66eabd96d9d6699098fea0cf6f68c8fbcef9

Package.swift
```diff
@@ -70,9 +71,12 @@ let package = Package(
                 .product(name: "NIOHTTPCompression", package: "swift-nio-extras"),
                 .product(name: "NIOSOCKS", package: "swift-nio-extras"),
                 .product(name: "NIOTransportServices", package: "swift-nio-transport-services"),
-                .product(name: "Logging", package: "swift-log"),
                 .product(name: "Atomics", package: "swift-atomics"),
                 .product(name: "Algorithms", package: "swift-algorithms"),
+                // Observability support
+                .product(name: "Logging", package: "swift-log"),
+                .product(name: "Tracing", package: "swift-distributed-tracing"),
+                .product(name: "InMemoryTracing", package: "swift-distributed-tracing"),
```
fabianfett: Why does the main target depend on the in memory tracer? Shouldn't this be only for tests?

ktoso: Totally just my silly mistake; removing

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1012,6 +1035,90 @@ public final class HTTPClient: Sendable {
...
+                connectionPool: connectionPool,
+                proxy: proxy,
+                ignoreUncleanSSLShutdown: ignoreUncleanSSLShutdown,
+                decompression: decompression
+            )
+            self.http1_1ConnectionDebugInitializer = http1_1ConnectionDebugInitializer
+            self.http2ConnectionDebugInitializer = http2ConnectionDebugInitializer
+            self.http2StreamChannelDebugInitializer = http2StreamChannelDebugInitializer
+            self.tracing = tracing
+        }
```
fabianfett, line 1065: General comment, nothing to act upon here: This is becoming weirder and weirder. We should just force users to set properties instead, like in TLSConfiguration.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1012,6 +1035,90 @@ public final class HTTPClient: Sendable {
...
+
+        @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+        public init(
+            tracer: (any Tracer)? = InstrumentationSystem.tracer,
+            attributeKeys: AttributeKeys = .init()
+        ) {
+            self._tracer = tracer
+            self.attributeKeys = attributeKeys
+        }
+
```
fabianfett: Do we need a callback to allow overwriting the span name?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
```diff
@@ -36,12 +37,33 @@ extension HTTPClient {
...
+            )
+        }
+    }
+
+    @inlinable
+    func withRequestSpan<ReturnType>(
+        _ request: HTTPClientRequest,
+        _ body: () async throws -> ReturnType
+    ) async rethrows -> ReturnType {
+        #if TracingSupport
```
czechboy0: Leftover trait?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1012,6 +1035,90 @@ public final class HTTPClient: Sendable {
...
+        /// Span attribute keys that the HTTPClient should set automatically.
+        /// This struct allows the configuration of the attribute names (keys) which will be used for the apropriate values.
+        public struct AttributeKeys: Sendable {
```
czechboy0: Could we hide this as internal, to avoid baking this into the API? Whether attribute keys need to be customizable seems like a discussion we can have separately, and start simpler for now.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
ktoso: Oh thanks!

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -62,15 +63,29 @@ public final class HTTPClient: Sendable {
...
+
     /// Shared thread pool used for file IO. It is lazily created on first access of ``Task/fileIOThreadPool``.
     private let fileIOThreadPool: NIOLockedValueBox<NIOThreadPool?>
 
     private let state: NIOLockedValueBox<State>
     private let canBeShutDown: Bool
 
+    /// Tracer configured for this HTTPClient at configuration time.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    public var tracer: (any Tracer)? {
```
czechboy0, line 79: Also could be internal.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -924,6 +925,12 @@ extension HTTPClient {
         /// The `Logger` used by the `Task` for logging.
         public let logger: Logger  // We are okay to store the logger here because a Task is for only one request.
 
+        @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+        public var tracer: (any Tracer)? {
```
czechboy0, line 929: Same re public - does it have to be?

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: Maybe, I'd love to keep all the extra customization to future PRs though

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: Let's hide it for now, agreed; We can figure out the customization approach later

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: I'd keep this one public tbh; you want to know if a client is tracing or not perhaps when debugging some "why am i not getting traces" etc?

Sources/AsyncHTTPClient/HTTPHandler.swift
ktoso: I kinda think it's the same as Logger; if Logger is public this would be too

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
czechboy0: I'm a bit surprised the tests passed, since this was compiled out

ktoso: Unused function! I moved it elsewhere and forgot to remove this one

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Logging
+import NIOCore
+import NIOHTTP1
+import Tracing
+
+import struct Foundation.URL
```
glbrntt: Some of these imports don't seem necessary. Maybe only `Tracing` is required here?

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
```diff
@@ -34,6 +35,23 @@ final class Transaction:
 
     private let state: NIOLockedValueBox<StateMachine>
 
+    init(
+        request: HTTPClientRequest.Prepared,
+        requestOptions: RequestOptions,
+        logger: Logger,
+        connectionDeadline: NIODeadline,
+        preferredEventLoop: EventLoop,
+        span: (any Span)?,
```
glbrntt: This `span` isn't used; is that intentional? (Maybe the `init` isn't used?)

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1012,6 +1035,84 @@ public final class HTTPClient: Sendable {
...
+        }
+
+        /// Span attribute keys that the HTTPClient should set automatically.
+        /// This struct allows the configuration of the attribute names (keys) which will be used for the apropriate values.
+        @usableFromInline
+        package struct AttributeKeys: Sendable {
+            @usableFromInline package var requestMethod: String = "http.request.method"
+            @usableFromInline package var requestBodySize: String = "http.request.body.size"
+
+            @usableFromInline package var responseBodySize: String = "http.response.size"
```
glbrntt: Is this missing a "body"? `requestBodySize` has one.

```suggestion
            @usableFromInline package var responseBodySize: String = "http.response.body.size"
```

ktoso: Good one, both actually exist it seems: 

- `http.response.body.size` - The size of the response payload body in bytes. This is the number of bytes transferred excluding headers and is often, but not always, present as the Content-Lengthheader. For requests using transport encoding, this should be the compressed size.
- `http.response.body.size` - The size of the response payload body in bytes. This is the number of bytes transferred excluding headers and is often, but not always, present as the [Content-Length](https://www.rfc-editor.org/rfc/rfc9110.html#field.content-length) header. For requests using transport encoding, this should be the compressed size.

but that's pending a whole discussion how far we wanna push the otel specific semantics here... 

Thanks, I'll fix this to be the body!

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
ktoso: Hm yeah sure, I can trim it down -- just HTTP1 and Tracing is fine

Sources/AsyncHTTPClient/AsyncAwait/Transaction.swift
ktoso: Thank you, indeed not used init; we're not modifying Transaction in the end design actually

Sources/AsyncHTTPClient/HTTPClient.swift
ktoso: Thank you! Actually I just realized that otel BOTH `http.response.size` and `http.response.body.size` exist, I guess I copied from the wrong row when checking these.

Let's make these only the body ones;

Tests/AsyncHTTPClientTests/HTTPClientTracingTests.swift
```diff
@@ -0,0 +1,150 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the AsyncHTTPClient project authors
```
glbrntt:
```suggestion
// Copyright (c) 2025 Apple Inc. and the AsyncHTTPClient project authors
```

glbrntt (approved):
LGTM modulo one nit

ktoso:
FYI; Hmm, that failure is a warning/as-error in a file this PR didn't touch.

ktoso:
Since everyone involved has generally agreed and we have a formal LGTM as well, I'm going to merge this.

More than happy to work on follow-ups and we can improve what information we're emitting anytime.

PR #862, opened 2025-10-09
[Tracing] Implement trace header context propagation

ktoso:
Somewhere along the way of reviewing previous work this bit got lost;

This implements context propagation by injecting the apropriate headers from service context into the headers when forming a `Prepared` http request.

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient.swift
@@ -1095,7 +1095,11 @@ public final class HTTPClient: Sendable {
         package var attributeKeys: AttributeKeys
 
         public init() {
-            self._tracer = nil
+            if #available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) {
+                self._tracer = InstrumentationSystem.tracer
+            } else {
+                self._tracer = nil
+            }
             self.attributeKeys = .init()
         }
 


Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Prepared.swift
```diff
@@ -45,7 +47,11 @@ extension HTTPClientRequest {
 
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 extension HTTPClientRequest.Prepared {
-    init(_ request: HTTPClientRequest, dnsOverride: [String: String] = [:]) throws {
+    init(
+        _ request: HTTPClientRequest,
+        tracing: HTTPClient.TracingConfiguration? = nil,
```
slashmo: Nitpick: Shouldn't the non-optional dnsOverride argument go before the optional tracing one?

ktoso: Both are defaulted/optional. Could invert if @glbrntt thinks that'd be better tho hmm

slashmo: Isn't `dnsOverride` a non-optional dictionary defaulting to empty? Again, very nitpick and feel free to ignore the comment.

glbrntt: It's internal API so I don't mind either way, and neither requires the caller to set it explicitly so the order doesn't matter. 

However, I would've put `dnsOverride` before `tracing` as the `dnsOverride` is used for "core" functionality (i.e. actually making the request) vs. "extra" functionality (observability).

Tests/AsyncHTTPClientTests/HTTPClientTracingTests.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(Tracing) import AsyncHTTPClient  // NOT @testable - tests that need @testable go into HTTPClientInternalTests.swift
+@_spi(Tracing) @testable import AsyncHTTPClient
```
glbrntt: The previous comment was there for a reason, so that internal API isn't accidentally used in these tests. Can you move the new test?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Prepared.swift
ktoso: ok i'll switch it :)

Tests/AsyncHTTPClientTests/HTTPClientTracingTests.swift
ktoso: Okey, should I perhaps make HTTPClientTracing**Internal**Tests.swift then?

glbrntt: Yeah, that seems like the right place

PR #878, opened 2026-01-08
Add swift-configuration support

hamzahrmalik:
Users should be able to use swift-configuration to create the http client configuration object

Changes
- duplicate package.swift to create a separate version for 6.0 and 6.1, since Configuration is 6.2+
- add helper to create http client configuration using ConfigReader

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/Transaction+StateMachine.swift
@@ -38,6 +38,7 @@ extension Transaction {
             case requestHeadSent
             case producing
             case paused(continuation: CheckedContinuation<Void, Error>?)
+            case endForwarded
             case finished
         }
 
@@ -97,7 +98,8 @@ extension Transaction {
                 bodyStreamContinuation: CheckedContinuation<Void, Error>?
             )
 
-            case failRequestStreamContinuation(CheckedContinuation<Void, Error>, Error)
+            case failRequestStreamContinuation(CheckedContinuation<Void, Error>, Error, HTTPRequestExecutor)
+            case cancelExecutor(HTTPRequestExecutor)
         }
 
         mutating func fail(_ error: Error) -> FailAction {
@@ -135,7 +137,7 @@ extension Transaction {
                         bodyStreamContinuation: continuation
                     )
 
-                case .requestHeadSent, .finished, .producing, .paused(continuation: .none):
+                case .requestHeadSent, .endForwarded, .finished, .producing, .paused(continuation: .none):
                     self.state = .finished(error: error)
                     return .failResponseHead(
                         context.continuation,
@@ -156,12 +158,29 @@ extension Transaction {
                         context.executor,
                         bodyStreamContinuation: bodyStreamContinuation
                     )
-                case .finished, .producing, .requestHeadSent:
+                case .endForwarded, .finished, .producing, .requestHeadSent:
                     return .failResponseStream(source, error, context.executor, bodyStreamContinuation: nil)
                 }
 
-            case .finished(error: _),
-                .executing(_, _, .finished):
+            case .executing(let context, let requestStreamState, .finished):
+                // an error occured after full response received, but before the full request was sent
+                self.state = .finished(error: error)
+                switch requestStreamState {
+                case .paused(let bodyStreamContinuation):
+                    if let bodyStreamContinuation {
+                        return .failRequestStreamContinuation(
+                            bodyStreamContinuation,
+                            error,
+                            context.executor
+                        )
+                    } else {
+                        return .cancelExecutor(context.executor)
+                    }
+                case .endForwarded, .finished, .producing, .requestHeadSent:
+                    return .cancelExecutor(context.executor)
+                }
+
+            case .finished(error: _):
                 return .none
             }
         }
@@ -232,7 +251,7 @@ extension Transaction {
                 self.state = .executing(context, .producing, responseState)
                 return .resumeStream(continuation)
 
-            case .executing(_, .finished, _):
+            case .executing(_, .endForwarded, _), .executing(_, .finished, _):
                 // the channels writability changed to writable after we have forwarded all the
                 // request bytes. Can be ignored.
                 return .none
@@ -254,6 +273,7 @@ extension Transaction {
                 self.state = .executing(context, .paused(continuation: nil), responseSteam)
 
             case .executing(_, .paused, _),
+                .executing(_, .endForwarded, _),
                 .executing(_, .finished, _),
                 .finished:
                 // the channels writability changed to paused after we have already forwarded all
@@ -298,7 +318,7 @@ extension Transaction {
                     "A write continuation already exists, but we tried to set another one. Invalid state: \(self.state)"
                 )
 
-            case .finished, .executing(_, .finished, _):
+            case .finished, .executing(_, .endForwarded, _), .executing(_, .finished, _):
                 return .fail
             }
         }
@@ -309,6 +329,7 @@ extension Transaction {
                 .queued,
                 .deadlineExceededWhileQueued,
                 .executing(_, .requestHeadSent, _),
+                .executing(_, .endForwarded, _),
                 .executing(_, .finished, _):
                 preconditionFailure(
                     "A request stream can only produce, if the request was started. Invalid state: \(self.state)"
@@ -343,6 +364,7 @@ extension Transaction {
             case .initialized,
                 .queued,
                 .deadlineExceededWhileQueued,
+                .executing(_, .endForwarded, _),
                 .executing(_, .finished, _):
                 preconditionFailure("Invalid state: \(self.state)")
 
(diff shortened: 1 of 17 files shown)


Sources/AsyncHTTPClient/HTTPClientConfiguration+SwiftConfiguration.swift
```diff
@@ -0,0 +1,88 @@
+import Configuration
+import NIOCore
+
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+extension HTTPClient.Configuration {
+    public init(configReader: ConfigReader) throws {
```
czechboy0, line 33: Here are some tips on documenting the consumed config keys: https://swiftpackageindex.com/apple/swift-configuration/1.0.0/documentation/configuration/best-practices#Document-configuration-keys

Sources/AsyncHTTPClient/HTTPClientConfiguration+SwiftConfiguration.swift
```diff
@@ -0,0 +1,88 @@
...
+        self.init()
+
+        // Each entry in the list should be a colon separated pair e.g. localhost:127.0.0.1 or localhost:::1
```
czechboy0, line 36: Why not comma-separated?

hamzahrmalik: colon felt more natural to me for defining key-value pairs. I don't mind though

hamzahrmalik: comma would probably make the ipv6 case slightly eaiser to parse too

czechboy0: Oh right, comma might be used for splitting the pairs themselves e.g. in env vars. Hmm then comma isn't great, but colon is a bit confusing.

Sources/AsyncHTTPClient/HTTPClientConfiguration+SwiftConfiguration.swift
czechboy0: This is great, but you could also not flatten the nested keys, with e.g.:

```swift
/// - `redirect` (scoped): Redirect handling read by ``MyRedirectConfiguration/init(config:)``.
```

hamzahrmalik: those inits arent public though. Should they be?

czechboy0: If the types themselves are public then yeah, I think so.

Sources/AsyncHTTPClient/HTTPClientConfiguration+SwiftConfiguration.swift
```diff
@@ -0,0 +1,121 @@
...
+extension HTTPClient.Configuration {
+    /// Initializes HTTPClient configuration from a ConfigReader.
+    ///
+    /// ## Configuration keys:
+    /// - `dnsOverrides` (string array, optional): Colon-separated host:IP pairs for DNS overrides (e.g., "localhost:127.0.0.1").
+    /// - `redirect` (scoped): Redirect handling configuration read by ``RedirectConfiguration/init(configReader:)``.
+    /// - `timeout` (scoped): Timeout configuration read by ``Timeout/init(configReader:)``.
+    /// - `connectionPool` (scoped): Connection pool configuration read by ``ConnectionPool/init(configReader:)``.
+    /// - `httpVersion` (string, optional, default: automatic): HTTP version to use ( "automatic" or  "http1Only").
+    /// - `maximumUsesPerConnection` (int, optional, default: nil, no limit): Maximum uses per connection.
```
FranzBusch, line 29: Should we do camelcase or dot separated? In Temporal we choose the latter https://github.com/apple/swift-temporal-sdk/blob/eb356f25b53d4ac50b1505dea1b7306ba7296bc8/Sources/Temporal/Client/TemporalClient%2BConfiguration.swift#L111

hamzahrmalik: I used dots for hierarchy but camel case for words

hamzahrmalik: dns.overrides and http.version make sense in that system. But maximum.uses.per.connection does not

FranzBusch: This could be `connection.limits.maximum.uses`. It is not the end of the world but it would be great if we could be a bit consistent around the ecosystem. @czechboy0 WDYT?

hamzahrmalik: oh yeah i like that idea

czechboy0: I've been doing the same as Hamzah - use dots for separating components, with individual components being camelCase. So `http.maxConnectionCount`, for example.

FranzBusch: Should we document this somewhere? It feels like that would be good to have a recommendation at least for the ecosystem.

czechboy0: https://github.com/apple/swift-configuration/issues/145

Feel free to add more questions to the issue, will add docs on this

FranzBusch (approved):

Tests/AsyncHTTPClientTests/SwiftConfigurationTests.swift
```diff
@@ -0,0 +1,261 @@
...
+        #expect(config.dnsOverride["localhost"] == "::1")
+        #expect(config.dnsOverride["example.com"] == "2001:db8::1")
+    }
+
+    @Test
+    @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+    func dnsOverridesWithInvalidFormatIgnored() throws {
+        let testProvider = InMemoryProvider(values: [
+            "dnsOverrides": .init(.stringArray(["invalidentry", "localhost:127.0.0.1"]), isSecret: false)
+        ])
```
fabianfett, line 164: what about spaces:
```
localhost: 127.0.0.1
```
Are we fine with the silent error here?

Tests/AsyncHTTPClientTests/SwiftConfigurationTests.swift
```diff
@@ -0,0 +1,261 @@
...
+        let testProvider = InMemoryProvider(values: [
+            "dnsOverrides": .init(.stringArray(["api.example.com:10.0.0.1"]), isSecret: false)
+        ])
+
+        let configReader = ConfigReader(provider: testProvider)
+        let config = try HTTPClient.Configuration(configReader: configReader)
+
+        #expect(config.dnsOverride["api.example.com"] == "10.0.0.1")
+        #expect(config.dnsOverride.count == 1)
+    }
```
fabianfett: what's the point of this test? what is tested here, that the other tests miss?

Package@swift-6.0.swift
```diff
@@ -0,0 +1,124 @@
+// swift-tools-version:6.0
```
fabianfett, line 1: Is there a way without duplicating, all the Package.swifts?

Tests/AsyncHTTPClientTests/SwiftConfigurationTests.swift
hamzahrmalik: i will make it trim spaces, i think that's going to be the least surprising behaviour

Tests/AsyncHTTPClientTests/SwiftConfigurationTests.swift
hamzahrmalik: it's not needed. i think i wrote this one first but the other ones cover this and more so i'll remove it

Tests/AsyncHTTPClientTests/SwiftConfigurationTests.swift
hamzahrmalik: do we think it's preferable to throw an error on invalid config? It would make it much more clear i suppose

FranzBusch: Yeah I think throwing an error if the value of a key isn't valid is very reasonable

Package@swift-6.0.swift
hamzahrmalik: not that i know of

Tests/AsyncHTTPClientTests/SwiftConfigurationTests.swift
hamzahrmalik: done

Sources/AsyncHTTPClient/HTTPClientConfiguration+SwiftConfiguration.swift
```diff
@@ -0,0 +1,140 @@
...
+extension HTTPClient.Configuration.RedirectConfiguration {
+    /// Initializes redirect configuration from a ConfigReader.
+    ///
+    /// ## Configuration keys:
+    /// - `mode` (string, optional, default: "follow"): Redirect handling mode ("follow" or "disallow").
+    /// - `maxRedirects` (int, optional, default: 5): Maximum allowed redirects when mode is "follow".
+    /// - `allowCycles` (bool, optional, default: false): Allow cyclic redirects when mode is "follow".
+    ///
+    /// - Throws: `HTTPClientError.invalidRedirectConfiguration` if mode is specified but invalid.
+    public init(configReader: ConfigReader) throws {
```
fabianfett, line 71: I'm not sure about the developer experience, here I think it would be better to keep this method internal and doc all the keys in the OG init. WDYT?

fabianfett: @czechboy0 do you have any guidance for this?

czechboy0: We just have this: https://swiftpackageindex.com/apple/swift-configuration/main/documentation/configuration/configuring-libraries

But it doesn't go into what to do with nested structs.

My rule of thumb would be: if the nested structs are public API, and you can construct them using public API, I think the config-based initializer should then also be public. Otherwise we move those docs onto a parent struct that isn't easy to find unless you already know where to find it. So I think it's just easier to follow this way.

Issue #860, opened 2025-10-08
[Tracing] Add more span attributes


ktoso:
We could further refine how many span attributes we set, currently only the bare minimum of request method is set.

czechboy0:
As I'm starting to use this in practice, I agree we need to add more. Some for consideration, based on what a well-instrumented server reports (we should mirror the similar attributes on the client side):
- `http.request.body.size`
- `http.response.body.size`
- `http.response.status_code` (right now we set the legacy `http.status_code`)
- `network.peer.address`
- `network.peer.port`
- `network.protocol.version`
- `url.path`
- `url.scheme`
- `server.hostname`
- `server.port`

PR #881, opened 2026-01-21
Add more span attributes (URL, network)

candiun:
This PR adds some span attributes mentioned in https://github.com/swift-server/async-http-client/issues/860

- `http.response.body.size`
- changes `http.status_code` -> `http.response.status_code`
- `network.protocol.version`
- `url.path`
- `url.scheme`
- `server.hostname`
- `server.port`
- `http.request.header`
- `http.response.header`


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1111,7 +1111,15 @@ public final class HTTPClient: Sendable {
...
-            @usableFromInline package var responseStatusCode: String = "http.status_code"
+            @usableFromInline package var responseStatusCode: String = "http.response.status_code"
+
+            @usableFromInline package var networkProtocolVersion: String = "network.protocol.version"
+
+            @usableFromInline package var urlPath: String = "url.path"
+            @usableFromInline package var urlScheme: String = "url.scheme"
+
+            @usableFromInline package var serverHostname: String = "server.hostname"
+            @usableFromInline package var serverPort: String = "server.port"
```
czechboy0, line 1159: I wonder if we should just take the direct dependency on https://github.com/swift-otel/swift-otel-semantic-conventions

candiun: This sounds like a great addition - I've added it. Please let me know (you and the maintainers) if the implementation is fine.

I've also added request and response header attributes to the span.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
```diff
@@ -27,13 +28,32 @@ extension HTTPClient {
         }
 
         return try await tracer.withSpan(request.method.rawValue, ofKind: .client) { span in
-            let keys = self.configuration.tracing.attributeKeys
-            span.attributes[keys.requestMethod] = request.method.rawValue
-            // TODO: set more attributes on the span
+            span.attributes.http.request.method = .init(rawValue: request.method.rawValue)
+
+            // set request headers
+            for header in request.headers {
```
czechboy0, line 43: Does OTel recommend redacting any headers? For example, should this include `Authorization`/`Cookie` headers?

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
```diff
@@ -27,13 +28,32 @@ extension HTTPClient {
...
+                span.attributes.http.request.header.set(header.name, to: [header.value])
+            }
+
+            // set url attributes
+            if let deconstructedURL = try? DeconstructedURL(url: request.url) {
+                span.attributes.url.path = deconstructedURL.uri
```
czechboy0: The query (part of the path, I believe here?) could also include a secret token, but it's tricky to try to filter it out as it can have any name.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
```diff
@@ -27,13 +28,32 @@ extension HTTPClient {
...
+                span.attributes.server.port = deconstructedURL.connectionTarget.port
+            }
+
             let response = try await body()
 
             // set response span attributes
-            TracingSupport.handleResponseStatusCode(span, response.status, keys: tracing.attributeKeys)
+            TracingSupport.handleResponseStatusCode(span, response.status)
+
+            for header in response.headers {
```
czechboy0, line 90: Same question as on the request side, I wonder if we should redact or remove `Set-Cookie` here.

czechboy0 (commented):
Added a few suggestions, but otherwise it's great to see more attributes added here.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
candiun: Yes and thanks for the question. It seems like OTel does not recommend capturing anything by default

> Instrumentations SHOULD require an explicit configuration of which headers are to be captured. Including all request headers can be a security risk - explicit configuration helps avoid leaking sensitive information.

I have changed the implementation to an opt-in array residing inside TracingConfiguration.

One thing I am unsure about is how these headers would be configured in frameworks like Vapor for instance where a default client is supplied in the `Request` object. Maybe in that case the configuration would be easier once swift-configuration is adopted? For newly instantiated clients that should be easier, though.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
candiun: I have changed the implementation to redact query names based on a list inside `TracingConfiguration`. 

OTel says:

```
Query string values for the following keys SHOULD be redacted by default and replaced by the value REDACTED:

[AWSAccessKeyId](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth)
[Signature](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth)
[sig](https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token)
[X-Goog-Signature](https://cloud.google.com/storage/docs/access-control/signed-urls)
```

So I have added them by default to the redaction list, with the possibility to extend it upon creating a client.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
candiun: Response headers are treated now the same way as request headers and are filtered through an allow list. I wasn't sure whether keeping two separate (one for request, other for response) would be beneficial, so I stayed with one

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1070,6 +1070,15 @@ public final class HTTPClient: Sendable {
         @usableFromInline
         var _tracer: Optional<any Sendable>  // erasure trick so we don't have to make Configuration @available
 
+        public var allowedHeaders: Set<String> = []
```
czechboy0, line 1097: Allowed headers is one way, though I wonder if it should be "redacted headers" instead, similar to the path and query items below.

candiun: This would work well when it comes to request headers. But I'm thinking if for the cases like the response headers, where we're sometimes communicating with an external services that we have no control over, whether this wouldn't lead to leaking some secrets of custom names, either at the current time, or in case that service evolves and starts sending other headers in response. Maybe in that case ignoring everything unless explicitly allowed would be safer, what do you think?

candiun:
Would be happy to make any changes to move forward with the PR

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1089,33 +1098,12 @@ public final class HTTPClient: Sendable {
...
-        package struct AttributeKeys: Sendable {
-            @usableFromInline package var requestMethod: String = "http.request.method"
-            @usableFromInline package var requestBodySize: String = "http.request.body.size"
-
-            @usableFromInline package var responseBodySize: String = "http.response.body.size"
-            @usableFromInline package var responseStatusCode: String = "http.status_code"
-
-            @usableFromInline package var httpFlavor: String = "http.flavor"
-
-            @usableFromInline package init() {}
```
fabianfett, line 1142: I'm not a fan of those options and their use being removed. The goal here is to allow users to change the attribute keys.

Sources/AsyncHTTPClient/TracingSupport.swift
```diff
@@ -12,11 +12,13 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Foundation
 import Logging
 import NIOConcurrencyHelpers
 import NIOCore
 import NIOHTTP1
 import NIOSSL
+import OTelSemanticConventions
```
fabianfett: I'm not a fan of adding another package dependency just to import some Strings. As the OTEL strings should never change, we can just have those strings hardcoded here in this package.

fabianfett (changes requested):
I'm fine with adding the additional tracing attributes. However I'm not a fan of changing the previously agreed upon approach to tracing.

Sources/AsyncHTTPClient/TracingSupport.swift
candiun: I've removed the dependency and brought back the keys

Sources/AsyncHTTPClient/TracingSupport.swift
```diff
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Foundation
```
fabianfett: I think we had a change so that we only rely on FoundationEssentials.

Tests/AsyncHTTPClientTests/HTTPClientTracingAttributeTests.swift
```diff
@@ -0,0 +1,305 @@
...
+import NIOTestUtils
+import NIOTransportServices
+import Tracing
+import XCTest
+
+#if canImport(Network)
+import Network
+#endif
+
+final class HTTPClientTracingAttributeTests: XCTestCaseHTTPClientTestsBaseClass {
```
fabianfett: for new test files, lets use Swift Testing.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
```diff
@@ -29,12 +30,64 @@ extension HTTPClient {
...
+            // set explicitly allowed request headers
+            let allowedRequestHeaderNames = Set(request.headers.map(\.name)).intersection(configuration.tracing.allowedHeaders)
+
+            for headerName in allowedRequestHeaderNames {
+                let values = request.headers[headerName]
+
+                if !values.isEmpty {
+                    span.attributes["\(keys.requestHeader).\(headerName)"] = values
+                }
+            }
```
fabianfett: That is quite expensive. We should reduce this down to a loop and a O(1) lookup in the `allowedHeaders` list.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+tracing.swift
```diff
@@ -29,12 +30,64 @@ extension HTTPClient {
...
+
             let response = try await body()
 
             // set response span attributes
             TracingSupport.handleResponseStatusCode(span, response.status, keys: tracing.attributeKeys)
 
+            // set explicitly allowed response headers
+            let allowedResponseHeaderNames = Set(response.headers.map(\.name)).intersection(configuration.tracing.allowedHeaders)
```
fabianfett: See comment above, re making this less computational heavy.

mapedd:
Is there anything left to be added here? Would be great to see this merged, very nice improvement! Great work @candiun !

PR #882, opened 2026-01-23
Support sending and receiving trailers in HTTPExecutableRequest

fabianfett:
This should be landed after #879.

---

This pr adds all the internal wiring in `HTTP1ClientChannelHandler` and `HTTP2ClientRequestHandler` to send and receive HTTP trailers.

### DIFF
--- a/Tests/AsyncHTTPClientTests/HTTP1ClientChannelHandlerTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTP1ClientChannelHandlerTests.swift
@@ -908,7 +908,7 @@ class HTTP1ClientChannelHandlerTests: XCTestCase {
         let channel = EmbeddedChannel(handlers: [handler], loop: eventLoop)
         XCTAssertNoThrow(try channel.connect(to: .init(ipAddress: "127.0.0.1", port: 80)).wait())
 
-        // non empty body is important to trigger this bug as we otherwise finish the request in a single flush
+        // non empty body is important to allow sending trailers as the request is finished in a single flush otherwise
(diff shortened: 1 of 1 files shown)


Tests/AsyncHTTPClientTests/HTTP1ClientChannelHandlerTests.swift
```diff
@@ -897,6 +897,55 @@ class HTTP1ClientChannelHandlerTests: XCTestCase {
...
+        let eventLoop = EmbeddedEventLoop()
+        let handler = HTTP1ClientChannelHandler(
+            eventLoop: eventLoop,
+            backgroundLogger: Logger(label: "no-op", factory: SwiftLogNoOpLogHandler.init),
+            connectionIdLoggerMetadata: "test connection"
+        )
+        let channel = EmbeddedChannel(handlers: [handler], loop: eventLoop)
+        XCTAssertNoThrow(try channel.connect(to: .init(ipAddress: "127.0.0.1", port: 80)).wait())
+
+        // non empty body is important to trigger this bug as we otherwise finish the request in a single flush
```
glbrntt: which bug?

fabianfett:
> Could you add an end-to-end test as well?

The API does not allow for E2E tests yet. This will come later, when we have more capability inside AHC.

Tests/AsyncHTTPClientTests/HTTP1ClientChannelHandlerTests.swift
fabianfett: comment updated.

PR #883, opened 2026-02-02
Add SPKI-based certificate pinning support

o-nnerb:
This PR introduces **SPKI-based certificate pinning** to `AsyncHTTPClient`, enabling clients to enforce explicit trust in server certificates beyond standard PKI validation. This provides protection against compromised Certificate Authorities and MITM attacks.

### 🔑 Key features
- **SPKI pinning** (not full certificate pinning) — survives legitimate certificate rotations by pinning the public key structure (RFC 5280, Section 4.1)
- **Two verification modes**: 
  - `.strict` (production): blocks connections on pin mismatch
  - `.audit` (staging): allows connections but logs warnings for observability
- **Multi-algorithm support**: SHA-256 (default), SHA-384, and SHA-512 via generic init `SPKIHash(algorithm:*:)`
- **Constant-time comparison**: Resistant to timing attacks during pin validation
- **Seamless integration**: Works alongside existing TLS configuration (OpenSSL/BoringSSL backend only)

### 🛡️ Security by design
- **Rotation safety**: Emits runtime warning when `< 2 pins` are configured in `.strict` mode (caller-managed redundancy)
- **No silent failures**: Explicit `HTTPClientError.invalidCertificatePinning` on mismatch
- **OWASP/NIST aligned**: Compliant with MSTG-NETWORK-4 and NIST SP 800-52 Rev. 2 Section 3.4.3

### 🧪 Usage example
```swift
let config = HTTPClient.Configuration(
    tlsPinning: SPKIPinningConfiguration(
        pins: [
            SPKIHash(base64: "dJEGUdm2BuEalPybF+8enWB7R4AeiwE0gsQHyQrhzhY="),
            SPKIHash(base64: "backup-pin-for-rotation") // Redundancy required for safe rotation
        ],
        policy: .strict
    )
)

let client = HTTPClient(configuration: config)
```

### ⚠️ Important notes
- **Backend requirement**: Pinning requires OpenSSL/BoringSSL backend — Network.framework (NIOTS) connections ignore pinning configuration
- **Rotation safety**: Always deploy ≥2 pins in production (current + upcoming certificates) to avoid lockout during rotation
- **Hash generation**: Pins must be Base64-encoded hashes of the DER-encoded SPKI structure:
  ```bash
  openssl s_client -connect example.com:443 -servername example.com 2>/dev/null | \
    openssl x509 -pubkey -noout | \
    openssl pkey -pubin -outform der | \
    openssl dgst -sha256 -binary | \
    openssl base64 -A
  ```

### 📚 References
- [OWASP MSTG: Network Communication](https://mas.owasp.org/MASTG/0x06g-Testing-Network-Communication)
- [NIST SP 800-52 Rev. 2, Section 3.4.3](https://csrc.nist.gov/publications/detail/sp/800-52/rev-2/final)
- [RFC 5280, Section 4.1 (SPKI)](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1)


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -393,6 +396,15 @@ extension HTTPConnectionPool.ConnectionFactory {
                         serverHostname: sslServerHostname
                     )
                     try channel.pipeline.syncOperations.addHandler(sslHandler)
+
+                    if #available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *), let tlsPinning {
```
Lukasa: It's very dangerous to silently ignore TLS pinning configuration. We need to refuse to tolerate TLS pinning configuration in contexts where we cannot actually enforce it.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -596,7 +608,17 @@ extension HTTPConnectionPool.ConnectionFactory {
                             let tlsEventHandler = TLSEventsHandler(deadline: deadline)
 
                             try sync.addHandler(sslHandler)
+
+                            if #available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *), let tlsPinning {
```
Lukasa: Same availability note here.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,316 @@
...
+    // MARK: - Initialization
+
+    /// Creates an SPKI hash from a base64-encoded SHA-256 digest.
+    ///
+    /// - Parameters:
+    ///   - base64: Base64-encoded hash digest (whitespace is stripped).
+    ///
+    /// - Throws: `HTTPClientError.invalidDigestLength` if decoded data isn't 32 bytes.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    public init(base64: String) throws {
```
Lukasa: This initializer strikes me as a bit too easy to misuse. Is it necessary?

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /// Creates an SPKI hash from raw SHA-256 digest bytes.
+    ///
+    /// - Parameters:
+    ///   - bytes: Raw SHA-256 digest bytes (must be 32 bytes).
+    ///
+    /// - Throws: `HTTPClientError.invalidDigestLength` if byte count isn't 32.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    public init(bytes: Data) throws {
```
Lukasa: Similarly, this seems a bit easy to misuse as well.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,316 @@
...
+}
+
+/// Constant-time comparison to prevent timing attacks.
+internal func constantTimeAnyMatch(_ target: Data, _ candidates: [SPKIHash]) -> Bool {
+    guard !candidates.isEmpty else { return false }
+
+    var anyMatch: UInt8 = 0
+    for candidate in candidates {
+        var diff: UInt8 = 0
+        for (a, b) in zip(target, candidate.bytes) {
```
Lukasa: This incorrectly truncates digests. We need a length-extending `zip`, not a length-truncating one, and we need to include the lengths in the comparison.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,316 @@
...
+    ///   - activePins: Hashes of currently deployed certificates.
+    ///   - backupPins: Hashes for upcoming certificate rotations (required in production).
+    ///   - policy: Validation failure policy (`.strict` for production, `.audit` for debugging).
+    ///
+    /// - Warning: Empty `backupPins` in `.strict` mode risks catastrophic lockout during
+    ///   certificate rotation.
+    public init(
+        activePins: [SPKIHash],
+        backupPins: [SPKIHash],
+        policy: SPKIPinningPolicy = .strict
```
Lukasa: The active vs backup pins distinction doesn't appear to be meaningful: they're both equally valid. Probably better to remove the distinction entirely.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,316 @@
...
+            let computedHash = first.hash(spkiData)
+            let isMatch = constantTimeAnyMatch(computedHash, hashes)
+            anyMatch |= isMatch ? 1 : 0
+        }
+        return anyMatch != 0
+    }
+}
+
+/// Policy for handling SPKI pin validation failures.
+public enum SPKIPinningPolicy: Sendable, Hashable {
```
Lukasa: Let's avoid new `public enum`s: prefer to make this a struct.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,316 @@
...
+            )
+        }
+    }
+
+    func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        guard
+            let tlsEvent = event as? TLSUserEvent,
+            case .handshakeCompleted = tlsEvent
+        else {
+            context.fireUserInboundEventTriggered(event)
```
Lukasa: This should still be fired even if we do end up handling the rest of the body.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -909,6 +912,7 @@ public final class HTTPClient: Sendable {
 
         public init(
             tlsConfiguration: TLSConfiguration? = nil,
+            tlsPinning: SPKIPinningConfiguration? = nil,
```
Lukasa: We can't add new parameters to an `init` like this: they have to have a new `init` instead.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -929,6 +934,7 @@ public final class HTTPClient: Sendable {
...
```
Lukasa: same here.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -948,6 +955,7 @@ public final class HTTPClient: Sendable {
 
         public init(
             certificateVerification: CertificateVerification,
+            tlsPinning: SPKIPinningConfiguration? = nil,
```
Lukasa: Same here.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -970,6 +979,7 @@ public final class HTTPClient: Sendable {
...
```
Lukasa: Same here.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -993,6 +1004,7 @@ public final class HTTPClient: Sendable {
...
```
Lukasa: Same here.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1012,6 +1025,7 @@ public final class HTTPClient: Sendable {
 
         public init(
             tlsConfiguration: TLSConfiguration? = nil,
+            tlsPinning: SPKIPinningConfiguration? = nil,
```
Lukasa: Same here.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1038,6 +1053,7 @@ public final class HTTPClient: Sendable {
...
```
Lukasa: Same here.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -274,13 +285,21 @@ extension HTTPClient {
             method: HTTPMethod = .GET,
             headers: HTTPHeaders = HTTPHeaders(),
             body: Body? = nil,
-            tlsConfiguration: TLSConfiguration?
+            tlsConfiguration: TLSConfiguration?,
+            tlsPinning: SPKIPinningConfiguration?
```
Lukasa: We can't add new parameters here either.

Sources/AsyncHTTPClient/HTTPHandler.swift
```diff
@@ -318,7 +345,8 @@ extension HTTPClient {
...
```
Lukasa: Or here.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: Actually, I’ll take that off.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: You're right! I'll remove this one too.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
o-nnerb: Ah, good point. I’ll make sure it fails explicitly in that case. Thanks!

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
o-nnerb: Got it, will apply the same fix here.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: You’re absolutely right. Applied changes.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: Alright, I’ve made the change to the struct as requested. Could you please review it again?

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: Ah, thanks for bringing this to my attention. I’ve made some changes to prevent any future misunderstandings. I believe it’s correct now. Could you please check it again?

Sources/AsyncHTTPClient/HTTPClient.swift
o-nnerb: My apologies for the oversight 😅. I’ve removed the previous implementation and replaced it with a new one that uses the one with the most parameters.

Sources/AsyncHTTPClient/HTTPClient.swift
o-nnerb: done

Sources/AsyncHTTPClient/HTTPClient.swift
o-nnerb: Done.

Sources/AsyncHTTPClient/HTTPClient.swift
o-nnerb: Done.

Sources/AsyncHTTPClient/HTTPClient.swift
o-nnerb: Done.

Sources/AsyncHTTPClient/HTTPClient.swift
o-nnerb: Done.

Sources/AsyncHTTPClient/HTTPClient.swift
o-nnerb: Done.

Sources/AsyncHTTPClient/HTTPHandler.swift
o-nnerb: Done. Thanks!

Sources/AsyncHTTPClient/HTTPHandler.swift
o-nnerb: My apologies. I’ve removed it. Thanks.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: Fixed! Now using a length-extending comparison that checks full digest lengths before element-wise equality.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -114,35 +88,40 @@ public struct SPKIHash: Sendable, Hashable {
...
     var anyMatch: UInt8 = 0
     for candidate in candidates {
-        var diff: UInt8 = 0
-        for (a, b) in zip(target, candidate.bytes) {
+        var diff: UInt8 = (target.count == candidate.bytes.count) ? 0 : 1
+
+        let maxLength = max(target.count, candidate.bytes.count)
+        for i in 0 ..< maxLength {
+            let a = i < target.count ? target[i] : 0
+            let b = i < candidate.bytes.count ? candidate.bytes[i] : 0
```
Lukasa: This doesn't quite meet the criteria either: the iteration time will speed up once one of these containers has reached its end. The thing you actually need is a wrapping index after:

```
extension Collection {
    func wrappingIndex(after index: Index) -> Index {
        let nextCandidate = self.index(after: index)
        if nextCandidate == self.endIndex {
            return self.startIndex
        } else {
            return nextCandidate
        }
    }
}
```

Then turn this into a loop that iterates over indices.

The reason this works is that each loop iteration will validly subscript both collections, and the saturatingIndexAfter should perform the same amount of work no matter what.

There is an important caveat here: the above operation doesn't work correctly on zero-length collections. In this case we can safely take a shortcut and detect if either operand has zero length and return immediately, as it is not possible for a comparison with a zero-length collection to leak anything about the data being compared.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -184,37 +157,58 @@ public struct SPKIPinningConfiguration: Sendable, Hashable {
...
+        return name
+    }
+
+    private let name: String
+    private let rawValue: UInt8
+
+    private init(name: String, rawValue: UInt8) {
+        self.name = name
+        self.rawValue = rawValue
+    }
```
Lukasa: This works better if it wraps an enum. That way we don't have to copy a string around.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -229,88 +223,130 @@ final class SPKIPinningHandler: ChannelInboundHandler, RemovableChannelHandler {
             return
         }
 
+        // ⚠️ Security-critical: handshake propagation is delayed until validation completes
         context.pipeline.handler(type: NIOSSLHandler.self).assumeIsolated().whenComplete {
```
Lukasa: Looks like we're still not using `context.pipeline.syncOperations`, which will make life a lot easier.

Tests/AsyncHTTPClientTests/SPKIPinningTests.swift
```diff
@@ -309,15 +277,26 @@ class SPKIPinningTests: XCTestCase {
     }
 }
 
-private final class ContextHandler: ChannelInboundHandler {
-    typealias InboundIn = NIOAny
-    let context: (ChannelHandlerContext) -> Void
+final class MockSPKIPinningExecutor: SPKIPinningExecutor {
```
Lukasa: I can find no evidence that this is ever used.

Lukasa: If it's not necessary, we can remove the `SPKIPinningExecutor` protocol as well, because it only exists to enable this type.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: Hmm, I wasn’t aware of that existence. I’ll replace it with that.

Tests/AsyncHTTPClientTests/SPKIPinningTests.swift
o-nnerb: I recently implemented this to make the tests cleaner, but I’ll remove it. I also didn’t like the solution very much.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: Thanks for pointing this out, I wasn't aware of this behavior. I'll adjust based on your feedback. If it's still incorrect, feel free to suggest a pre-existing algorithm. There's even a function like this built into `NIOSSL`:

> swift-nio-ssl/Sources/NIOSSL/SwiftCrypto/SafeCompare.swift

```swift
/// A straightforward constant-time comparison function for any two collections of bytes.
@inlinable
internal func constantTimeCompare<LHS: Collection, RHS: Collection>(_ lhs: LHS, _ rhs: RHS) -> Bool
where LHS.Element == UInt8, RHS.Element == UInt8 {
    guard lhs.count == rhs.count else {
        return false
    }

    return zip(lhs, rhs).reduce(into: 0) { $0 |= $1.0 ^ $1.1 } == 0
}
```

Tests/AsyncHTTPClientTests/SPKIPinningTests.swift
o-nnerb: In the end I wasn't even using it 😅

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: Wrapped the policy in a type-safe enum-based structure. Ready for another look, does this align with what you had in mind?

Sources/AsyncHTTPClient/HTTPHandler.swift
o-nnerb: Apologies for this oversight. I’ve forgotten to revert this. 😅

o-nnerb:
@fabianfett, could you please explain the linkage error and its relation to this implementation? I’m unable to identify the specific change that would resolve this issue.

o-nnerb:
~@fabianfett, I’ve added `traits: []` to `Package.swift` as per the comment on this file. The comment suggests disabling all traits to prevent linking Foundation. If this change is correct, could you please rerun the CI check?~

Actually, this doesn't work 😩

o-nnerb:
@Lukasa @fabianfett any updates on this?

o-nnerb:
@Lukasa @fabianfett, could you please run CI again? I believe that based on the recent updates to swift-crypto, everything should be set now.

o-nnerb:
@Lukasa @fabianfett, could you please run the CI check again? Or at the very least, provide any updates on this?

PR #887, opened 2026-02-16
Add option to retain request method on 301/302/303 redirects

hamzahrmalik:
Add a configuration option to retain the HTTP method and body receiving 301 or 302 responses.

Currently we automatically change the method to GET, and remove the body, before following a 301 or 302. This is compliant with the fetch specification: https://fetch.spec.whatwg.org/#http-redirect-fetch

However, it is useful to be able to override this behaviour and retain the method and body.

Changes
- Add a new struct to encapsulate the (now 4) arguments of the follow case of the redirect mode
- Add new options `retainHTTPMethodAndBodyOn301` and `retainHTTPMethodAndBodyOn302`. Defaults to false because thats the existing behaviour today
- When it is true, do not convert requests to GET after following a redirect
- Note: this does not affect 307/308 (or any other) redirects. They always preserve their method

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClient+execute.swift
@@ -141,7 +141,7 @@ extension HTTPClient {
                 from: preparedRequest.url,
                 to: redirectURL,
                 status: response.status,
-                config: redirectState.config,
+                config: redirectState.config
             )
 
             guard newRequest.body.canBeConsumedMultipleTimes else {
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Prepared.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest+Prepared.swift
@@ -134,7 +134,7 @@ extension HTTPClientRequest {
             body: self.body,
             to: redirectURL,
             status: status,
-            config: config,
+            config: config
         )
         var newRequest = HTTPClientRequest(url: redirectURL.absoluteString)
         newRequest.method = method


Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1269,13 +1269,35 @@ extension HTTPClient.Configuration {
...
+            public var allowCycles: Bool
+            /// Whether to convert POST requests into GET requests when following a 301, 302 or 303 redirect.
+            /// This does not apply to 307 or 308. Those always retain the original method.
+            public var convertToGet: Bool
+
+            /// Create a new ``FollowConfiguration``
+            /// - Parameters:
+            ///   - max: The maximum number of allowed redirects.
+            ///   - allowCycles: Whether cycles are allowed.
+            ///   - convertToGet: Whether to convert POST requests into GET requests when following a 301, 302 or 303 redirect. This does not apply to 307 or 308. Those always retain the original method.
```
fabianfett: I think we should call out that this should be `true` except, if users have clear needs.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1293,7 +1315,16 @@ extension HTTPClient.Configuration {
...
         public static func follow(max: Int, allowCycles: Bool) -> RedirectConfiguration {
-            .init(configuration: .follow(max: max, allowCycles: allowCycles))
+            .follow(configuration: .init(max: max, allowCycles: allowCycles, convertToGet: true))
+        }
+
+        /// Redirects are followed with a specified limit.
+        ///
+        /// - parameters
+        ///
+        /// - warning: Cycle detection will keep all visited URLs in memory which means a malicious server could use this as a denial-of-service vector.
```
fabianfett: This comments looks weird.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1269,13 +1269,35 @@ extension HTTPClient.Configuration {
...
+
+        /// Configuration for following redirects.
+        public struct FollowConfiguration: Hashable, Sendable {
+            /// The maximum number of allowed redirects.
+            public var max: Int
+            /// Whether cycles are allowed.
+            public var allowCycles: Bool
+            /// Whether to convert POST requests into GET requests when following a 301, 302 or 303 redirect.
+            /// This does not apply to 307 or 308. Those always retain the original method.
+            public var convertToGet: Bool
```
fabianfett: I like that curls specifically calls out 301, 302 and 303. I wonder if we should do the same here? `convertToGet` sounds too broad for me.

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -1269,13 +1269,35 @@ extension HTTPClient.Configuration {
             /// Redirects are not followed.
             case disallow
             /// Redirects are followed with a specified limit.
-            case follow(max: Int, allowCycles: Bool)
+            case follow(FollowConfiguration)
+        }
+
+        /// Configuration for following redirects.
+        public struct FollowConfiguration: Hashable, Sendable {
```
fabianfett, line 1276: I wonder if it is worth adding this new struct now? If I see this correctly, we don't pass it around at all. Lets just add another property on the enum for now.

fabianfett (commented):
Love the tests. Let the naming bike-shed begin.

Sources/AsyncHTTPClient/HTTPClient.swift
hamzahrmalik: i copied it from the existing one above

Sources/AsyncHTTPClient/HTTPClient.swift
hamzahrmalik: given that we're doing 3 properties now, imo we should keep the struct. We can actually take advantage of it by holding the struct inside the RedirectState

hamzahrmalik: see last commit

Sources/AsyncHTTPClient/HTTPClientConfiguration+SwiftConfiguration.swift
```diff
@@ -66,6 +66,8 @@ extension HTTPClient.Configuration.RedirectConfiguration {
     /// - `mode` (string, optional, default: "follow"): Redirect handling mode ("follow" or "disallow").
     /// - `maxRedirects` (int, optional, default: 5): Maximum allowed redirects when mode is "follow".
     /// - `allowCycles` (bool, optional, default: false): Allow cyclic redirects when mode is "follow".
+    /// - `retainHTTPMethodAndBodyOn301` (bool, optional, default: false): Retain the original HTTP method and body on 301 redirect when mode is "follow". This is contrary to the fetch specification, but is allowed by RFC 9110.
+    /// - `retainHTTPMethodAndBodyOn302` (bool, optional, default: false): Retain the original HTTP method and body on 302 redirect when mode is "follow". This is contrary to the fetch specification, but is allowed by RFC 9110.
```
fabianfett: Lets call out POST here explicitly. Because this is just about POST.

PR #904, opened 2026-05-19
Extend swift-configuration support to proxy configuration

hamzahrmalik:
Extending #878 to also allow configuring the proxy via ConfigReader


aryan-25 (approved):
LGTM, just one nit.

Sources/AsyncHTTPClient/HTTPClientConfiguration+SwiftConfiguration.swift
```diff
@@ -148,4 +153,83 @@ extension HTTPClient.Configuration.HTTPVersion {
...
+    /// ## Configuration keys:
+    /// - `enabled` (bool, optional, default: false): Whether proxy support is enabled. If `false`, the initializer returns `nil`.
+    /// - `host` (string): Proxy server host. Required when `enabled` is `true`.
+    /// - `port` (int): Proxy server port. Required for `http` proxies; defaults to `1080` for `socks` proxies.
+    /// - `type` (string, optional, default: "http"): Proxy type ("http" or "socks").
+    /// - `authorization` (scoped, optional): Authorization configuration read by ``HTTPClient/Authorization/init(configReader:)``.
+    ///   Only supported for `http` proxies.
+    ///
+    /// - Throws: `HTTPClientError.invalidProxyConfiguration` if `enabled` is `true` but `host` is missing, `type` is unknown,
+    ///   `port` is missing for an HTTP proxy, or authorization is specified for a SOCKS proxy.
```
aryan-25: nit:
```suggestion
    ///   `port` is missing for an HTTP proxy, or `authorization` is specified for a SOCKS proxy.
```

Also, I know the `HTTPClient.Authorization` config init has this documented, but it might be worth calling out here that this error is also thrown when the provided `authorization` for an HTTP proxy is invalid.

Sources/AsyncHTTPClient/HTTPClientConfiguration+SwiftConfiguration.swift
```diff
@@ -148,4 +153,83 @@ extension HTTPClient.Configuration.HTTPVersion {
...
+    /// - `authorization` (scoped, optional): Authorization configuration read by ``HTTPClient/Authorization/init(configReader:)``.
+    ///   Only supported for `http` proxies.
+    ///
+    /// - Throws: `HTTPClientError.invalidProxyConfiguration` if `enabled` is `true` but `host` is missing, `type` is unknown,
+    ///   `port` is missing for an HTTP proxy, or `authorization` is specified for a SOCKS proxy, or `authorization` is invalid (see ``HTTPClient/Authorization/init(configReader:)``)
+    public init?(configReader: ConfigReader) throws {
+        guard configReader.bool(forKey: "enabled", default: false) else {
+            return nil
+        }
+        guard let host = configReader.string(forKey: "host") else {
```
glbrntt: Are you deliberately not using `requiredString(forKey:)` so that you can thrown an `HTTPClientError`? (Ignore me if I'm misremembering the API.)

Minor concern is that `invalidProxyConfiguration` doesn't tell you what you got wrong, only that you got something wrong. I assume the `swift-configuration` error would tell you what required key was missing.

hamzahrmalik: there's no precedent within this file, because we dont have any required props until now

There is invalidHTTPVersionConfiguration and invalidDNSOverridesConfiguration but these are for invalid values rather than missing values.

I think youre right and using the requiredString etc functions makes more sense

hamzahrmalik: you do still need the custom error for Authentication though, since we need one of several keys to exist

glbrntt: Ack. Let's just make sure the error is clear about what the requirements are for Authentication

glbrntt (approved):

PR #906, opened 2026-05-22
Set `url.full` attribute on spans logged for HTTP requests

ahoppen:
We previously only set the request method on these spans, which made it hard to identify which exact HTTP request was causing this span to be emitted.

Also record the URL of the HTTP request to add more information to these spans. While at it, also record the request body size because we already had an attribute key configured for it.

### DIFF
--- a/Sources/AsyncHTTPClient/TracingSupport.swift
+++ b/Sources/AsyncHTTPClient/TracingSupport.swift
@@ -19,10 +19,91 @@ import NIOHTTP1
 import NIOSSL
 import Tracing
 
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import Foundation
+#endif
+
 // MARK: - Centralized span attribute handling
 
 @usableFromInline
 struct TracingSupport {
+    @inlinable
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    static func handleRequestTracingAttributes(
+        _ span: Span,
+        _ request: HTTPClientRequest,
+        configuration: HTTPClient.TracingConfiguration
+    ) {
+        let requestBodySize: Int? =
+            switch request.body?.mode {
+            case .asyncSequence(.known(let length), _), .sequence(.known(let length), _, _):
+                Int(length)
+            case .byteBuffer(let byteBuffer):
+                byteBuffer.readableBytes
+            case .asyncSequence(.unknown, _), .sequence(.unknown, _, _), nil:
+                nil
+            #if UnstableHTTPAPIsSupport
+            case .httpClientRequestBody(.known(let length), startUpload: _):
+                Int(length)
+            case .httpClientRequestBody(.unknown, startUpload: _):
+                nil
+            #endif
+            }
+        let url = URL(string: request.url)
+        let deconstructedURL: DeconstructedURL? =
+            if let url {
+                try? DeconstructedURL(url: url)
+            } else {
+                nil
+            }
+        handleRequestTracingAttributes(
+            span,
+            requestMethod: request.method.rawValue,
+            url: url,
+            deconstructedURL: deconstructedURL,
+            requestBodySize: requestBodySize,
+            configuration: configuration
+        )
+    }
+
+    @inlinable
+    static func handleRequestTracingAttributes(
+        _ span: Span,
+        _ request: HTTPClient.Request,
+        configuration: HTTPClient.TracingConfiguration
+    ) {
+        handleRequestTracingAttributes(
+            span,
+            requestMethod: request.method.rawValue,
+            url: request.url,
+            deconstructedURL: request.deconstructedURL,
+            requestBodySize: request.body?.contentLength.map { Int($0) },
+            configuration: configuration
+        )
+    }
+
+    @usableFromInline
+    static func handleRequestTracingAttributes(
+        _ span: Span,
+        requestMethod: String,
+        url: URL?,
+        deconstructedURL: DeconstructedURL?,
+        requestBodySize: Int?,
+        configuration: HTTPClient.TracingConfiguration
+    ) {
+        let keys = configuration.attributeKeys
+        span.attributes[keys.requestMethod] = requestMethod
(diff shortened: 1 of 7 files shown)


Sources/AsyncHTTPClient/TracingSupport.swift
```diff
@@ -23,6 +24,64 @@ import Tracing
...
+
+    static func handleRequestTracingAttributes(
+        _ span: Span,
+        _ request: HTTPClient.Request,
+        keys: HTTPClient.TracingConfiguration.AttributeKeys
+    ) {
+        handleRequestTracingAttributes(
+            span,
+            requestMethod: request.method.rawValue,
+            url: request.url.description,
```
FranzBusch: I think we need to be careful with that. The full URL can include PII.

ahoppen: Fair point. What do you think of the following: Only log only the host name by default. I wouldn’t consider that as PII and it might already give a hint as to what's being requested. We then add a configuration option to `TracingConfiguration` that allows users to opt into recording the entire URL in traces.

FranzBusch: Yeah I think I would recommend doing this

1. Emit `server.address`, `server.port`, `url.scheme`, `HTTP method`, `status` by default.
2. Emit `url.path` by default but document that callers should configure redaction if they hit user-scoped paths.
3. Make `url.full` and `query` string opt-in.

ahoppen:
Added the more scoped logging and configuration options to opt into full URL logging and to opt out to path logging. 

I chose to mark `DeconstructedURL` as `@usableFromInline` so that I can use it for the structured field tracing. Let me know if there’s an issue with that.

czechboy0:
I think `url.full` should be enabled by default, following OpenTelemetry standard attributes: https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/#url-full

Almost any string field can, in theory, have PII, and I think this is the wrong place to try to make that call. Let application owners filter PII in their telemetry backends instead.

ahoppen:
> I think url.full should be enabled by default, following OpenTelemetry standard attributes

Are you suggesting that we don’t have any configuration options for this in async-http-client? I personally don’t have an opinion either way.

czechboy0:
That's right. That's both how OpenTelemetry and the Swift observability libraries are designed.

ahoppen:
Updated the PR to remove the configuration options.

ktoso:
Just chiming in to say this looks good! Agree on not doing tens of settings in the lib itself, collectors can handle that 👍

ktoso (approved):

ahoppen:
@czechboy0 @FranzBusch could I get another review on this?

Sources/AsyncHTTPClient/TracingSupport.swift
```diff
@@ -19,10 +19,85 @@ import NIOHTTP1
...
+        configuration: HTTPClient.TracingConfiguration
+    ) {
+        let keys = configuration.attributeKeys
+        span.attributes[keys.requestMethod] = requestMethod
+        span.attributes[keys.urlScheme] = deconstructedURL?.scheme.rawValue
+        span.attributes[keys.serverAddress] = deconstructedURL?.connectionTarget.host
+        span.attributes[keys.serverPort] = deconstructedURL?.connectionTarget.port
+        span.attributes[keys.requestBodySize] = requestBodySize
+        span.attributes[keys.urlPath] = url?.path
+        span.attributes[keys.fullUrl] = url?.absoluteString
```
glbrntt: Can you strip the username and password from the URL if they're present (per https://opentelemetry.io/docs/specs/semconv/url/#sensitive-information)?

glbrntt (changes requested):

Sources/AsyncHTTPClient/TracingSupport.swift
ahoppen: I’ll make the changes once we’ve converged on what should and what should not be included in the logs.

Sources/AsyncHTTPClient/TracingSupport.swift
ahoppen: Syned offline and agreed that we should only strip username + password, log everything else as @glbrntt suggested. We don’t need to strip query parameters because https://opentelemetry.io/docs/specs/semconv/url/#sensitive-information says

> Instrumentations *that are aware of specific sensitive query string* […]

Since async-http-client is not aware of sensitive query strings, we don’t have to scrub it.

ahoppen:
Updated to only strip username and password. I would appreciate another review.

ktoso (approved):
Looks good to me, we do our best but it’s collectors and vis config that may need to do more scrubbing anyway.

glbrntt (approved):
LGTM. I think the docs failure will be resolved by https://github.com/swift-server/async-http-client/pull/913

ahoppen:
If you’re happy with this, could you merge the PR? I don’t have merge access to this repo.

glbrntt:
I'm going to wait for the other PR to be merged first as the failure was introduced by your code and we'd rather not merge new CI regressions.

glbrntt:
Agh, sorry @ahoppen -- I was wrong, that change didn't fix it. It's not obvious to me why your change fails to build for docs but does for "regular" CI runs.

```
/__w/async-http-client/async-http-client/Sources/AsyncHTTPClient/TracingSupport.swift:40:13: error: switch must be exhaustive
 38 |     ) {
 39 |         let requestBodySize: Int? =
 40 |             switch request.body?.mode {
    |             |- error: switch must be exhaustive
    |             `- note: add missing case: '.some(.httpClientRequestBody(length: let length, startUpload: let startUpload))'
 41 |             case .asyncSequence(.known(let length), _), .sequence(.known(let length), _, _):
 42 |                 Int(length)
```

ahoppen:
Oh, interesting that this got caught as part of the doc check. I just didn’t check that code builds with `UnstableHTTPAPIsSupport` enabled. Should be fixed now.

PR #907, opened 2026-05-22
Add randomized DNS resolver option for load balancing

pavansai1:
## Summary

  - Adds `HTTPClient.Configuration.DNSResolver` with `.system` (default) and `.randomized`
  cases. When `.randomized` is selected, `NIORandomizedDNSResolver` is installed on the
  underlying `ClientBootstrap`, shuffling addresses returned by `getaddrinfo`.
  - Motivation: enables DNS-based load balancing for services that publish multiple A/AAAA
  records (e.g. Kubernetes headless services). The system resolver returns addresses in a
  deterministic order (typically RFC 6724 destination-address selection), which defeats
  round-robin DNS — clients hammer the first address.


Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -486,6 +486,12 @@ extension HTTPConnectionPool.ConnectionFactory {
                 nioBootstrap
                 .connectTimeout(deadline - NIODeadline.now())
                 .enableMPTCP(clientConfiguration.enableMultipath)
+            switch clientConfiguration.dnsResolver.backing {
+            case .system:
+                break
+            case .randomized:
+                bootstrap = bootstrap.resolver(NIORandomizedDNSResolver(loop: eventLoop))
```
glbrntt, line 493: You'll need to update `Package.swift` and the various `Package@swift-{VERSION}.swift` to require `swift-nio` `from: "2.100.0"` as well.

glbrntt (changes requested):
One small thing to fix but looks good otherwise!

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
pavansai1: Done.

glbrntt (approved):
LGTM, thanks @pavansai1

glbrntt:
Agh CI is failing because SwiftNIO doesn't support Swift 6.0. AHC also shouldn't support it. Dropping support in #908 which should unblock the CI.

glbrntt:
Sorry @pavansai1, you'll need to resolve the conflict as `Package@swift-6.0.swift` has been deleted.

Package@swift-6.0.swift
```diff
@@ -0,0 +1,124 @@
+// swift-tools-version:6.0
```
glbrntt: We dropped Swift 6 support so you need to remove this file

glbrntt (changes requested):

Package@swift-6.0.swift
pavansai1: Done. removed the file. Thank you.

glbrntt (approved):
Thank you!

PR #910, opened 2026-06-11
Add SPKI certificate pinning support

o-nnerb:
This PR introduces SPKI-based certificate pinning to `AsyncHTTPClient` to mitigate MITM attacks and compromised CAs. 

**Highlights:**
- Implements `.strict` and `.audit` modes with multi-algorithm hashing (SHA-256/384/512) and constant-time comparison.
- Aligned with OWASP/NIST security standards, featuring rotation safety warnings and explicit error handling.
- **Requirement:** Requires OpenSSL/BoringSSL backend (Network.framework connections ignore pinning configuration).


o-nnerb:
@czechboy0 @FranzBusch can you review this?

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,359 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```
glbrntt: It's 2026

```suggestion
// Copyright (c) 2026 Apple Inc. and the AsyncHTTPClient project authors
```

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -53,6 +53,9 @@ public struct HTTPClientRequest: Sendable {
     /// Request-specific TLS configuration, defaults to no request-specific TLS configuration.
     public var tlsConfiguration: TLSConfiguration?
 
+    /// Optional SPKI pinning configuration for TLS certificate validation.
+    public var tlsPinning: SPKIPinningConfiguration?
```
glbrntt, line 57: I noted in the description that the Network.framework backend ignores this. I think it'd be better to validate the configuration and throw an error if pinning is enabled and Network.framework is being used, otherwise a user may not realize that they aren't actually using pinning. This behavior should also be documented. Same goes for using NIOSSL with unsupported platforms (i.e. macOS < 10.15, however unlikely that is)

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,359 @@
...
+/// 1. Iterating a fixed number of times (determined by hash algorithm)
+/// 2. Processing all candidates even after a match is found
+/// 3. Performing all secret-dependent operations before any conditional branches
+///
+/// SECURITY INVARIANT: All candidates share identical length (enforced by
+/// SPKIPinningConfiguration grouping and SPKIHash validation). Length mismatches
+/// are rejected early using public knowledge (algorithm-determined digest size),
+/// which cannot leak secret information.
+internal func constantTimeAnyMatch(_ target: Data, _ candidates: [SPKIHash]) -> Bool {
+    guard !candidates.isEmpty else { return false }
```
glbrntt: nit: `if X { ... }` is typically more readable than `guard !X else { ... }`

```suggestion
    if candidates.isEmpty { return false }
```

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,359 @@
...
+        self.rawValue = rawValue
+    }
+
+    public static func == (lhs: Self, rhs: Self) -> Bool {
+        lhs.rawValue == rhs.rawValue
+    }
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(rawValue)
+    }
```
glbrntt: Are these necessary? The compiler should be able to synthesize them.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,359 @@
...
+    ///
+    /// Use exclusively for debugging, testing, or migration scenarios. Never use in production.
+    public static let audit = SPKIPinningPolicy(rawValue: .audit)
+
+    /// Reject connections with untrusted certificates.
+    ///
+    /// Use in production environments where security is paramount.
+    public static let strict = SPKIPinningPolicy(rawValue: .strict)
+
+    public var description: String {
```
glbrntt, line 223: Did you want this type to conform to `CustomStringConvertible`?

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,359 @@
...
+    private let logger: Logger
+
+    init(
+        tlsPinning: SPKIPinningConfiguration,
+        logger: Logger
+    ) {
+        self.tlsPinning = tlsPinning
+        self.logger = logger
+
+        if tlsPinning.pins.count < 2 && tlsPinning.policy == .strict {
```
glbrntt: Is there anywhere higher in the stack we can verify this? Can (and should) we disallow this via construction?

Sources/AsyncHTTPClient/HTTPClient.swift
```diff
@@ -837,6 +837,9 @@ public final class HTTPClient: Sendable {
         /// TLS configuration, defaults to `TLSConfiguration.makeClientConfiguration()`.
         public var tlsConfiguration: Optional<TLSConfiguration>
 
+        /// Optional SPKI pinning configuration for TLS certificate validation.
```
glbrntt, line 840: Please document the supported platforms here and what happens when a platform isn't supported by the config is set.

Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -1020,6 +1021,197 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
         }
     }
 
+    func testSPKIPinning_ValidPin_AllowsConnection() {
+        XCTAsyncTest {
```
glbrntt: I don't think `XCTAsyncTest` is required anymore, XCTest has had solid `async` support for long enough now.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -0,0 +1,359 @@
...
+    ///
+    /// - Parameters:
+    ///   - algorithm: Hash algorithm used to generate the digest.
+    ///   - base64: Base64-encoded hash digest.
+    ///
+    /// - Throws: `HTTPClientError.invalidDigestLength` if length doesn't match algorithm.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    public init<Algorithm: HashFunction>(algorithm: Algorithm.Type, base64: String) throws {
+        guard let data = Data(base64Encoded: base64) else {
+            throw HTTPClientError.invalidDigestLength
```
glbrntt: Should this be a separate error case?

Tests/AsyncHTTPClientTests/SPKIHashTests.swift
```diff
@@ -0,0 +1,133 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```
glbrntt: 2026

Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
```diff
@@ -1020,6 +1021,197 @@ final class AsyncAwaitEndToEndTests: XCTestCase {
         }
     }
 
+    func testSPKIPinning_ValidPin_AllowsConnection() {
```
glbrntt: Can you add tests for http/1.1 as well?

glbrntt: There's also a lot of duplicated code across these four tests, can you refactor to have a test function which varies the pin kind (valid/invalid) and policy?

Tests/AsyncHTTPClientTests/SPKIPinningTests.swift
```diff
@@ -0,0 +1,278 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the AsyncHTTPClient open source project
+//
+// Copyright (c) 2021 Apple Inc. and the AsyncHTTPClient project authors
```
glbrntt: 2026

glbrntt (changes requested):
Thanks! This mostly looks good but I left some feedback inline which needs addressing.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: Hmm, what are your suggestions for this? I couldn’t think of any feasible solution with the current implementation. Using `HTTPClientError.invalidCertificatePinning(_:)` or any of the cases of `SPKIPinningHandlerError` doesn’t seem logical to me.

One possibility is to change the initializer to `init?`. However, I’m also unsure if this would be a good idea. The developer might get confused about the cause of the error if they encounter it.

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
o-nnerb: Thanks for the feedback! I updated the documentation directly on `SPKIPinningConfiguration` to centralize the warnings about Network.framework and the platform requirements.

Regarding the validation: good catch. I'll add the runtime check to throw an error instead of silently ignoring/failing it.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
o-nnerb: I’ve moved this validation to `HTTPConnectionPool.ConnectionFactory`, as shown below:

```swift
private extension HTTPConnectionPool.ConnectionFactory {

   func setupSPKIPinningHandlerIfNeeded(_ sync: ChannelPipeline.SynchronousOperations, logger: Logger) throws {
        guard let tlsPinning else {
            return
        }

        guard #available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) else {
            throw SPKIPinningHandlerError.platformNotSupported
        }

        if tlsPinning.pins.count < 2 && tlsPinning.policy == .strict {
            logger.warning(
                "SPKIPinningHandler deployed with < 2 pins in strict mode — catastrophic lockout risk on certificate rotation!",
                metadata: [
                    "current_pin_count": .stringConvertible(tlsPinning.pins.count),
                    "recommendation": .string("Deploy multiple pins to enable safe certificate rotation"),
                ]
            )
        }

        let pinningHandler = SPKIPinningHandler(
            tlsPinning: tlsPinning,
            logger: logger
        )

        try sync.addHandler(pinningHandler)
    }
}
```

Tests/AsyncHTTPClientTests/AsyncAwaitEndToEndTests.swift
o-nnerb: Done! I refactored the tests to use a shared helper function `runSPKIPinningTest` that takes parameters for pin validity, policy, and HTTP mode. 

Added:
- HTTP/1.1 tests (4 tests: valid/invalid × strict/audit)
- HTTP/2 tests (4 tests: valid/invalid × strict/audit)
- Network.framework backend test that validates the unsupported error is thrown

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
glbrntt: You can add a new error case to `HTTPClientError` -- maybe `invalidBase64Digest`, `invalidDigest` or something along those lines. I find the length part misleading for the existing error.

Sources/AsyncHTTPClient/ConnectionPool/ChannelHandler/SPKIPinningHandler.swift
```diff
@@ -126,8 +126,19 @@ internal func constantTimeAnyMatch(_ target: Data, _ candidates: [SPKIHash]) ->
...
 /// - Warning: Always deploy multiple pins to enable safe certificate rotation.
 ///   Single-pin configurations in `.strict` mode risk catastrophic lockout.
+///
+/// - Note: SPKI pinning is only supported when the client uses the NIOSSL
+///   backend. When `Network.framework` is configured as the TLS backend,
+///   providing a non-nil configuration will cause the request to fail at
+///   runtime.
+///
+/// - Note: SPKI pinning requires Apple platforms with a minimum deployment
+///   target of macOS 10.15, iOS 13, tvOS 13, or watchOS 6.
```
glbrntt: Can you document the behavior if the Apple platform being used doesn't meet these requirements?

glbrntt (commented):
This is looking good, just a couple of small things left. The e2e tests are much better now as well 👍

glbrntt (approved):
Thanks, this looks great! I would like @Lukasa to have a look before this is merged though.

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
```diff
@@ -465,6 +498,9 @@ extension HTTPConnectionPool.ConnectionFactory {
                         try channel.pipeline.syncOperations.addHandler(
                             NWWaitingHandler(requester: requester, connectionID: connectionID)
                         )
+                        if tlsPinning != nil {
+                            throw SPKIPinningHandlerError.networkFrameworkNotSupported
+                        }
```
o-nnerb: @glbrntt, while reviewing this code with Claude, it identified two errors:

The if statement here incorrectly rejects:
 
- Pure HTTP requests (without TLS) when tlsPinning is configured globally.
- HTTPS requests made via proxy (HTTP/SOCKS) on Apple platforms, even though the actual TLS is performed using NIOSSL (which supports pinning correctly).

What do you say?

Tests/AsyncHTTPClientTests/SPKIPinningTests.swift
```diff
@@ -0,0 +1,479 @@
...
+            privateKey: .privateKey(TestTLS.privateKey)
+        )
+
+        let bin = HTTPBin(.http2(tlsConfiguration: tlsConfig))
+        defer { XCTAssertNoThrow(try bin.shutdown()) }
+
+        var config = HTTPClient.Configuration().enableFastFailureModeForTesting()
+        config.tlsConfiguration = TLSConfiguration.makeClientConfiguration()
+        config.tlsConfiguration?.trustRoots = .certificates([certificate])
+        // Network.framework não suporta .noHostnameVerification, então removemos essa linha
```
o-nnerb: @glbrntt, there’s also this Portuguese line!  🤣

Tests/AsyncHTTPClientTests/SPKIPinningTests.swift
```diff
@@ -0,0 +1,479 @@
...
+import XCTest
+
+@testable import AsyncHTTPClient
+
+#if canImport(Network)
+import Network
+import NIOTransportServices
+#endif
+
+class SPKIPinningTests: XCTestCase {
```
o-nnerb, line 33: @glbrntt it's missing tests with proxy configuration

Tests/AsyncHTTPClientTests/SPKIPinningTests.swift
glbrntt: I saw that in the last review and thought it'd be a fun bit of history 😅

Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
glbrntt: Writing some tests to verify those claims and any fixes would be good.

o-nnerb:
I've marked this PR as a draft because I've identified some deeper architectural concerns with the original implementation. Specifically, it feels conceptually off to bundle the entire SSL Pinning functionality within the scope of `AsyncHTTPClient`. 
 
If we look at `URLSession`, this kind of built-in functionality doesn't exist. Furthermore, the current implementation hits technical limitations due to **Transport Services on NIO** and its reliance on the underlying **Network framework**. 
 
Because of these details, I'm stepping back to re-evaluate the approach. I'm exploring a deeper, more decoupled solution, similar to how `URLSession` handles it. I'll share more details and update the PR status once I've finished the new implementation and testing.
 
Thank you all for the reviews and the great discussions. I apologize for the extra effort and rework this pivot has caused on your end.

PR #912, opened 2026-07-23
add proxy-header support

benrobby:
Motivation:

when connecting via a HTTP proxy, this now allows users to set HTTP headers on the CONNECT request to the proxy. This is useful for caller attribution purposes.

Modifications:
- adds a `connectHeaders` parameter to HTTPClient.Configuration
- parses `connectHeaders` swift-configuration property as list of colon-separated `key: value` pairs
- sets proxy headers on CONNECT request

Result:
implements proxy-header support

### DIFF
--- a/Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
@@ -63,7 +63,7 @@ extension HTTPClient.Configuration {
         /// These headers are not sent to the
         /// destination server, and are ignored for SOCKS proxies.
         /// The `host` and `proxy-authorization` headers cannot be overridden through this property
-        /// Note: Excluded from equality / hash, because HTTPHeaders are not hashable.
+        /// Note: Excluded from hash, because HTTPHeaders are not hashable.
         public var connectHeaders: HTTPHeaders = [:]
 
         /// Create an HTTP proxy configuration.
--- a/Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
+++ b/Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
@@ -66,7 +66,7 @@ extension HTTPClient.Configuration {
         /// Note: Excluded from equality / hash, because HTTPHeaders are not hashable.
         public var connectHeaders: HTTPHeaders = [:]
 
-        /// Create a HTTP proxy configuration.
+        /// Create an HTTP proxy configuration.
         ///
         /// - parameters:
         ///     - host: proxy server host.
@@ -75,7 +75,7 @@ extension HTTPClient.Configuration {
             .init(host: host, port: port, type: .http(nil))
         }
 
-        /// Create a HTTP proxy configuration.
+        /// Create an HTTP proxy configuration.
         ///
         /// - parameters:
         ///     - host: proxy server host.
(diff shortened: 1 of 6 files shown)


Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
```diff
@@ -76,12 +84,37 @@ extension HTTPClient.Configuration {
             .init(host: host, port: port, type: .http(authorization))
         }
 
+        /// Create a HTTP proxy.
```
glbrntt: I appreciate the same mistake exists in the other docs in this file but this function definitely doesn't create an HTTP proxy!

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
```diff
@@ -76,12 +84,37 @@ extension HTTPClient.Configuration {
...
+
         /// Create a SOCKSv5 proxy.
         /// - parameter host: The SOCKSv5 proxy address.
         /// - parameter port: The SOCKSv5 proxy port, defaults to 1080.
         /// - returns: A new instance of `Proxy` configured to connect to a `SOCKSv5` server.
         public static func socksServer(host: String, port: Int = 1080) -> Proxy {
             .init(host: host, port: port, type: .socks)
         }
+
+        // `connectHeaders` is omitted (HTTPHeaders is not hashable)
```
glbrntt, line 114: That's unfortunate... Could you document the `connectHeaders` property to say they're excluded too?

glbrntt (commented):
LGTM module a couple of nits

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
benrobby: changed this one and the others to `Create an HTTP proxy configuration`.

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
benrobby: documented `/// Note: Excluded from equality / hash, because HTTPHeaders are not hashable.`

Sources/AsyncHTTPClient/HTTPClient+Proxy.swift
```diff
@@ -57,7 +58,15 @@ extension HTTPClient.Configuration {
 
         var type: ProxyType
 
-        /// Create a HTTP proxy.
+        /// Extra headers sent only on the HTTP `CONNECT` request to the proxy.
+        ///
+        /// These headers are not sent to the
+        /// destination server, and are ignored for SOCKS proxies.
+        /// The `host` and `proxy-authorization` headers cannot be overridden through this property
+        /// Note: Excluded from equality / hash, because HTTPHeaders are not hashable.
```
glbrntt: They aren't excluding from equality checks: `HTTPHeaders` are `Equatable`

benrobby: right, adjusted the comment. I debated whether it makes sense to also exclude them from equality, but i think it's fine as is now (it only requires a == b -> hash(a) == hash(b), but not the other way around)

glbrntt (approved):
Nice patch, thank you!

PR #914, opened 2026-08-03
Add local port binding to HTTPClient.Configuration

VictorDebray:
Adds a `localPort: Int` knob on `HTTPClient.Configuration` that pins the TCP source port the connection binds to before connecting, plus a per-request `HTTPClientRequest.localPort` override. `0` (the default) keeps the previous behaviour of letting the OS assign an ephemeral port.

`localPort` is threaded through `RequestOptions` and `ConnectionPool.Key` alongside the existing `localAddress`, so connections with different `(localAddress, localPort)` pairs are pooled separately. Both the NIOTS (`requiredLocalEndpoint`) and NIOPosix (`bind(to:)`) bootstraps in the plain and TLS paths honour it.

### DIFF
--- a/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
+++ b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
@@ -60,6 +60,14 @@ public struct HTTPClientRequest: Sendable {
     /// Defaults to `nil` (use client configuration default).
     public var localAddress: String?
 
+    /// The local TCP source port to bind this request's connection to.
+    ///
+    /// When set, overrides ``HTTPClient/Configuration/localPort`` for this
+    /// request. Only consulted when a local address (request-level or
+    /// configuration-level) is also set. Defaults to `nil` (use client
+    /// configuration default).
+    public var localPort: Int?
+
     public init(url: String) {
         self.url = url
         self.method = .GET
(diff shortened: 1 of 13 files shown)


fabianfett (changes requested):
This looks like a solid PR. Thank you. However before merging, we will need tests, that validate that the outgoing request has actually been made from the specified port. We will need this for the old (EventLoopFuture) API as well as the new one. Overwriting the settings also needs to be tested.

VictorDebray:
I have added the needed tests. Thanks for the review!

Sources/AsyncHTTPClient/AsyncAwait/HTTPClientRequest.swift
```diff
@@ -60,13 +60,23 @@ public struct HTTPClientRequest: Sendable {
     /// Defaults to `nil` (use client configuration default).
     public var localAddress: String?
 
+    /// The local TCP source port to bind this request's connection to.
+    ///
+    /// When set, overrides ``HTTPClient/Configuration/localPort`` for this
+    /// request. Only consulted when a local address (request-level or
+    /// configuration-level) is also set. Values outside of `0...65535` fail the
```
fabianfett, line 67: > Only consulted when a local address (request-level or configuration-level) is also set.

What does this mean? This is only consulted if a client-level setting is set? Is that true?

PR #915, opened 2026-08-15
Add network path/interface restriction fields to HTTPClient.Configuration

o-nnerb:
## Motivation

`URLSessionConfiguration` exposes `allowsCellularAccess`, `allowsExpensiveNetworkAccess`, and `allowsConstrainedNetworkAccess`. Clients migrating from `URLSession`, or apps that need to respect a user's Low Data Mode / cellular preferences, expect parity. `HTTPClient.Configuration` has no equivalent today — and critically, no library built on top of `AsyncHTTPClient` can add this either, since the only place these `NWParameters` constraints can be applied is at connection-establishment time, inside `HTTPConnectionPool+Factory.swift`, which is internal to this package. There is no generic escape hatch for `NWParameters` customization from the outside.

`networkFrameworkWaitForConnectivity` is the existing precedent for exactly this kind of field: declared unconditionally (no `#if canImport(Network)` on the declaration, it's simply inert elsewhere), and consumed only inside the two `#if canImport(Network)`-guarded bootstrap factories.

## Modifications

- Adds `HTTPClient.Configuration.NetworkInterfaceType`, a public struct mirroring `NWInterface.InterfaceType`'s cases (`.other`, `.wifi`, `.cellular`, `.wiredEthernet`, `.loopback`) without requiring `Network` to be imported at the declaration site — following the same `Backing`-enum pattern `DNSResolver`/`HTTPVersion` already use in this file, so the public type stays frozen/ABI-stable while the backing can evolve.
- Adds five new `Configuration` fields:
  - `prohibitedInterfaceTypes: Set<NetworkInterfaceType>` — mirrors `NWParameters.prohibitedInterfaceTypes`.
  - `requiredInterfaceType: NetworkInterfaceType?` — mirrors `NWParameters.requiredInterfaceType`, exposed as `Optional` instead of that property's non-optional `.other`-as-"unrestricted"-sentinel default, since `.other` is otherwise both a real interface category and the "don't care" default, which is easy to misread.
  - `allowsExpensiveNetworkAccess` / `allowsConstrainedNetworkAccess: Bool` — inverted mirrors of `NWParameters.prohibitExpensivePaths` / `.prohibitConstrainedPaths`.
  - `allowsUltraConstrainedPaths: Bool` — direct mirror of `NWParameters.allowUltraConstrainedPaths` (available macOS/iOS/watchOS/tvOS/visionOS 26.0+; a no-op on older OS versions, not just non-Darwin platforms).
  
  All defaults match `NWParameters`' own defaults, confirmed both against the `Network.swiftinterface` shipped in Xcode 26.5 and empirically against a live `NWParameters` instance.
- Wires all five into both `makePlainBootstrap` and `makeTLSBootstrap` via one new `configureNWParameters(_:localAddress:)` helper. `NIOTSConnectionBootstrap.configureNWParameters(_:)` only retains the most recently passed closure rather than composing across calls, so the existing `requiredLocalEndpoint` local-address-binding closure and the new field wiring are folded into a single closure — two separate `configureNWParameters { }` calls would have silently dropped the local-address binding.
- Adds two tests to `HTTPClientNIOTSTests.swift`: one exercising all five fields together with `localAddress` set (regression coverage for the closure-composition issue above), one confirming `.other` as an explicit `requiredInterfaceType` behaves the same as the `nil`/unrestricted default.

## Result

`HTTPClient.Configuration` can restrict connections by interface category and by expensive/constrained/ultra-constrained network path, matching what `URLSessionConfiguration` already offers, on Darwin platforms using Network.framework as the transport. No-op everywhere else (non-Darwin platforms, and for `allowsUltraConstrainedPaths`, older OS versions), same as the existing `networkFrameworkWaitForConnectivity` field.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
@@ -434,6 +434,53 @@ extension HTTPConnectionPool.ConnectionFactory {
         }
     }
 
+    #if canImport(Network)
+    /// Applies every `HTTPClient.Configuration` field that maps onto `NWParameters` in one place.
+    /// `NIOTSConnectionBootstrap.configureNWParameters(_:)` stores only the most recently passed
+    /// closure rather than composing across calls, so every concern that needs `NWParameters` must
+    /// be folded into a single call to this method instead of separate `configureNWParameters { }`
+    /// invocations. Callers already run under `#available(OSX 10.14, iOS 12.0, tvOS 12.0,
+    /// watchOS 6.0, *)`, the minimum for `NWParameters` itself; fields with a higher minimum
+    /// (`prohibitConstrainedPaths`, `allowUltraConstrainedPaths`) are individually re-checked below
+    /// instead of raising that requirement for the whole method.
+    @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
+    private func configureNWParameters(_ params: NWParameters, localAddress: String?) {
+        if let localAddress {
+            params.requiredLocalEndpoint = NWEndpoint.hostPort(
+                host: NWEndpoint.Host(localAddress),
+                port: .any
+            )
+        }
+        if !self.clientConfiguration.prohibitedInterfaceTypes.isEmpty {
+            params.prohibitedInterfaceTypes = self.clientConfiguration.prohibitedInterfaceTypes.map {
+                switch $0.backing {
+                case .other: return .other
+                case .wifi: return .wifi
+                case .cellular: return .cellular
+                case .wiredEthernet: return .wiredEthernet
+                case .loopback: return .loopback
+                }
+            }
+        }
+        if let required = self.clientConfiguration.requiredInterfaceType {
+            switch required.backing {
+            case .other: params.requiredInterfaceType = .other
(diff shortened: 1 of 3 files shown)


fabianfett:
We highly recommend using `URLSession` or `Network.framework` on Apple platforms if those features are required. 

> Clients migrating from `URLSession`, or apps that need to respect a user's Low Data Mode / cellular preferences, expect parity.

For libraries that want to abstract over the underlying HTTP transport, we recommend looking at the new abstract HTTPClient protocol:

https://forums.swift.org/t/designing-an-http-client-api-for-swift/85254
https://github.com/apple/swift-http-api-proposal/

If you have a specific use-case in mind, please open up this PR and state your use-case.

o-nnerb:
Thanks for the links — I wasn't aware of those discussions. Based on them, I think there's still room for a focused contribution that gives real parity with `URLSession`'s behavior. I went back and reviewed the original PR, and you're right that it overreached: only `allowsExpensiveNetworkAccess` and `allowsConstrainedNetworkAccess` actually mirror `URLSessionConfiguration` 1:1 — `prohibitedInterfaceTypes`, `requiredInterfaceType`, and `allowsUltraConstrainedPaths` go beyond what `URLSession` itself exposes. I've trimmed the PR down to just those two fields. Would appreciate a second look when you have time.

PR #918, opened 2026-08-25
Add allowsExpensiveNetworkAccess/allowsConstrainedNetworkAccess to HTTPClient.Configuration

o-nnerb:
## Context

Follow-up to #915, which was closed with the feedback that it overreached `URLSessionConfiguration` parity and that Apple-platform-specific transport tuning is better left to `URLSession`/`Network.framework` directly, or to the emerging cross-platform HTTP client abstraction ([forum thread](https://forums.swift.org/t/designing-an-http-client-api-for-swift/85254), [apple/swift-http-api-proposal](https://github.com/apple/swift-http-api-proposal/)).

Going back through `URLSessionConfiguration`'s actual surface, only two of the original five fields have a real 1:1 equivalent there:

- `allowsExpensiveNetworkAccess` ↔ `URLSessionConfiguration.allowsExpensiveNetworkAccess`
- `allowsConstrainedNetworkAccess` ↔ `URLSessionConfiguration.allowsConstrainedNetworkAccess`

The other three (`prohibitedInterfaceTypes`, `requiredInterfaceType`, `allowsUltraConstrainedPaths`) go beyond what `URLSessionConfiguration` itself exposes — `URLSession` only has the coarser `allowsCellularAccess` for interface restriction, and nothing for ultra-constrained paths. Adding those wasn't really "parity," so this PR drops them and keeps only the two fields that are genuinely mirroring `URLSession`.

## Modifications

- Add `HTTPClient.Configuration.allowsExpensiveNetworkAccess` / `allowsConstrainedNetworkAccess: Bool`, inverted mirrors of `NWParameters.prohibitExpensivePaths` / `.prohibitConstrainedPaths`, defaulting to `NWParameters`' own defaults (`true`/`true`).
- Wire both into `makePlainBootstrap` and `makeTLSBootstrap` in `HTTPConnectionPool+Factory.swift` via a `configureNWParameters(_:localAddress:)` helper, since `NIOTSConnectionBootstrap.configureNWParameters(_:)` only keeps the most recently passed closure — the existing `requiredLocalEndpoint`-binding closure and the new field wiring had to be folded into one closure to avoid silently dropping local-address binding.
- Add a regression test in `HTTPClientNIOTSTests.swift` exercising both fields together with `localAddress` set, covering the closure-composition issue above.

## Result

`HTTPClient.Configuration` can restrict connections to non-expensive / non-constrained network paths, matching the two fields `URLSessionConfiguration` actually exposes for this, on Darwin platforms using Network.framework as the transport. No-op everywhere else, same as the existing `networkFrameworkWaitForConnectivity` field.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTPConnectionPool+Factory.swift
@@ -440,9 +440,9 @@ extension HTTPConnectionPool.ConnectionFactory {
     /// closure rather than composing across calls, so every concern that needs `NWParameters` must
     /// be folded into a single call to this method instead of separate `configureNWParameters { }`
     /// invocations. Callers already run under `#available(OSX 10.14, iOS 12.0, tvOS 12.0,
-    /// watchOS 6.0, *)`, the minimum for `NWParameters` itself; fields with a higher minimum
-    /// (`prohibitConstrainedPaths`, `allowUltraConstrainedPaths`) are individually re-checked below
-    /// instead of raising that requirement for the whole method.
+    /// watchOS 6.0, *)`, the minimum for `NWParameters` itself; `prohibitConstrainedPaths`, which
+    /// has a higher minimum, is individually re-checked below instead of raising that requirement
+    /// for the whole method.
     @available(OSX 10.14, iOS 12.0, tvOS 12.0, watchOS 6.0, *)
     private func configureNWParameters(_ params: NWParameters, localAddress: String?) {
         if let localAddress {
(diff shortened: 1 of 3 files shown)


fabianfett:
This PR is just scoped down #915. The original statement in #915 is still true.

We highly recommend using `URLSession` or `Network.framework` on Apple platforms if those features are required.

> Clients migrating from `URLSession`, or apps that need to respect a user's Low Data Mode / cellular preferences, expect parity.

For libraries that want to abstract over the underlying HTTP transport, we recommend looking at the new abstract HTTPClient protocol:

https://forums.swift.org/t/designing-an-http-client-api-for-swift/85254
https://github.com/apple/swift-http-api-proposal/

@o-nnerb Please refrain from posting PRs that are end to end implemented and proposed by an AI without human supervision, as it generates an additional burden on us maintainers.

o-nnerb:
@fabianfett I apologize for the inconvenience and thank you for the heads-up. I will make sure to thoroughly review and supervise any AI-assisted contributions before submitting them in the future.

o-nnerb:
Just for context, this PR was based on an old proposal I've been planning to implement since 2023 (https://github.com/orgs/request-dl/discussions/191). However, I understand now that this falls outside the current scope of this repository. Thank you for the clarification.

PR #925, opened 2026-09-10
Fix HTTP/2 error after response completion

### LINKED ISSUE #922: Crash in `HTTP2ClientRequestHandler`
Despite what the comment says, this can crash on the client if the server sends a GOAWAY or RST_STREAM at the right time:

https://github.com/swift-server/async-http-client/blob/f95c908967e98c68c5ce3fd61a7974e7e869e303/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift#L247-L249

AI Disclosure: yes I used Gemini.

I asked Gemini to write the repro and the following comments, but I reviewed them myself. I found the initial problem while working on the SDK for Google Cloud Storage. That is artisanal, lovingly hand-crafted code, not AI slop.  Reducing the problem to a smaller repro did involve LLMs.

[HTTP2ClientCrashRepro.tar.gz](https://github.com/user-attachments/files/31807696/HTTP2ClientCrashRepro.tar.gz)

----

A reproducer for a crash in [`swift-server/async-http-client`](https://github.com/swift-server/async-http-client) (`HTTP2ClientRequestHandler.swift:249: Fatal error: Unexpectedly found nil while unwrapping an Optional value`).

This package runs an in-process mock HTTP/2 server on `127.0.0.1` and reproduces the crash without external network requests.

---

## How to Run

```bash
cd HTTP2ClientCrashRepro
swift run
```

Or run with an explicit trigger action:

```bash
# Test stream reset (RST_STREAM) after early response headers (default)
swift run HTTP2ClientCrashRepro rst

# Test connection termination (GOAWAY) after early response headers
swift run HTTP2ClientCrashRepro goaway
```

---

## Crash Trace

```text
=== Starting HTTP2ClientCrashRepro ===
Testing isolated HTTP/2 repro without external network calls...
Configured server action: respondThenRstStream(errorCode: HTTP2ErrorCode<0x8 Cancel>)
Mock HTTP/2 Server listening at https://127.0.0.1:39483
Sending streaming request to mock server...
AsyncHTTPClient/HTTP2ClientRequestHandler.swift:249: Fatal error: Unexpectedly found nil while unwrapping an Optional value
Response received: 200 OK

💣 Program crashed: Signal 4: Backtracing from 0x7f...
```

---

## Root Cause Analysis

The crash occurs when the server returns a complete response while the client's request body is still streaming, followed by a stream reset (`RST_STREAM`) or connection closure (`GOAWAY`):

1. **Streaming Request Initiated**:
   The client begins an HTTP/2 request with a streaming body (`HTTPClientRequest.Body.stream`).
2. **Server Responds Early**:
   The server sends response headers (`200 OK`, `412 Precondition Failed`, etc.) with `endStream: true`.
3. **Response Delivered & Request Cleared**:
   `HTTP2ClientRequestHandler.run(.forwardResponseEnd)` delivers the response to the caller and clears its stored request reference:
   ```swift
   // Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift:234
   self.request = nil
   ```
4. **Server Resets Stream**:
   While the client's request body is still streaming, the server issues a `RST_STREAM` frame.
5. **Stream Error Forwarded**:
   SwiftNIO delivers the stream reset to `errorCaught`.
6. **State Machine Returns `.failRequest`**:
   Because the request state machine was in `.running(.streaming(...), .endReceived)`, it transitions to `.failed` and returns the action `.failRequest(error, ...)`.
7. **Force-Unwrap Crash**:
   `HTTP2ClientRequestHandler.run(.failRequest)` force-unwraps `self.request!`:
   ```swift
   // Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift:249
   case .failRequest(let error, let finalAction):
       self.request!.fail(error) // CRASH: self.request was set to nil in step 3
   ```

---

## Why TLS / HTTPS Is Required

`AsyncHTTPClient` does not implement cleartext HTTP/2 (`h2c` / prior knowledge). In `HTTPConnectionPool+Factory.swift` (lines 256 and 386), all plain `http://` schemes use HTTP/1.1:

```swift
private func makePlainChannel(...) {
    ...
    bootstrap.connect(target: self.key.connectionTarget).map {
        .http1_1($0)
    }.cascade(to: promise)
}
```

When connecting to `http://`, `AsyncHTTPClient` sends HTTP/1.1 text (`POST /test HTTP/1.1\r\n...`). An HTTP/2 server expecting the HTTP/2 client connection preface (`PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n`) rejects this with `invalid constant string`.

In `AsyncHTTPClient`, `HTTP2ClientRequestHandler` is only added when ALPN negotiates `"h2"` during TLS (`HTTPConnectionPool+Factory.swift:428`). Therefore, an in-process TLS server with ALPN is required to exercise HTTP/2 and reproduce this bug.

---

## Test Certificates

The certificate and private key in `Sources/main.swift` are test credentials:
- They are generated for loopback testing on `127.0.0.1` and not used in production.
- The client sets `config.tlsConfiguration?.certificateVerification = .none` to accept the local test certificate.
- They are safe to check into source control.

---

## Why This Matters for Google Cloud & HTTP/2 SDKs

In Google Cloud SDKs (such as Cloud Storage):
- **Streaming Uploads**: Resumable uploads and multipart object uploads stream request bodies via `HTTPClientRequest.Body.stream`.
- **Early Server Responses**: If an upload fails early (e.g. `412 Precondition Failed` on `ifGenerationMatch`, quota exhaustion, or authorization failure), Cloud Storage closes or resets the stream before the client completes writing the request body.
- **Process Termination**: Rather than delivering an error to the caller (so that retry or recovery logic can execute), `AsyncHTTPClient` triggers a fatal error (`SIGILL`), crashing the process.

---

## System & Environment Information

The crash was captured and reproduced in the following environment:

| Property | Value |
|---|---|
| **Swift Version** | `Swift version 6.3.3 (swift-6.3.3-RELEASE)` / Target: `x86_64-unknown-linux-gnu` |
| **Operating System** | `Debian GNU/Linux rodete` (`VERSION_CODENAME=rodete`) |
| **Linux Kernel** | `6.18.14-1rodete4-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.18.14-1rodete4 (2026-06-15) x86_64` |
| **C Library (glibc)** | `ldd (Debian GLIBC 2.43-4+gl0) 2.43` |
| **Architecture** | `x86_64` |
| **Available CPUs (`nproc`)** | `64` (`AMD EPYC 7B13`, 1 socket, 32 cores, 64 threads) |
| **System Memory** | `117 GiB RAM` (Total) / `~102 GiB` (Available) |

### Tested Dependency Versions

| Package | Pinned Version | Revision |
|---|---|---|
| `swift-server/async-http-client` | `1.36.0` | `9544287b9416c0bc71e58b9f3aead8dd14b16103` |
| `apple/swift-nio` | `2.101.3` | `0b18836bd8b0162e7e17a995a3fbee20ed8f3b2b` |
| `apple/swift-nio-http2` | `1.45.0` | `45bdf670248be5f16ec0340e125dca285536f0fb` |
| `apple/swift-nio-ssl` | `2.37.2` | `d930168b86f46ca51a4bc09c5ca45c1833db8067` |

Alex-Wengg:
Motivation:

An HTTP/2 stream error can arrive after a response end has completed the request while its request body is still streaming. `HTTP2ClientRequestHandler` clears the request on response completion and then force-unwraps it while processing the later error, causing the crash reported in #922.

Modifications:

- Only notify the request of failure when it is still present.
- Preserve timeout cancellation and HTTP/2 stream cleanup for the late error.
- Add a regression test for an error after response completion during a streaming upload.

Result:

Late HTTP/2 stream errors no longer crash the handler or fail an already-completed request.

Validation:

- `swift build`
- `swift format lint --strict --configuration .swift-format` on both changed files
- The exact reproducer attached to #922, pointed at this checkout: receives `200 OK`, completes normally, and exits 0 (unmodified 1.36.1 exits 133 at `HTTP2ClientRequestHandler.swift:249`).
- The XCTest suite could not run locally because this Command Line Tools installation does not include the XCTest Swift module; CI will provide that coverage.

Fixes #922

### DIFF
--- a/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
+++ b/Sources/AsyncHTTPClient/ConnectionPool/HTTP2/HTTP2ClientRequestHandler.swift
@@ -244,9 +244,9 @@ final class HTTP2ClientRequestHandler: ChannelDuplexHandler {
             self.request!.receiveResponseBodyParts(parts)
 
         case .failRequest(let error, let finalAction):
-            // We can force unwrap the request here, as we have just validated in the state machine,
-            // that the request object is still present.
-            self.request!.fail(error)
+            // The request may already have completed after receiving a response end while its body
+            // was still streaming. A subsequent stream error must not fail that request twice.
+            self.request?.fail(error)
             self.request = nil
             self.runTimeoutAction(.clearIdleReadTimeoutTimer, context: context)
             self.runTimeoutAction(.clearIdleWriteTimeoutTimer, context: context)
--- a/Tests/AsyncHTTPClientTests/HTTP2ClientRequestHandlerTests.swift
+++ b/Tests/AsyncHTTPClientTests/HTTP2ClientRequestHandlerTests.swift
@@ -622,4 +622,37 @@ class HTTP2ClientRequestHandlerTests: XCTestCase {
         )
     }
 
+    func testErrorAfterResponseEndWhileRequestBodyIsStreamingDoesNotCrash() throws {
+        struct TestError: Error {}
+
+        let eventLoop = EmbeddedEventLoop()
+        let handler = HTTP2ClientRequestHandler(eventLoop: eventLoop)
+        let channel = EmbeddedChannel(handlers: [handler], loop: eventLoop)
+        try channel.connect(to: .init(ipAddress: "127.0.0.1", port: 80)).wait()
+
+        let request = MockHTTPExecutableRequest(
+            head: .init(version: .http1_1, method: .POST, uri: "http://localhost/"),
+            framingMetadata: .init(connectionClose: false, body: .stream),
+            raiseErrorIfUnimplementedMethodIsCalled: false
+        )
+
+        try channel.writeOutbound(request)
+        XCTAssertEqual(try channel.readOutbound(as: HTTPClientRequestPart.self), .head(request.requestHead))
+
+        try channel.writeInbound(HTTPClientResponsePart.head(.init(version: .http1_1, status: .ok)))
+        try channel.writeInbound(HTTPClientResponsePart.end(nil))
+
+        channel.pipeline.fireErrorCaught(TestError())
+        eventLoop.run()
+
+        XCTAssertFalse(channel.isActive)
+        XCTAssertEqual(
+            request.events.map(\.kind),
+            [
+                .willExecuteRequest, .requestHeadSent, .resumeRequestBodyStream, .receiveResponseHead,
+                .receiveResponseEnd,
+            ]
+        )
+    }
+
 }


Alex-Wengg:
Closing because I’m changing contribution focus. Thank you.

